PR #35, opened 2018-03-21
Fix getting result details in get_notes.

timopollmeier:
In the get_notes GMP command, get the correct result details by setting
 the id field in results_get.
Also check if uuid_result is defined in both get_notes and get_overrides
 and use the same method of setting the UUID for consistency.


src/gmp.c
```diff
@@ -11049,16 +11049,16 @@ buffer_overrides_xml (GString *buffer, iterator_t *overrides,
...
+              && override_iterator_result (overrides))
             {
               iterator_t results;
               get_data_t *result_get;
               result_get = report_results_get_data (1, 1,
                                                     1, /* apply_overrides */
                                                     0, /* autofp */
                                                     0  /* min_qod */);
-              result_uuid (override_iterator_result (overrides),
-                           &(result_get->id));
```
bjoernricks: Don't know what these two lines have been intended for. Is it ok to remove them?

timopollmeier: They are there to get the UUID of the result but that is already done assigning it to uuid_result near the start of the function.
The replacement line then copies uuid_result to the get data struct instead of getting it again using result_uuid.

PR #70, opened 2018-04-13
Remove log_gpgme. Depends on gvm-libs PR 70

jjnicola:
* src/lsc_crypt.c (log_gpgme): Remove static function and use instead
the global function from gvm-libs, since they were repetead and the
one from gvm-libs became global.

### DIFF
--- a/src/lsc_crypt.c
+++ b/src/lsc_crypt.c
@@ -156,41 +156,6 @@ get32 (const void *buffer)
   return value;
 }
 
-/**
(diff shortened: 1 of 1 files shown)


mattmundell:
@wiegandm I don't understand why the checks are failing.  It builds fine for me locally.  Perhaps because the checks ran before gvm-libs PR 70 was approved?

I don't seems to have permission to rerun the checks.  Could you try that?

PR #72, opened 2018-04-15
Update documentation.

janowagner:
* ChangeLog: Removed.

* CHANGES: Added notes on how to get detailed ChangeLog information.

* INSTALL: Increase to Debian 9 as the default build system.


CHANGES
```diff
@@ -1,3 +1,30 @@
...
+$ git clone https://github.com/greenbone/gvm.git
+$ cd gvm && git log
+
+
+gvm 8.0+beta1 (2018-XX-XX)
+
+This is the first beta release of the gvm module 1.0 for the Greenbone
+Vulnerability Management (GVM) framework.
+
+It was renamed from openvas-manager to gvm (greenbone vulnerability manager).
```
bjoernricks: Can we use Greenbone Vulnerability Manager here for consistency?

mattmundell: I also wondered about this, but I think he's matching the case of "gvm" (versus "GVM" above).

PR #112, opened 2018-05-18
Simplify .deb and .rpm LSC package generation

timopollmeier:
The gvm-lsc-rpm-creator script is replaced with a simpler script that
does not include makeself and a new script is used to generate Debian
packages directly with dpkg instead of using alien to convert the RPM.
Because of this, checks for alien are removed and the INSTALL doc is
updated accordingly.


tools/gvm-lsc-deb-creator.sh
```diff
@@ -0,0 +1,217 @@
...
+  exit 1
+fi
+
+#
+# Create data files
+#
+
+# Create .ssh directory
+mkdir -p "${SSH_DATA_DIR}"
+if [ 0 -ne "$?" ]
```
wiegandm: Consider using `if ! mkdir -p "${SSH_DATA_DIR}"` here

tools/gvm-lsc-deb-creator.sh
```diff
@@ -0,0 +1,217 @@
...
+
+# Copy public key
+AUTH_KEYS_FILE="${SSH_DATA_DIR}/authorized_keys"
+cp "${PUBKEY_FILE}" "${AUTH_KEYS_FILE}"
+
+# Creat doc directory
+mkdir -p "${DOC_DATA_DIR}"
+
+# Create Changelog
+cd "${DOC_DATA_DIR}"
```
wiegandm: Consider using `'cd ... || exit'` to not simply continue on `cd` failures

tools/gvm-lsc-deb-creator.sh
```diff
@@ -0,0 +1,217 @@
...
+CHANGELOG_FILE="${DOC_DATA_DIR}/changelog.Debian"
+echo "${PACKAGE_NAME} (${PACKAGE_VERSION}) experimental; urgency=low" > ${CHANGELOG_FILE}
+echo "" >> ${CHANGELOG_FILE}
```
wiegandm: Consider using `{ cmd1; cmd2; } >> file` instead of individual redirects.

timopollmeier:
@wiegandm: I've changed the scripts according to most of those hints but I'm not sure about the one about adding `|| exit` to all cd commands.
I wonder if we should add "set -e" to abort on any error or add `|| exit 1` to all commands.

wiegandm:
> I wonder if we should add "set -e" to abort on any error or add || exit 1 to all commands.

Good point! I think in the interest of both readability and consistent error handling, using `set -e` is indeed a good idea. I would suggest combining it with a `trap` handling the `ERR` signal. In this trap, you could do clean-ups (if necessary) and report/log that your script is exiting abnormally.

PR #121, opened 2018-05-30
Add create_tags GMP command.

timopollmeier:
This new command allows creating multiple tags for a list of resources.


timopollmeier:
Closing this because the tags are going to support multiple resources instead.

timopollmeier:
The PR for the new method of multiple resources per tag is #129.

PR #124, opened 2018-06-07
Reassign correct hosts when setting identifiers

mattmundell:
At the end of a scan, when the host identifiers lead to a new host being
required, only reassign the identifiers that pertain to the current
host, instead of reassigning all host identifiers.

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -60029,7 +60029,7 @@ hosts_set_identifiers ()
                  * HOST_START in otp.c refer to the new host. */
                 sql ("UPDATE host_identifiers SET host = %llu"
                      " WHERE source_id = '%s'"
-                     " AND name = 'ip'",
+                     " AND name = 'ip'"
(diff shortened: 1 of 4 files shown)


src/manage_sql.c
```diff
@@ -60027,13 +60025,15 @@ hosts_set_identifiers ()
...
               if (host_new)
-                /* Make sure all existing identifiers from this report refer to
-                 * this host.  Currently these will only be the Report Host
-                 * identifiers added for OTP HOST_START in otp.c. */
+                /* Make sure the Report Host identifiers added for OTP
+                 * HOST_START in otp.c refer to the new host. */
                 sql ("UPDATE host_identifiers SET host = %llu"
-                     " WHERE source_id = '%s';",
+                     " WHERE source_id = '%s'"
+                     " AND name = 'ip'",
```
timopollmeier: The comma in line 60032 should be removed.

PR #127, opened 2018-06-11
Amend schedule data returned by get_tasks.

timopollmeier:
When listing the scheduled tasks with get_tasks, the schedule now
includes the iCalendar and timezone and the order of the elements is now
more like the order in get_schedules.
The first_time and next_time ISO strings are copied to avoid one
overwriting the other when sending schedule data in get_tasks.
The schedule_info function now also calculates the next due time using
iCalendar.


bjoernricks:
@timopollmeier thanks a lot! :tada:

PR #142, opened 2018-06-29
added systemd units

GIJack:
Systemd .service file and errata
openvas-manager.default -> /etc/default/openvas

openvas-manager.logrotate -> /etc/logrotate.d/openvas-manager

systemd .service is self explaintory

### DIFF
--- a/systemd/openvas-manager.default
+++ b/systemd/openvas-manager.default
@@ -1,0 +1,8 @@
+OPTIONS=""
+
+# TLS Encryption options
+# Somewhat paranoid defaults. TLS1.2 only
+GNUTLS_PRIORITIES="SECURE256:+SUITEB192:+SECURE192:+SECURE128:+SUITEB128:-MD5:-SHA1:-VERS-TLS1.0:-VERS-TLS1.1:-VERS-SSL3.0"
+
+# Reasonably secure, if you need to connect with an old web browser.
+#GNUTLS_PRIORITIES="SECURE256:+SUITEB192:+SECURE192:+SECURE128:+SUITEB128:-MD5:+VERS-TLS1.0:+VERS-TLS1.1:-VERS-SSL3.0"
--- a/systemd/openvas-manager.logrotate
+++ b/systemd/openvas-manager.logrotate
@@ -1,0 +1,11 @@
+# logrotate for openvas manager
+/var/log/openvas/openvasmd.log {
+        rotate 4
+        weekly
+        compress
+        delaycompress
+        missingok
+	postrotate
+	    /bin/kill -HUP `pidof openvasmd`
+	endscript
+}
--- a/systemd/openvas-manager.service
+++ b/systemd/openvas-manager.service
@@ -1,0 +1,18 @@
+[Unit]
+Description=OpenVAS Manager
+After=network.target
+After=openvas-scanner.service
+Before=gsad.service
+
+[Service]
+Type=forking
+EnvironmentFile=-/etc/sysconfig/openvas-manager
+ExecStart=/usr/sbin/openvasmd --gnutls-priorities $GNUTLS_PRIORITIES $OPTIONS
+Restart=always
+RestartSec=1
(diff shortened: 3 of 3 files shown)


janowagner:
To be honest, this is more a packaging aspect which we try to keep out of the actual code modules.

mattmundell:
Thanks @GIJack.  Closing because we expect packagers to maintain these.

janowagner:
I do accept the point of view for package maintainers.
And we will reconsider this to help 3rd party integrations (see also here:
https://community.greenbone.net/t/about-the-source-edition-gse-category/176).

The patch however is not suitable for master, we would need to rework it, let us check about it.

Do you regard the word "hostile" appropriate when jointly supporting the Open Source community?
Our plate is overloaded like yours.

eli-schwartz:
So firstly, thank you for reconsidering this. It's very much appreciated. :)

Now to put this into context: Nearly three years ago, @GIJack requested of us to package some systemd units for the sake of system integration: https://bugs.archlinux.org/task/47873 (With the implication that these were further modifications of yet another collection of systemd unit files, actually, that originally came from Red Hat.) At the time, our distro packager said he'd look into it... then it promptly dropped off the radar for 2.5 years. Actually, during that time the packager in question retired from Arch altogether. Eventually @GIJack asked whatever happened, and I suggested trying to get this added upstream, hence this PR. And now you're suggesting that they aren't suitable for use in master at all -- maybe it's a good thing we never accepted it? This, right here, is why we prefer seeing service files that upstream has approved and coordinated, because you will always know the software better than we do.

Now, I totally 100% understand overload, and I'm neither surprised nor disappointed that these service files did not yet exist. And I would be very understanding if you had responded -- as you now did -- that you would consider it when you had the time, acknowledging that it is likely not a priority.
I was however surprised to see an initial rejection on the grounds that you didn't consider it to be anything to do with you. (My use of the word hostile was, in intent, 100% directed at this approach of "we don't want to have anything to do with it". Also, it was after midnight here, so, I may not have been at my most eloquent.)

Keep in mind that a fairly common pattern with community projects in general, is to see non-core things like this in, e.g. a contrib/ subdirectory of not-officially-endorsed resources that are nevertheless thought to be potentially useful, which is, sometimes, primarily maintained via third-party entities submitting improvements via pull requests. This form of coordination of downstream efforts is, on its own, an incredibly valuable thing for us to have, because it means every downstream can borrow each others' ideas.

GIJack:
What is not suitable about the patch?

cfi-gb:
@GIJack The PR is not suitable for master because the binary is now called ``gvmd`` instead of ``openvasmd``:

https://github.com/greenbone/gvm/#howto-use

Furthermore the log file location is now ``$PREFIX/log/gvm`` instead of ``$PREFIX/log/openvas``:

https://github.com/greenbone/gvm/blob/master/CMakeLists.txt#L180

and the logfile itself is called ``gvmd.log`` instead of ``openvasmd.log``:

https://github.com/greenbone/gvm/blob/master/src/gvmd_log_conf.cmake_in#L10

shibumi:
Hi, I would like to get this merged. I am the Arch Linux Maintainer for OpenVAS. I really don't think that this is just a 'packaging aspect'. One reason why the linux distributions have adopted `systemd` was that init scripts would be the same across all distributions. Furthermore upstream is now able to manage these systemd service files. If you include systemd service files in your project you can:

1. r**educe bug reports**, because you know that OpenVAS is started on all systems the same way. If something non default happens, you can close the bug report and refer to the systemd service files that are supported by upstream.
2. **harden your services**. You are able to harden your security application against malicious behavior using cgroups and namespaces (both supported by the linux kernel). If all distributions would do harden the service files by themselves you can expect more work for your team, because of more bug reports in the future regarding system startup or weird behavior of your application.

cfi-gb:
I have re-submitted a new PR with updated systemd files for the upcoming GVM-20.08 in #1240

PR #150, opened 2018-07-12
Limit resources shown by get_tags, update GMP doc

timopollmeier:
The number of resources shown in the details of a tag is now limited and
the resources_first and resources_rows attributes have been added to
control which and how many resources to list.
By default the number is 100, which can be controlled with the new
setting "Tag Resources Per Page".

The GMP documentation for get_tags has also been updated accordingly
and to consider some changes related to multiple resources per tag
previously not included.

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -3014,8 +3014,10 @@ get_system_reports_data_reset (get_system_reports_data_t *data)
  */
 typedef struct
 {
-  get_data_t get;    ///< Get args.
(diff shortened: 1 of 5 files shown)


bjoernricks:
I am not sure if this change is the best approach. GSA API wise it would be better to be able to get the resources of a tag via an extra request. This request could use the _normal_ rows and first parameters.

timopollmeier:
To keep things simpler, #155 has been created to remove the resources list altogether.
Other get_... commands and the filter keyword tag_id should be used instead.

PR #181, opened 2018-09-11
Some path settings are not changeable.

tuxmaster5000:


tuxmaster5000:
The else blocks are needed, to change the default path.

PR #183, opened 2018-09-11
Ranges and CIDR blocks in host_contains 7.0

timopollmeier:
This is a backport of #153 and #182 to OpenVAS-Manager 7.0

### DIFF
--- a/src/manage_pg.c
+++ b/src/manage_pg.c
@@ -132,7 +132,7 @@ manage_create_sql_functions ()
 
   sql ("CREATE OR REPLACE FUNCTION hosts_contains (text, text)"
        " RETURNS boolean"
-       " AS '%s/libgvm-pg-server', 'sql_hosts_contains'"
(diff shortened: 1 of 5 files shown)


src/manage_pg.c
```diff
@@ -130,6 +130,13 @@ manage_create_sql_functions ()
 
   sql ("SET role dba;");
 
+  sql ("CREATE OR REPLACE FUNCTION hosts_contains (text, text)"
+       " RETURNS boolean"
+       " AS '%s/libgvm-pg-server', 'sql_hosts_contains'"
```
mattmundell: This should be '%s/openvasmd/pg/libmanage-pg-server'.

mattmundell: Hmm, I wonder why the checks passed.  It doesn't build for me.

mattmundell: Oh, it builds, but does not run.

PR #188, opened 2018-09-19
Speed up report import

mattmundell:


src/manage_sql.c
```diff
@@ -21444,30 +21447,60 @@ create_report (array_t *results, const char *task_id, const char *task_name,
...
+                              quoted_description,
+                              quoted_scan_nvt_version,
+                              quoted_severity,
+                              quoted_qod,
+                              quoted_qod_type,
+                              quoted_nvt_oid,
+                              report);
+
+      /* Insert at most 300 results at a time. */
+      if (insert_count == 300)
```
timopollmeier: As said in the summary, I'd prefer using a macro here.
Somewhere near CREATE_REPORT_CHUNK_SIZE would probably be a good place to define it.

src/manage_sql.c
```diff
@@ -21528,13 +21565,73 @@ create_report (array_t *results, const char *task_id, const char *task_name,
...
+
+        g_free (quoted_host);
+        g_free (quoted_source_type);
+        g_free (quoted_source_name);
+        g_free (quoted_source_desc);
+        g_free (quoted_name);
+        g_free (quoted_value);
+
+        /* Insert at most 300 results at a time. */
+        if (insert_count == 300)
```
timopollmeier: As above, I'd prefer a macro here.

mattmundell: Thanks, both done in latest commit.

PR #193, opened 2018-09-22
Updated various db names after GVM renaming.

### LINKED ISSUE #175: gvm-portnames-update has a hardcoded old db name (tasks)
Default database name has been changed to 'gvmd' in GVM-10, but the script still uses the hard-coded old name ('tasks') for PostgreSQL.

`
if [ $POSTGRES -eq 1 ]
then
  SQL="psql -v ON_ERROR_STOP=1 -q --pset pager=off --no-align -d tasks -t"
else
  TASKS_DB="$DB_DIR/tasks.db"
  if [ ! -f $TASKS_DB ]; then
      echo "$TASKS_DB not found. Rebuild DB before updating port names.";
      exit 1;
  fi
  SQL="sqlite3 $TASKS_DB"
fi
`

cfi-gb:
Some remaining renaming for the changes in https://github.com/greenbone/gvm/commit/36c0e525ef9b53760f075670decf5a332a85d61c

Fixes #175

I guess https://github.com/greenbone/gvm/blob/v8.0%2Bbeta1/src/manage_migrators.c#L3510-L3516 is nothing what needs to be updated as well?


mattmundell:
> I guess https://github.com/greenbone/gvm/blob/v8.0%2Bbeta1/src/manage_migrators.c#L3510-L3516 is nothing what needs to be updated as well?

Yes that's correct.  Thanks for the patch.

PR #206, opened 2018-10-12
Add SQL preselection of users for Super permissions check 7.0

mattmundell:
Backport of PR #187 to Manager 7:

This adds "preselection" of Super permissions almost everywhere.  It's a continuation of #178, which did the same thing, but only for results.

Basically it takes this SQL:

    select * from examples where ... and exists (select * from permission where permission gives current_user super access over example.owner);

and turns it into:

    with has_super_access_over as (select stuff from permissions) select * from examples where ... and examples.owner is in super_access_over.

So it's checking who the current user has super access over first, then using that list to decide more quickly for each row if the user has permission to access the row.


mattmundell:
This patch is causing problems.  Globals are now not visible in GSA.  Probably also causing greenbone/gvmd/issues/242.

The problem is due to master having other changes in the permissions area.  At least greenbone/gvmd/commit/6afba0fbeb885368b0a8cd4de8970b307e6cd111.

So I'm going to revert this.  It's a bit crazy to be backporting such a big change anyway.

PR #207, opened 2018-10-13
[7.0] Remove alien_found() call for NSIS/EXE generation

cfi-gb:
EXE generation via NSIS is not using alien at all so just drop this call. The same was also already done in the master branch with the simplification of the .deb and .rpm calls: https://github.com/greenbone/gvm/pull/112

https://github.com/greenbone/gvm/pull/112/files#diff-2009e77ea7710f4ea32dfcde1ca1ff44L896

### DIFF
--- a/src/lsc_user.c
+++ b/src/lsc_user.c
@@ -888,12 +888,6 @@ lsc_user_exe_recreate (const gchar *name, const gchar *password,
   gchar *exe_path;
   int ret = -1;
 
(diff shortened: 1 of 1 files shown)


cfi-gb:
As an alternative of the alien_found() a new function similar to https://svn.wald.intevation.org/svn/openvas/branches/tools-attic/openvas-check-setup could be checking for the existence of makensis and reporting it separately if the binary wasn't found.

PR #209, opened 2018-10-16
Improve SCAP info update performance

timopollmeier:
This PR improves the performance of the SCAP data update with the following:
* a new command line option to split large updates into multiple transaction, which can help with very large updates like the first sync of the CVEs.
* selecting the CVE rowid only once for each one (not every time a affected product is processed) which speeds up large CVE updates by a few seconds.

### DIFF
--- a/src/manage_sql_secinfo.c
+++ b/src/manage_sql_secinfo.c
@@ -2424,8 +2424,17 @@ update_cve_xml (const gchar *xml_path, int last_scap_update,
                 {
                   entity_t product;
                   entities_t products;
+                  resource_t cve_rowid;
 
                   products = list->entities;
+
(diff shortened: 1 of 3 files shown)


src/manage_sql_secinfo.c
```diff
@@ -94,6 +99,21 @@ string_replace (const gchar *string, const gchar *to, ...)
   return ret;
 }
 
+/**
+ * @brief Increment transaction size, commit and reset at secinfo_commit_size.
+ *
+ * @param[in,out] current_size Pointer to current size to increment and compare.
+ */
+inline static void
+increment_transaction_size (int* current_size) {
```
mattmundell: Timo, could you move the closing brace to a new line?

PR #210, opened 2018-10-16
Fix privacy_algorithm check in modify_credential

timopollmeier:
This makes it possible to set the privacy_algorithm to a "aes", "des" or
an empty string when modifying an SNMP credential.


timopollmeier:
I've fixed the missing breaks in 485d5eb but I'm not sure how to best handle the empty privacy_algorithm consistently between creating and modifying a credential.

SNMPv3 allows the following cases according to [RFC 3411](https://tools.ietf.org/html/rfc3411#section-3.4.3):
* without authentication and without privacy
* with authentication but without privacy  
* with authentication and with privacy

My main question is how to handle unused data like a privacy password if no algorithm is selected. Should we keep it, discard it or require omitting it / setting it to empty?

mattmundell:
> My main question is how to handle unused data like a privacy password if no algorithm is selected. Should we keep it, discard it or require omitting it / setting it to empty?

So in GSA terms you're saying that the user can't have "Privacy Password" if "Privacy Algorithm" is None?

If so, then I think it's up to GSA to check this, and GMP should require the privacy password to be
omitted/empty if the privacy algorith is "".

timopollmeier:
I've updated the commands so "Privacy Password" and "Privacy Algorithm" have to be both set or both empty.

mattmundell:
Can confirm this works.  But this prevents the user from having an empty privacy password with a privacy algorthm.  Isn't that a vaild option?  Perhaps someone will want to use an empty password?

mattmundell:
Thanks for checking those.  I understand that it's a bad idea.  Just wondering if someone will need it because they have an unusual situation, or for testing.

PR #216, opened 2018-10-22
Email encryption

timopollmeier:
This PR is about adding PGP/GPG and S/MIME encryption options to the email alert method.


mattmundell:
Out of time this week so haven't tried it, but it looks good.

mattmundell:
Oh, didn't notice the circleci failures.  It built for me.  Maybe on happens on the release build?

timopollmeier:
The failed builds should be due to the changes in gvm-libs not being merged yet.
I'll try starting a new build once that's done.

PR #227, opened 2018-11-01
In create_report, init count before incrementing it

mattmundell:


mattmundell:
Was moved down in #188, but needed to appear twice.

PR #231, opened 2018-11-05
Fix various typos.

cfi-gb:
Fix various typos found by [codespell](https://github.com/codespell-project/codespell) 1.14

```
codespell . -S ".git,build,*.png,*.ttf,*.tiff" | egrep -v "([Mm]itre|Tim|MITRE|therefor|dependant)"
```
(For some words the -S parameter doesn't work correctly)

In some of the .c files i had removed some existing control chars as well.


src/manage_sql.c
```diff
@@ -22558,7 +22558,7 @@ where_levels_auto (const char *levels, const char *new_severity_sql,
       g_string_free (levels_sql, TRUE);
       levels_sql = g_string_new ("");
       /* It's not possible to override from or to the error severity, so no
-       * need to use the overriden severity here (new_severity_sql).  This
+       * need to use the overwritten severity here (new_severity_sql).  This
```
mattmundell: This should be "overridden" (past participle of "override").

src/manage_sql.c
```diff
@@ -44565,7 +44565,7 @@ find_override_with_permission (const char* uuid, override_t* override,
  * @brief Create an override.
  *
  * @param[in]  active      NULL or -1 on, 0 off, n on for n days.
- * @param[in]  nvt         OID of overrided NVT.
+ * @param[in]  nvt         OID of overwritten NVT.
```
mattmundell: "overridden"

src/lsc_user.c
```diff
@@ -43,7 +43,7 @@
  */
 #define G_LOG_DOMAIN "md manage"
 
-
```
mattmundell: These are page breaks.  Useful in editors like Emacs to navigate between sections in a file.

Happy for them to be removed though, because I don't use them much and they confuse everyone.  And we should rather have smaller files.

cfi-gb: Ah, good to know. I have removed this from the PR, probably should be handled in a separate PR then.

PR #239, opened 2018-11-08
Remove shadowing of variables

mattmundell:


mattmundell:
Depends on greenbone/gvm-libs#146

PR #246, opened 2018-11-10
Remove redundant declarations

mattmundell:


mattmundell:
SQLite warnings should be gone with 9b59bc3.

PR #250, opened 2018-11-13
Document all undocumented symbols

mattmundell:
Enables Doxygen to warn about all undocument symbols, then resolves the warnings.


mattmundell:
Thanks!  I resolved the conflict.

PR #254, opened 2018-11-13
Use debug builds for CircleCI

mattmundell:


mattmundell:
@wiegandm Hoping to enable this catch more issues during the checks.  What do you think?

wiegandm:
> @wiegandm Hoping to enable this catch more issues during the checks. What do you think?

Looks reasonable, anything for catching more issues in CI is good. :+1: 

That being said, there are also issue which only show up in the `Release` build type, like code depending on `assert()` statements which are not built when `NDEBUG` is defined.

I would suggest building for both build types in parallel. Feel free to merge this PR, I will create
a new PR for you to review.

mattmundell:
Great, thanks.

PR #255, opened 2018-11-13
Remove header, it's defined in the file

mattmundell:


mattmundell:
Hmm, I see that the header would be needed if sql_explain was used for debugging further up in the file.  But it looked wrong in that section for headers defined outside the file.

Good to hear we only need sql_explain_internal, thanks.

PR #259, opened 2018-11-14
Add CircleCI jobs to build with Clang

mattmundell:
These are the Debug builds, so the job fails if there are any Clang
warnings.


mattmundell:
Needs greenbone/gvmd/pull/258 for the Clang builds to succeed.

PR #260, opened 2018-11-14
Add CircleCI jobs for analyzing code with Clang

mattmundell:


mattmundell:
The Clang scans need greenbone/gvmd/pull/210 and greenbone/gvmd/pull/253 to pass.

PR #261, opened 2018-11-15
Adjust task status sorting for progress and Containers

mattmundell:


src/manage_sql.c
```diff
@@ -3367,6 +3367,24 @@ filter_clause (const char* type, const char* filter,
                   "                  AS temp_sub)"
                   "    END)"
                   " ASC");
+              else if ((strcmp (type, "task") == 0)
+                       && (strcmp (keyword->string, "status") == 0))
+                g_string_append_printf
+                 (order,
+                  " ORDER BY"
+                  "  (CASE WHEN target = 0"
```
timopollmeier: For the sorting to work as intended, this should also include an `OR target IS NULL`
or alternatively use `coalesce (target, 0)` instead of just `target`.

The same applies for all the other `target = 0` checks.

mattmundell: I saw this, but I think it is wrong that some tasks have NULL targets.  I am going to prepare another PR that ensures that container targets are 0 and not NULL.

mattmundell: PR to remove NULL targets: greenbone/gvmd/pull/263

PR #270, opened 2018-11-21
Add a remediation module

mattmundell:


mattmundell:
Requires greenbone/gvm-libs/pull/154

timopollmeier:
I also haven't been able to successfully test the confirmation but I'm not sure if this because of me editing the task to get different results.
However, I'd be fine with testing this again and maybe fixing it later so we don't have to deal with database version and code structure conflicts in other PRs for the time being.

mattmundell:
> The ticket name was empty when it should contain the NVT name according to the documentation.

This doesn't happen for me.  Is it happening in a certain situation?  Does the result NVT exist?  Is it an old "Open Port" result?

> When copying a ticket " Clone 1" is added. I think this should either be removed so consistently stays the NVT name or the user should be able to define the name.
> Result and various NVT fields are missing after copying a ticket (from a CVE scan result).
> There's a missing space in the "Attempt to delete a predefined resource" message.

These should be working now.

mattmundell:
> I also haven't been able to successfully test the confirmation but I'm not sure if this because of me editing the task to get different results.
> However, I'd be fine with testing this again and maybe fixing it later so we don't have to deal with database version and code structure conflicts in other PRs for the time being.

That's fine with me.

For confirmation the same host must have been scanned, the result must be gone, and there must be no SSH/SMB login failure results.

PR #286, opened 2018-12-11
Add tooltip to badges

davidak:
And remove text

@wiegandm what do you think about this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@
 
 # Greenbone Vulnerability Manager
 
-[![GitHub releases](https://img.shields.io/github/release/greenbone/gvmd.svg)](https://github.com/greenbone/gvmd/releases)
-[![Code Documentation Coverage](https://codecov.io/gh/greenbone/gvmd/branch/master/graphs/badge.svg?flag=documentation)](https://codecov.io/gh/greenbone/gvmd)`(Documentation Coverage)`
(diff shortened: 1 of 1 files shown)


wiegandm:
> @wiegandm what do you think about this?

As the tooltip is not visible by default, I'm afraid this gives the wrong impression. The initial assumption would be that this percentage refers to the test coverage, which is not the case.

mattmundell:
I guess Michael is right, but I think this patch is on the right track, because having
"(Documentation Coverage)" in brackets looks sloppy.  Maybe there's a badge
for documentation coverage?

wiegandm:
We could also re-include the Codecov logo by adding `logo=codecov` at the end:

[Code Documentation Coverage](https://codecov.io/gh/greenbone/gvmd)

mattmundell:
> We could also re-include the Codecov logo by adding `logo=codecov` at the end:
> 
> [Code Documentation Coverage](https://codecov.io/gh/greenbone/gvmd)

Looks good.

davidak:
> We could also re-include the Codecov logo by adding `logo=codecov` at the end:
> 
> [Code Documentation Coverage](https://codecov.io/gh/greenbone/gvmd)

Yes, that's much better. I implemented it in the last commit.

(Don't forget to squash when merging!)

PR #288, opened 2018-12-13
Make asset type mandatory in create_asset

timopollmeier:
The TYPE subelement of the ASSET is now mandatory and must be a type
supported by the command (currently only 'host').


bjoernricks:
We have discussed the issue and no objections from my side against merging anymore. Haven't looked at the code details so @mattmundell should decide if the pr is fine.

PR #301, opened 2018-12-26
Update with new standard headers.

janowagner:


src/comm.h
```diff
@@ -23,6 +17,13 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
+/**
+ * @file comm.h
```
mattmundell: Adding @file sections to these .h files includes them in the doxygen run so now the doc check is failing
because doxygen expects these headers to be documented.

I think we should check if the doxygen config can be adjusted to allow function declarations to be undocumented.

PR #307, opened 2019-01-01
Fix some of the errors found by Infer

mattmundell:
Most of these are minor issues.  Some are just cleanups that make it easier for Infer to understand the code.


src/manage_utils.c
```diff
@@ -812,16 +812,20 @@ icalendar_timezone_from_tzid (const char *tzid)
...
+          if (tz == NULL)
+            /* tzid is not a built-in timezone, fall back to UTC. */
+            tz = icaltimezone_get_utc_timezone ();
+        }
     }
-
-  // Fall back to UTC if tzid was null or not a built-in timezone
-  if (tzid == NULL)
+  else
+    /* tzid is NULL, fall back to UTC. */
```
mattmundell: @timopollmeier I've changed the behaviour of this function to match what was described in the comments.  Let me know if it's now incorrect.

timopollmeier: I think the difference would only matter for user-defined timezones but the `icalendar_timezone_from_tzid ()` function is only used for cases where such a timezone is not available, so it should be okay.

mattmundell: OK great I'll leave it like that.

PR #326, opened 2019-01-10
Fix rsync errors due to IP blocking

argusb:
The synchronization scripts do two rsyncs in quick succession, this leads to frequent IP blocking by the rsync server.

Fixed by adding a short sleep before the second (synchronization) rsync.

Also fixed a small typo.

### DIFF
--- a/tools/greenbone-certdata-sync.in
+++ b/tools/greenbone-certdata-sync.in
@@ -366,6 +366,8 @@ do_sync_community_feed () {
 }
 
 sync_certdata(){
+  # Sleep for ten seconds to prevent IP blocking.
+  sleep 10
   if [ -e $ACCESSKEY ]
   then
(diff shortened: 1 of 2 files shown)


xBytez:
@timopollmeier Any update on this? This is still ongoing.

cfi-gb:
Subsequent rsyncs shouldn't be a problem according to:

https://community.greenbone.net/t/failed-greenbone-certdata-sync-connection-timed-out-110-follow-up/1843/2

Could you check if your package maintainer is running both sync scripts at the same time like observed in:

https://community.greenbone.net/t/how-many-times-i-can-sync-nvts-scapdata-certdata/1978/4

cfi-gb:
Both scripts got a big overhaul and an additional one got added in GVM-20.08. As resolving the conflicts isn't really possible a replacement PR is available here: #1372 

Let's hope that this is getting merged soon.

PR #334, opened 2019-01-14
Make missing credential data no longer block slave_setup

timopollmeier:


src/manage.c
```diff
@@ -2978,45 +2982,48 @@ slave_setup (gvm_connection_t *connection, const char *name, task_t task,
                   || privacy_algorithm == NULL)
                 {
                   cleanup_iterator (&credentials);
-                  goto fail_esxi_credential;
+                  global_slave_esxi_credential_uuid = NULL;
```
mattmundell: Should be global_slave_snmp_credential_uuid, I think.

src/manage.c
```diff
@@ -2977,26 +2977,46 @@ slave_setup (gvm_connection_t *connection, const char *name, task_t task,
...
+                }
+              if (((privacy_password && strcmp (privacy_password, ""))
+                    || (privacy_algorithm && strcmp (privacy_algorithm, "")))
+                  && (password == NULL
+                      || auth_algorithm == NULL
+                      || privacy_password == NULL
+                      || privacy_algorithm == NULL))
+                {
+                  cleanup_iterator (&credentials);
+                  slave_esxi_credential_uuid = NULL;
```
mattmundell: Also should be snmp.

src/manage.c
```diff
@@ -2977,26 +2977,46 @@ slave_setup (gvm_connection_t *connection, const char *name, task_t task,
...
                   global_slave_esxi_credential_uuid = NULL;
                   g_warning ("Could not create slave SNMP credential"
-                             " (missing community, login, password,"
-                             " auth_algorithm, privacy_password"
-                             " or privacy_algorithm)."
+                             " (auth_algorithm must be set if password is"
+                             " given)."
+                             " Continuing without credential.");
+                }
+              if (((privacy_password && strcmp (privacy_password, ""))
```
mattmundell: Surely this should be an "else if", because the if above says it will fail if true.

src/manage.c
timopollmeier: I've changed this and the above in d01d4f3.

src/manage.c
timopollmeier: You're right. I've changed it in d01d4f3 as well.

PR #369, opened 2019-01-31
Remove make_unique for target and filter names

timopollmeier:
The option was removed for simplification of GMP and consistency with
other creation commands.
The wizards are adjusted accordingly and now use a timestamp to avoid
name conflicts when creating a target.


bjoernricks:
@timopollmeier could you add a changelog entry too?

PR #392, opened 2019-02-11
Add option to update NVTs via OSP

mattmundell:


mattmundell:
Requires https://github.com/greenbone/gvm-libs/pull/188

mattmundell:
Now updates NVT preferences too.  ospd-openvas is messing up the values in
the VT params, but I think it's correct on Manager side.

Also, the OSP NVT update does not do chunking like the OTP update does, but
that can come in a later PR.  Combining INSERTs would also be good.

src/manage_sql_nvts.c
```diff
@@ -1147,30 +1212,605 @@ manage_complete_nvt_cache_update (GList *nvts_list, GList *nvt_preferences_list)
...
+    }
+
+  name = entity_child (vt, "name");
+  if (name == NULL)
+    {
+      g_warning ("%s: VT missing NAME", __FUNCTION__);
+      return -1;
+    }
+
+  vt_refs = entity_child (vt, "vt_refs");
```
jjnicola: This was updated, should be
```suggestion
  vt_refs = entity_child (vt, "refs");
```

mattmundell: I think we are making bad naming choices:

OSP: VT/VT_REFS/REF  (VT_REFS but contains REF)
libs: nvti_t contains vtref_t.  (no space like in OSP)

What I would expect:
OSP: VT/REFS/REF
libs: gvm_vt_t contains gvm_ref_t

mattmundell: Thanks for changing VT_PARAMS.  PR uses REFS and PARAMS now.

src/manage_sql_nvts.c
```diff
@@ -1147,30 +1212,605 @@ manage_complete_nvt_cache_update (GList *nvts_list, GList *nvt_preferences_list)
...
+              print_entity_to_string (vt_param, debug);
+              g_warning ("%s: VT_PARAM: %s", __FUNCTION__, debug->str);
+              g_string_free (debug, TRUE);
+            }
+          else if (def == NULL)
+            {
+              GString *debug = g_string_new ("");
+              g_warning ("%s: VT_PARAM missing DEFAULT", __FUNCTION__);
+              print_entity_to_string (vt_param, debug);
+              g_warning ("%s: VT_PARAM: %s", __FUNCTION__, debug->str);
```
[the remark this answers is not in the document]
mattmundell: Manager now allows DEFAULT to be missing.  Those preferences should be in the DB now.

jjnicola:
Hi @mattmundell, I tried it again with the last changes and the nvt update works as expected. I can see now the preferences without a default value. Also it is able to recognize a new feed version.

PR #394, opened 2019-02-12
Nvt pref name

kroosec:
Change NVT preferences format to OID:PrefType:PrefName. This allows plugins to change their script names without breaking existing configs.

PR contains changes needed to be able to rebuild/update nvt's and run scans with https://github.com/greenbone/openvas-scanner/pull/275 using default configs (+ docmentation updates, changes I did on my way while grepping the code for hardcoded values etc,.)

Still needed:
- Migrators for existing scan configs.
- Code for pattern matching that uses sscanf() calls could be simplified and made more readable by using g_str_split(pref, ":", 2) instead, but I didn't want to make the PR more intrusive.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,7 +117,7 @@ include (CPack)
 
 ## Variables
 
-set (GVMD_DATABASE_VERSION 205)
+set (GVMD_DATABASE_VERSION 206)
 
 set (GVMD_SCAP_DATABASE_VERSION 15)
 
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -10120,19 +10120,17 @@ void
 buffer_config_preference_xml (GString *buffer, iterator_t *prefs,
                               config_t config, int hide_passwords)
 {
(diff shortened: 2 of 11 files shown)


kroosec:
> I think with hardcoded OIDs it's harder to tell what NVTs are, for example to check for typos and
> copy-paste errors, so I think it may be a good idea to define macros for them.
> 
> Also, are you going to add the migrator yourself or would you prefer me or Matt to implement it?

Thanks for the review @timopollmeier. I updated the PR to use #define's for the hardcoded values. I prefer if you do the migrator, as I may miss something there.

kroosec:
> I also noticed now that the `init_nvt_preference_iterator()` function would also need to be changed to take the OID as a parameter and to use the new format.
> The other NVT preference iterator functions getting parts of the name should also be updated.

Fixed in https://github.com/greenbone/gvmd/pull/394/commits/c7ef673c301e71c04ea8ef009b31dfff5442dbb7

PR #397, opened 2019-02-13
Speed up GET_REPORTS for GSA Report Details page (8.0)

mattmundell:
GSA's Results page gets the entire report at once instead of using the GMP page mechanism, which is slow for big reports.

Most of the time seems to be taken on the 15 or so queries done for each result of the report.


src/gmp.c
```diff
@@ -10577,24 +10599,25 @@ results_xml_append_nvt (iterator_t *results, GString *buffer)
...
  */
 void
-buffer_results_xml (GString *buffer, iterator_t *results, task_t task,
-                    int include_notes, int include_notes_details,
-                    int include_overrides, int include_overrides_details,
-                    int include_tags, int include_tags_details,
-                    int include_details,
-                    const char *delta_state, iterator_t *delta_results,
-                    int changed)
+buffer_results_xml_cert (GString *buffer, iterator_t *results, task_t task,
```
timopollmeier: I'm not sure if passing the status of the CERT database around is the best idea here.
If avoiding the database queries helps, have you considered storing the state in a variable returned by `manage_cert_loaded ()` and having an update function called at the start of every GMP command 
(or every command that uses the CERT database)?

Maybe the scheduler could also update this regularly in the main process in case it becomes relevant for
scheduled tasks.

src/manage_sql.c
```diff
@@ -24744,6 +24886,96 @@ DEF_ACCESS (result_iterator_qod_type, GET_ITERATOR_COLUMN_COUNT + 18);
...
+ *
+ * @param[in]  iterator  Iterator.
+ *
+ * @return 1 if overrides may exist, else 0.
+ */
+int
+result_iterator_may_have_overrides (iterator_t* iterator)
+{
+  if (iterator->done) return 0;
+  return iterator_int64 (iterator, GET_ITERATOR_COLUMN_COUNT + 25);
```
timopollmeier: This looks like just a simple boolean check, so I think you meant to use `iterator_int ()` here.
```suggestion
  return iterator_int (iterator, GET_ITERATOR_COLUMN_COUNT + 25);
```

src/manage_sql.c
```diff
@@ -26967,6 +27199,11 @@ compare_port_severity (gconstpointer arg_one, gconstpointer arg_two)
 void buffer_results_xml (GString *, iterator_t *, task_t, int, int, int, int,
                          int, int, int, const char *, iterator_t *, int);
 
+/** @todo Defined in gmp.c! */
+void buffer_results_xml_cert (GString *, iterator_t *, task_t, int, int, int,
```
timopollmeier: I'd prefer to have only version of `buffer_results_xml ()`, but if you think updating all the other uses would be too much
for this PR, I'd be fine with having it done later.

src/manage_sql.c
mattmundell: Yes, forgot to revert it during testing.  Fixed with 8879de9.

mattmundell:
> When trying to get one of my CVE reports, I'm getting an SQL error that does not occur without these changes:
> ERROR: more than one row returned by a subquery used as an expression

I can't reproduce, could you show me the full error message?

timopollmeier:
The error occurs in the result iterator query and with some testing I found that the problem is in the column that gets the "detected_by" value from the report_host_details.

src/manage_sql.c
```diff
@@ -23728,14 +23712,89 @@ where_qod (int min_qod)
...
       KEYWORD_TYPE_STRING },                                                  \
     { "(SELECT cve FROM nvts WHERE oid = nvt)", "cve", KEYWORD_TYPE_STRING }, \
+    { "(SELECT value"                                                         \
+      " FROM report_host_details"                                             \
+      " WHERE report_host = (SELECT id"                                       \
+      "                      FROM report_hosts"                               \
+      "                      WHERE report = results.report"                   \
+      "                      AND host = results.host)"                        \
+      " AND name = 'detected_by'"                                             \
+      " AND source_name = results.nvt)",                                      \
```
timopollmeier: This appears to be the iterator column that caused the SQL error for me, so it probably needs some limit and/or other criteria.

src/manage_sql.c
```diff
@@ -23816,6 +23875,82 @@ where_qod (int min_qod)
...
```
timopollmeier: This also needs to be limited as above.

mattmundell: Thanks, does 80515ff solve it?

src/manage_sql.c
mattmundell: No problem, done in 55f6cf2.

src/manage_sql.c
timopollmeier: Yes, I no longer get the error with this.

src/gmp.c
mattmundell: Avoiding the database queries helps immensely when they are occurring for every result in a big report.

By having the option to pass the CERT status around we have the ability to avoid these queries when we can and need to, but also the ability to let it happen automatically the rest of the time.

I get that using some sort of variable in manage_cert_loaded is an option, but I don't really see that improving the situation -- we then need to deal with how to recheck the variable, which adds complexity.  Even if it was a better idea, I wouldn't think it's something we should be taking on in this PR.

timopollmeier: Okay, let's leave it the way it is now.

PR #400, opened 2019-02-14
Check if NVT preferences exist before inserting

timopollmeier:
NVT preferences with a duplicate name will be ignored with a warning
instead of causing an error that will make the process abort.


timopollmeier:
One of the commits in greenbone/openvas-scanner#275 removes duplicates on the scanner side.
This may make this PR at obsolete but we may want keep it just for safety.

PR #401, opened 2019-02-14
Remove first_report and second_last_report from get_tasks.

janowagner:


src/gmp.c
```diff
@@ -18846,9 +18846,6 @@ handle_get_tasks (gmp_parser_t *gmp_parser, GError **error)
           first_report_id = task_iterator_first_report (&tasks);
```
timopollmeier: I think if we fully remove the GMP elements we no longer need the UUID of the first report or its result counts.
For the trend calculation we could directly get the second to last report and its result counts instead.

timopollmeier: There isn't anything wrong with keeping it, though. So we can merge this and remove the first report references later.

PR #407, opened 2019-02-20
Clang format 8.0

kroosec:
Not included in this PR are files gmp.c, manage_sql.c and manage_migrators.c due to the long SQL queries strings and/or macro defines of code blocks botching the formatting and making readability harder.

They are better done in separate PRs with disabling/enabling of clang-format where needed.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,44 @@
+---
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: 'false'
+AlignConsecutiveDeclarations: 'false'
+AlignEscapedNewlines: Left
+AlignOperands: 'true'
+AlignTrailingComments: 'true'
+AllowAllParametersOfDeclarationOnNextLine: 'false'
+AllowShortBlocksOnASingleLine: 'false'
+AllowShortCaseLabelsOnASingleLine: 'false'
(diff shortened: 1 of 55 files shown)


mattmundell:
> I think we already changed this in the Scanner, but I wonder if we shouldn't enable the IndentCaseLabels as it's more like we are used to.

I'm OK with switching to the new style.  I really think we should stick to one format for all modules, at least for the case labels.

> I'm also not sure if enabling BinPackArguments and BinPackParameters always make the code more readable, especially with long parameter lists.

Agree with Timo.  These are much easier to read in many situations when they are one per line.

kroosec:
> Looking at the failed build, I think the order of the `#include`s may be important in some instances
> like in manage_pg_server.c, so I think they should either be left alone or grouped in a way that preserves
> the order where necessary.

Fixed.

kroosec:
> > I'm also not sure if enabling BinPackArguments and BinPackParameters always make the code more readable, especially with long parameter lists.
> 
> Agree with Timo. These are much easier to read in many situations when they are one per line.

@mattmundell @timopollmeier Disabled them in last two commits. The output seems to be a bit too wasteful (and verbose in terms of lines of code.) especially with BinPackParameters disabled.

PR #411, opened 2019-02-27
Cleanup migrators

kroosec:
With this cleanup PR, you can migrate from gvmd version 7 to 9 (Master), but not from earlier versions.

Total code removed: ~14k LOC. 3rd commit formats the code in this file too, as the file became small enough to verify and review easily.


src/manage_migrators.c
```diff
@@ -170,6 +167,8 @@ typedef struct
  * not just for migrators, so perhaps the SQL interface should keep
  * track of the transaction, and rollback before aborting. */
 
+// clang-format off
```
mattmundell: INSERT_PERMISSION should be a function (insert_permission) instead of a macro, if that helps with formatting.

kroosec: Yes, it does. Added a commit to fix that.

PR #422, opened 2019-03-12
Add scanner type "OSP-OpenVAS"

timopollmeier:
This new scanner type uses OpenVAS scan configs and connects to an OSP
scanner (ospd-openvas).

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -36922,7 +36922,8 @@ create_task_check_config_scanner (config_t config, scanner_t scanner)
   ctype = config_type (config);
   stype = scanner_type (scanner);
 
-  if (ctype == 0 && stype == SCANNER_TYPE_OPENVAS)
+  if (ctype == 0
+      && (stype == SCANNER_TYPE_OPENVAS || stype == SCANNER_TYPE_OSP_OPENVAS))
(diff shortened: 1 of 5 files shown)


bjoernricks:
Do we really need an extra scanner type? Isn't it just an OSP scanner?

src/manage_sql.c
```diff
@@ -32603,14 +32603,20 @@ parse_osp_report (task_t task, report_t report, const char *report_xml)
...
-          if (manage_report_host_detail (global_current_report,
-                                         host,
-                                         desc))
+          if (manage_report_host_detail (report, host, desc))
             g_warning ("%s: Failed to add report detail for host '%s': %s",
                       __FUNCTION__,
                       host,
                       desc);
         }
+      else if (host && nvt_id && desc && (strcmp (nvt_id, "HOST_START") == 0))
```
mattmundell: This is a strange feature of OSP.

timopollmeier: I also think it would be a good idea to change this. How would you handle the host start and end times in OSP?
My suggestion would be to add a new host detail result type to OSP and use "host_start" and "host_end" as the names.
I'm not sure if we should do it in this PR already.

PR #424, opened 2019-03-14
Make users "self aware" (8.0)

mattmundell:


src/manage_sql.c
```diff
@@ -65074,6 +65098,15 @@ create_user (const gchar * name, const gchar * password, const gchar *comment,
   if (new_user)
     *new_user = user;
 
+  /* Every user can see themself. */
+  create_permission_internal ("GET_USERS",
+                              "Automatically created when adding user",
+                              NULL,
+                              user_uuid (user),
```
timopollmeier: I think you should use a variable for the user_uuid value so you can later free it.

mattmundell: Oops, thanks.

PR #434, opened 2019-03-20
Skip free of context, in case we are in the context (8.0)

mattmundell:
This can happen, for example, when running TEST_ALERT on a "Start Task" alert.

Manager forks during the parsing and in the fork it initialises a new parser to serve
GMP to the alert.


mattmundell:
> As long as there's only one context per process that gets reused as long as the process runs, this makes sense to me.

Yes, the context is only reused in forked processes that setup an internal GMP connection,
serve some GMP, then exit.

> I'm not sure if we have to keep the now excluded g_markup_parse_context_free() here, though.

Agreed.  I won't keep it in the master version.

PR #443, opened 2019-03-29
Add --slave-commit-size option (7.0)

timopollmeier:
This option allows splitting large transactions when updating reports
from a slave into multiple smaller ones, for example when resuming a
scan with many hosts and results.

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -54422,10 +54422,10 @@ update_from_slave (task_t task, entity_t get_report, entity_t *report,
         }
       hosts = next_entities (hosts);
 
-      current_commit_size ++;
+      current_commit_size++;
       if (slave_commit_size && current_commit_size >= slave_commit_size)
         {
-          sql_commit();
+          sql_commit ();
           sql_begin_immediate ();
           current_commit_size = 0;
         }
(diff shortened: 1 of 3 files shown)


src/manage_sql.c
```diff
@@ -54400,6 +54421,14 @@ update_from_slave (task_t task, entity_t get_report, entity_t *report,
                                     entity_text (host_start));
         }
       hosts = next_entities (hosts);
+
+      current_commit_size ++;
```
mattmundell: Don't think there should be a space.

src/manage_sql.c
```diff
@@ -54400,6 +54421,14 @@ update_from_slave (task_t task, entity_t get_report, entity_t *report,
...
+      if (slave_commit_size && current_commit_size >= slave_commit_size)
+        {
+          sql_commit();
```
mattmundell: Missing space.

src/manage_sql.c
```diff
@@ -54451,6 +54481,14 @@ update_from_slave (task_t task, entity_t get_report, entity_t *report,
                                   threat_message_type (entity_text (threat)),
                                   entity_text (description));
             if (current_report) report_add_result (current_report, result);
+
+            current_commit_size ++;
```
mattmundell: Extra space.

src/manage_sql.c
```diff
@@ -54451,6 +54481,14 @@ update_from_slave (task_t task, entity_t get_report, entity_t *report,
...
+            if (slave_commit_size && current_commit_size >= slave_commit_size)
+              {
+                sql_commit();
```
mattmundell: Missing space.

src/manage_sql.c
timopollmeier: The extra space has been removed in fd2fb81.

src/manage_sql.c
timopollmeier: The missing space has been added in fd2fb81.

PR #460, opened 2019-04-05
Ensure in_use is 1 or 0 (gvmd-8.0)

timopollmeier:
The functions config_in_use, trash_config_in_use and port_list_in_use
returned a count instead of the expected 1 or 0.
Also, trash_port_list_in_use is made more consistent with the other
functions.


mattmundell:
Tests are stuck, forcing merging.

PR #474, opened 2019-04-10
 Allow special characters in credential login names (master)

timopollmeier:
The login usernames can now also contain these characters: - _ \ @ .
This is a port of #59.

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -41588,7 +41588,7 @@ validate_credential_username (const gchar *username)
   s = username;
   while (*s)
     if (isalnum (*s)
-        || strchr ("-_\\@", *s))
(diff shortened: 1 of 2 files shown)


src/manage_sql.c
```diff
@@ -41828,11 +41853,20 @@ create_credential (const char* name, const char* comment, const char* login,
 
   /* Add non-secret data */
   if (login)
-    set_credential_data (new_credential,
-                         "username", login);
-  if (key_public)
```
mattmundell: Looks like this was removed.  Is that intentional?

timopollmeier: No, removing the key_public part was accidental and I've added it again in 2eed1cc.

PR #489, opened 2019-04-18
Fix create_report cache update at end of results (master)

timopollmeier:
The last cache rebuild was run before the last insert, so the results
from this were not counted.


src/manage_sql.c
```diff
@@ -22501,12 +22501,10 @@ create_report (array_t *results, const char *task_id, const char *task_name,
       g_free (quoted_qod_type);
     }
 
-  report_cache_counts (report, 1, 1, NULL);
```
mattmundell: This one could go in the else of the if below.

PR #496, opened 2019-04-24
Anonymize more IPs and hostnames in Anonymous XML 

mattmundell:


mattmundell:
> This works for me but I think you should also anonymize the hostname filter keyword as well.

Done now.

PR #510, opened 2019-04-30
Change alerts using old removed report formats

timopollmeier:
In check_db_report_formats the checks for alerts using removed report
formats will also change the alert method data to the TXT report format
and also consider the "send_report_format" data.
Alerts where the report format was removed before (or missing for other
reasons) are also changed to the TXT report format.

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -17192,7 +17192,7 @@ check_db_report_formats ()
   GDir *dir;
   gchar *path;
   const gchar *report_format_path;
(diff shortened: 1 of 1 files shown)


timopollmeier:
I've created #652 for an alternate approach that only replaces the references with an explicit command line call (`gvmd --optimize cleanup-report-formats`).

timopollmeier:
An alternative approach has been implemented in #652, so I'm closing this.

PR #512, opened 2019-04-30
Change alerts using old removed report formats (7.0)

timopollmeier:
In check_db_report_formats the checks for alerts using removed report
formats will also change the alert method data to the TXT report format
and also consider the "send_report_format" data.
Alerts where the report format was removed before (or missing for other
reasons) are also changed to the TXT report format.


timopollmeier:
An alternative approach has been implemented in gvmd 8.0 (#651) and master (#652), so I'm closing this.

PR #522, opened 2019-05-06
Get MODIFY_TAG working for all types when given a filter

mattmundell:
This adds a check to type_build_select to prevent a segfault when type_filter_columns
returns NULL, but more importantly it adds all type to type_filter_columns and
type_select_columns, so that they don't return NULL when given a valid type.


timopollmeier:
The idea of the generic column getter functions like type_select_columns() and type_filter_columns() is to use them for the aggregate iterator and the idea of type_build_select() was to have a generic way to generate the inner subquery of the aggregate iterators and later apply it to the normal init_[...]_iterator() functions.

On the one hand this contradicts the current idea to split the source files so each GMP resource type has its own set but on the other hand it avoids some redundancy and would make it easier to keep the normal and aggregate iterators consistent.

Perhaps we should come up with a new approach for the aggregates that is more compatible with splitting the source files while still avoiding redundancy in generating the outer SELECT.

mattmundell:
> The idea of the generic column getter functions like type_select_columns() and type_filter_columns() is to use them for the aggregate iterator and the idea of type_build_select() was to have a generic way to generate the inner subquery of the aggregate iterators and later apply it to the normal init_[...]_iterator() functions.

Note that type_build_select is used by the tag functions, and type_build_select uses
type_filter_columns, so type_filter_columns can not only support the types used
by aggregators.  That's the issue that this PR is solving.

I think that using these builder functions to init the normal iterator functions is a good goal.

> 
> On the one hand this contradicts the current idea to split the source files so each GMP resource type has its own set but on the other hand it avoids some redundancy and would make it easier to keep the normal and aggregate iterators consistent.

I don't think it's a contradiction.  The per-type files will use generic functions like
init_get_iterator that are located in other files like manage_sql_resources.c.
The generic functions can use the helper functions like type_build_select.
If necessary the per-type files can use the helper functions too.

That way we have the isolation and also the shared code.

> 
> Perhaps we should come up with a new approach for the aggregates that is more compatible with splitting the source files while still avoiding redundancy in generating the outer SELECT.

I haven't looked in detail, but yes we can adjust the aggregates if we have to
so that they fit better in a split file setup.  But we can split first, and keep the
aggregates in manage_sql.c, and see where to go from there.

timopollmeier:
I should have made it clearer what I found a bit redundant:
It's having a case for every type in tag_add_resources_filter() and tag_remove_resources_filter() when there is no need for using the specialized init_<type>_iterator() functions.

The normal iterator also gets columns that are not really necessary, which could affect performance for types that use subqueries for the iterator columns. For the tag resources filter we only need the id and what is used in the filter.

mattmundell:
> I should have made it clearer what I found a bit redundant:
> It's having a case for every type in tag_add_resources_filter() and tag_remove_resources_filter() when there is no need for using the specialized init__iterator() functions.

Yes it's not ideal, but I don't think it's too bad.  Hopefully will come up with
improvements, but I prefer the idea of letting the different types deal with
their own SQL in their separate files.

> 
> The normal iterator also gets columns that are not really necessary, which could affect performance for types that use subqueries for the iterator columns. For the tag resources filter we only need the id and what is used in the filter.

Good point.  I think this is a general problem with all these iterators.  They are
used in places that don't need all the columns.

PR #536, opened 2019-05-08
Mention anonymization in Anonymous XML description

timopollmeier:


src/report_formats/XML/report_format.xml
```diff
@@ -2,7 +2,7 @@
   <name>XML</name>
   <summary>Raw XML report.</summary>
   <description>
-    Complete scan report in GMP XML format.
+    Anonymized scan report in GMP XML format.
```
cfi-gb, line 5: This should have been changed in the Anonymous_XML/report_format.xml instead. Opened a PR #1191 to fix this.

mattmundell: thanks!

PR #550, opened 2019-05-17
Trim whole report when resuming slave scans (master)

timopollmeier:
When a slave task is resumed, the whole report is fetched from the slave
including results from already finished hosts, so these results would be
duplicated on the master if it kept the results from before.

(Port of #499 from openvas-manager-7.0)


timopollmeier:
This should only affect the report on the master so the slave should still be able to keep the results of the finished host and resume the scan from there.
The only time wasted would be for getting those results, sending them to the master and having them copied into the report on the master.
Not having to get the whole report from the slave would still save some time, though.

mattmundell:
> This should only affect the report on the master so the slave should still be able to keep the results of the finished host and resume the scan from there.
> The only time wasted would be for getting those results, sending them to the master and having them copied into the report on the master.

Oh yes, you're right.  So not too bad.

PR #554, opened 2019-05-23
Drop Postgres server-side module

mattmundell:


mattmundell:
> About [9390a54](https://github.com/greenbone/gvmd/commit/9390a5453d5fb0495cb1b3c0c776c9336e742d57): The main reason for using a custom function was avoiding SQL errors if the given regular expression is invalid.
> This should also be possible by wrapping the built-in regexp operator in a plpgsql function similar to try_exclusive_lock().

Thanks, yes.  I have a note to sort that out later.

mattmundell:
This is old, and a big challenge that should be split into multiple PRs, so closing.

PR #560, opened 2019-05-24
Fix various typos.

cfi-gb:
Found via codespell 1.15

There is still one left:

```
./src/gmp.c:19602: Vulnerabilitie  ==> Vulnerability
```

where the following is used and where i'm not sure if this is really correct or not:

```
INIT_GET (vuln, Vulnerabilitie);
```


mattmundell:
> Vulnerabilitie

This is correct.  INIT_GET appends an "s", so it becomes "Vulnerabilities".

PR #409, opened 2019-02-27
Clang format

kroosec:
Same as https://github.com/greenbone/gvmd/pull/407 with adjustments due to code differences.

PR #561, opened 2019-05-27
Add source code format checks

mattmundell:
This adds source code format checking to gvmd, but it only checks a few files (mostly .h files).

The rest of the files need considerable changes or are problematic for clang-format so
we should do them in stages, in further PRs.


mattmundell:
For some examples of the trouble area in the formatting, see https://github.com/greenbone/gvmd/pull/409

PR #570, opened 2019-05-29
Add vt_refs table

janowagner:
The table "vt_refs" is created and when a next VT update  is running, the table will be filled with the references of the VTs.

In total currently about 370K entries should appear in the table.
A subsequent feed update takes about 4-5 minutes and there is a additional log entry to follow this activity more easily.

The content of the vt_refs table is not used yet. The same information redundantly is still in the nvts table. Switching over to using the vt_refs and removing the redundant data in nvts is subject to subsequent pull requests. 

The new table and index is only implemented for PostgreSQL. SQLite3-based setups will fail during run time.

### DIFF
--- a/src/manage_pg.c
+++ b/src/manage_pg.c
@@ -3225,9 +3225,6 @@ create_tables ()
   sql ("SELECT create_index ('nvts_by_solution_type',"
        "                     'nvts', 'solution_type');");
 
-  sql ("SELECT create_index ('vt_refs_by_vt_oid',"
-       "                     'vt_refs', 'vt_oid');");
-
   sql ("SELECT create_index ('permissions_by_name',"
        "                     'permissions', 'name');");
   sql ("SELECT create_index ('permissions_by_resource',"
@@ -3250,6 +3247,9 @@ create_tables ()
 
   sql ("SELECT create_index ('tag_resources_trash_by_tag',"
        "                     'tag_resources_trash', 'tag');");
+
+  sql ("SELECT create_index ('vt_refs_by_vt_oid',"
+       "                     'vt_refs', 'vt_oid');");
 
 
 #if 0
--- a/src/manage_pg.c
+++ b/src/manage_pg.c
@@ -2842,6 +2842,13 @@ create_tables ()
        "  name text,"
        "  value text);");
 
+  sql ("CREATE TABLE IF NOT EXISTS vt_refs"
+       " (id SERIAL PRIMARY KEY,"
+       "  vt_oid text NOT NULL,"
+       "  type text NOT NULL,"
+       "  ref_id text NOT NULL,"
+       "  ref_text text);");
+
   sql ("CREATE TABLE IF NOT EXISTS nvt_preferences"
        " (id SERIAL PRIMARY KEY,"
        "  name text UNIQUE NOT NULL,"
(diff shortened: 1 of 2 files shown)


src/manage_pg.c
```diff
@@ -3218,6 +3225,9 @@ create_tables ()
   sql ("SELECT create_index ('nvts_by_solution_type',"
        "                     'nvts', 'solution_type');");
 
+  sql ("SELECT create_index ('vt_refs_by_vt_oid',"
```
mattmundell: This would be better after tag, because they're alphabetical (except for result and report, for some reason).

src/manage_sql_nvts.c
```diff
@@ -347,17 +347,37 @@ insert_nvt (const nvti_t *nvti)
...
+           quoted_cve, quoted_bid, quoted_xref, quoted_tag,
+           nvti_category (nvti), quoted_family, quoted_cvss_base, creation_time,
+           modification_time, nvti_oid (nvti), quoted_solution_type,
+           qod, quoted_qod_type);
+
+      sql ("DELETE FROM vt_refs where vt_oid = '%s';", nvti_oid (nvti));
+
+      for (i = 0; i < nvti_vtref_len (nvti); i++)
+        {
+          vtref_t *ref = nvti_vtref (nvti, i);
```
mattmundell: Would much prefer the the declarations and initialisations to be separate.  Makes all the code easier to read.

src/manage_sql_nvts.c
```diff
@@ -347,17 +347,37 @@ insert_nvt (const nvti_t *nvti)
...
+           " qod, qod_type)"
+           " VALUES ('%s', '%s', '%s', '%s', '%s',"
+           " '%s', %i, '%s', '%s', %i, %i, '%s', '%s', %d, '%s');",
+           nvti_oid (nvti), quoted_name,
+           quoted_cve, quoted_bid, quoted_xref, quoted_tag,
+           nvti_category (nvti), quoted_family, quoted_cvss_base, creation_time,
+           modification_time, nvti_oid (nvti), quoted_solution_type,
+           qod, quoted_qod_type);
+
+      sql ("DELETE FROM vt_refs where vt_oid = '%s';", nvti_oid (nvti));
```
mattmundell: I think this should rather be done all at once in the caller, like with nvt_preferences (e.g. "TRUNCATE nvt_preferences;").  Or do you have something else planned?  These deletes are usually slow.

src/manage_sql_nvts.c
```diff
@@ -347,17 +347,37 @@ insert_nvt (const nvti_t *nvti)
...
+
+      for (i = 0; i < nvti_vtref_len (nvti); i++)
+        {
+          vtref_t *ref = nvti_vtref (nvti, i);
+          gchar *quoted_id = sql_quote (vtref_id (ref));
+          gchar *quoted_text = sql_quote (vtref_text (ref) ? vtref_text (ref) : "");
+
+          sql ("INSERT into vt_refs (vt_oid, type, ref_id, ref_text)"
```
mattmundell: Just to keep in mind: single INSERTs a slower than combined INSERTs.  So the plan in the future would be to insert chunks of NVTs in one INSERT statement (same for vt_preferences).

src/manage_pg.c
janowagner: Fixed.

src/manage_sql_nvts.c
janowagner: Fixed.

src/manage_sql_nvts.c
janowagner: I planned subsequent changes in separate PR's. Handling refs and prefs in a consistent way.

src/manage_sql_nvts.c
janowagner: Yes, performance works is planned for subsequent PR's.

src/manage_sql_nvts.c
mattmundell: OK, great.

PR #572, opened 2019-05-29
Clean special option keywords in filters (master)

timopollmeier:
When splitting a filter, some special option filter keywords
(like "first", "rows", "apply_overrides", etc.) will be changed so the
relation is changed to "=".
Also, numeric option keywords will be adjusted so their values are in
the valid range (e.g. "first" is set to 1 if it is 0 or less).

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -1707,7 +1707,7 @@ cleanup_keyword (keyword_t *keyword)
     }
   else if (strcasecmp (keyword->column, "autofp") == 0)
     {
-      /* min_qod must be a percentage (between 0 and 2) */
+      /* autofp must be either 0, 1 or 2) */
       if (keyword->integer_value < 0)
         {
           g_free (keyword->string);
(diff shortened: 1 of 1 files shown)


src/manage_sql.c
```diff
@@ -1652,6 +1652,100 @@ parse_keyword (keyword_t* keyword)
...
+        {
+          g_free (keyword->string);
+          keyword->integer_value = 100;
+          keyword->string = g_strdup ("100");
+        }
+      keyword->relation = KEYWORD_RELATION_COLUMN_EQUAL;
+    }
+  else if (strcasecmp (keyword->column, "autofp") == 0)
+    {
+      /* min_qod must be a percentage (between 0 and 2) */
```
```suggestion
      /* min_qod must be a either 0, 1 or 2 */
```

timopollmeier: Fixed the typo in 688525df01931e0d85cb343d524da8e01971c6ed

timopollmeier:
I think having some feedback would be nice but I'd rather make it a warning message than making
the command fail.

mattmundell:
I think sending a warning back in the GMP response is getting too complex.

Not a big deal.  GSA could even check the filter before sending the command.

PR #574, opened 2019-05-29
Drop table nvt_cves

janowagner:
Assuming we have the new table vt_refs as proposed in
https://github.com/greenbone/gvmd/pull/570
we do not need the table "nvt_cves" anymore.

This PR removes the table and takes care vt_refs is used to achieve the same.
Therefore the above cited PR is mandatory before applying this PR. No conflict for compile time, but at run-time it would fail with SQL errors.

The PR does not include a migrator to remove existing tables. This should be subject to another
PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,7 +117,7 @@ include (CPack)
 
 ## Variables
 
-set (GVMD_DATABASE_VERSION 208)
+set (GVMD_DATABASE_VERSION 209)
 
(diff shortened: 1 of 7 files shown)


janowagner:
> Looks good. nvt_cves was added for speed, so something to keep an eye on.

The new vt_refs table should be pretty fast as well. But yes, we keep an eye on it.

> Could you add a migrator to remove nvt_cves too?

Done

PR #575, opened 2019-05-30
Drop otp nvt update

janowagner:
This pull request removed the handling of NVT updates via OTP.

Any code parts that are subsequently not needed anymore are also removed.


src/scanner.c
```diff
@@ -359,13 +359,11 @@ openvas_scanner_realloc ()
...
  * @brief Write as much as possible from the to_scanner buffer to the scanner.
  *
- * @param[in]  nvt_cache_mode  NVT cache mode.
- *
  * @return 0 wrote everything, -1 error, -2 wrote as much as scanner accepted,
  *         -3 did an initialisation step.
  */
 int
-openvas_scanner_write (int nvt_cache_mode)
+openvas_scanner_write (void)
```
mattmundell:
```suggestion
openvas_scanner_write ()
```

We use () everywhere else.

src/scanner.c
```diff
@@ -798,20 +739,17 @@ openvas_scanner_connected ()
...
  * @brief Initializes the already setup connection with the Scanner.
  *
- * @param[in]  cache_mode   NVT Cache mode if true, which means sending NVT_INFO
- *                          command to scanner in initial negotiation.
- *
  * @return 0 success, -1 error.
  */
 int
-openvas_scanner_init (int cache_mode)
+openvas_scanner_init (void)
```
mattmundell:
```suggestion
openvas_scanner_init ()
```

src/scanner.h
```diff
@@ -31,7 +31,7 @@ int
 openvas_scanner_read ();
 
 int
-openvas_scanner_write (int);
+openvas_scanner_write (void);
```
mattmundell:
```suggestion
openvas_scanner_write ();
```

src/scanner.h
```diff
@@ -70,7 +70,7 @@ int
 openvas_scanner_connected ();
 
 int
-openvas_scanner_init (int);
+openvas_scanner_init (void);
```
mattmundell:
```suggestion
openvas_scanner_init ();
```

src/manage_sql_nvts.c
```diff
@@ -1592,6 +1468,10 @@ update_nvts_from_vts (entity_t *get_vts_response,
       nvti_t *nvti = nvti_from_vt (vt);
 
       insert_nvt (nvti);
+// TODO: perhaps use chunked inserts? The next line is simply
```
mattmundell: Please rather remove this.  We will not forget about the chunking, and these TODOs
clutter up the code because there are so many possibilities.  We always have the old
code to refer to anyway.

src/scanner.c
janowagner: Reverted accordingly.

src/scanner.c
janowagner: Reverted accordingly.

src/scanner.h
janowagner: Reverted accordingly.

src/scanner.h
janowagner: Reverted accordingly.

src/manage_sql_nvts.c
janowagner: Removed.

janowagner:
> gvmd.c should no longer be calling update_nvt_cache. As a result init_gmpd,
> init_gmp, init_manage, etc no longer need the nvt_cache_mode argument.

Done.

janowagner:
> Also --osp-vt-update should change, because it is the only option. Could be
> replaced with --vt-update-socket. And we need a default in case it is not given.

I suggest to keep as is and consolidate once the entire OTP is replaced by OSP.

mattmundell:
> > Also --osp-vt-update should change, because it is the only option. Could be
> > replaced with --vt-update-socket. And we need a default in case it is not given.
> 
> I suggest to keep as is and consolidate once the entire OTP is replaced by OSP.

Would like to see some sort of warning so this doesn't catch people out, but I want
to merge this.  We can add in a subsequent PR.

PR #580, opened 2019-06-05
cmocka based unit tests for gvmd

mattmundell:
Exploring cmocka as an option for gvmd unit tests.


mattmundell:
Decided to go with Cgreen.

PR #582, opened 2019-06-09
Drop bid and xref from table nvts

janowagner:
This removes the items "bid" and "xref" from table "nvts".

### DIFF
--- a/src/manage_migrators.c
+++ b/src/manage_migrators.c
@@ -1755,12 +1755,19 @@ migrate_209_to_210 ()
       return -1;
     }
 
-  /* Update the database. */
+  /* Do nothing when SQLite is used. During a later
+   * migration the columns will automatically be removed.
+   */ 
+  if (! sql_is_sqlite3 ())
+    {
 
-  /* Remove the fields "bid" and "xref" from table "nvts". */
+      /* Update the database. */
 
-  sql ("ALTER TABLE IF EXISTS nvts DROP COLUMN bid CASCADE;");
-  sql ("ALTER TABLE IF EXISTS nvts DROP COLUMN xref CASCADE;");
+      /* Remove the fields "bid" and "xref" from table "nvts". */
+
+      sql ("ALTER TABLE IF EXISTS nvts DROP COLUMN bid CASCADE;");
+      sql ("ALTER TABLE IF EXISTS nvts DROP COLUMN xref CASCADE;");
+    }
 
   /* Set the database version to 210. */
 
--- a/src/manage_migrators.c
+++ b/src/manage_migrators.c
@@ -1737,6 +1737,60 @@ migrate_208_to_209 ()
   return 0;
 }
 
+/**
+ * @brief Migrate the database from version 208 to version 209.
+ *
+ * @return 0 success, -1 error.
+ */
+int
+migrate_209_to_210 ()
+{
+  sql_begin_immediate ();
+
+  /* Ensure that the database is currently version 209. */
+
+  if (manage_db_version () != 209)
+    {
+      sql_rollback ();
+      return -1;
+    }
+
+  /* Update the database. */
+
+  /* Remove the fields "bid" and "xref" from table "nvts".
+   * This is done by dropping the entire table and
+   * unset the sync time stamps to force gvmd to refresh the content. */
+
+  sql ("DROP TABLE IF EXISTS nvts CASCADE;");
+
+  sql ("DELETE FROM meta where name = 'nvts_feed_version' OR name = 'nvts_check_time';");
+
+  sql ("CREATE TABLE IF NOT EXISTS nvts"
+       " (id SERIAL PRIMARY KEY,"
+       "  uuid text UNIQUE NOT NULL,"
+       "  oid text UNIQUE NOT NULL,"
+       "  name text,"
+       "  comment text,"
+       "  cve text,"
+       "  tag text,"
+       "  category text,"
+       "  family text,"
+       "  cvss_base text,"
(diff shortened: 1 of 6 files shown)


src/manage_migrators.c
```diff
@@ -1737,6 +1737,60 @@ migrate_208_to_209 ()
...
+
+  /* Remove the fields "bid" and "xref" from table "nvts".
+   * This is done by dropping the entire table and
+   * unset the sync time stamps to force gvmd to refresh the content. */
+
+  sql ("DROP TABLE IF EXISTS nvts CASCADE;");
+
+  sql ("DELETE FROM meta where name = 'nvts_feed_version' OR name = 'nvts_check_time';");
+
+  sql ("CREATE TABLE IF NOT EXISTS nvts"
```
mattmundell: This is not necessary.  Manager will create the table on startup.  We only need
to add a CREATE TABLE when we're modifying a table that may not have
been created.

src/manage_migrators.c
```diff
@@ -1737,6 +1737,60 @@ migrate_208_to_209 ()
...
+      return -1;
+    }
+
+  /* Update the database. */
+
+  /* Remove the fields "bid" and "xref" from table "nvts".
+   * This is done by dropping the entire table and
+   * unset the sync time stamps to force gvmd to refresh the content. */
+
+  sql ("DROP TABLE IF EXISTS nvts CASCADE;");
```
mattmundell: Wouldn't an ALTER TABLE be simpler?

src/manage_sql.c
```diff
@@ -15535,19 +15535,32 @@ update_nvti_cache ()
...
       nvti_set_family (nvti, iterator_string (&nvts, 2));
       nvti_set_cvss_base (nvti, iterator_string (&nvts, 3));
-      nvti_add_refs (nvti, "cve", iterator_string (&nvts, 4), "");
-      nvti_add_refs (nvti, "bid", iterator_string (&nvts, 5), "");
-      nvti_add_refs (nvti, NULL, iterator_string (&nvts, 6), "");
-      nvti_set_tag (nvti, iterator_string (&nvts, 7));
+      nvti_set_tag (nvti, iterator_string (&nvts, 4));
+
+      init_iterator (&refs,
+                     "SELECT vt_oid, type, ref_id, ref_text"
```
mattmundell: Maybe skip vt_oid because it's no longer used?

src/manage_sql.c
```diff
@@ -15535,19 +15535,32 @@ update_nvti_cache ()
...
+                     " FROM vt_refs"
+                     " WHERE vt_oid = '%s';", iterator_string (&nvts, 0));
```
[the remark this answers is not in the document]
janowagner: Done.

src/manage_migrators.c
janowagner: I tried it when I developed it and it did not work, the existence of the table was required.

src/manage_sql.c
janowagner: Done.

src/manage_migrators.c
janowagner: Since it the data can be rebuild easily I thought a clean drop fresh rebuild
is a safe way. It seemed simpler to me.

mattmundell: A potential problem is that if a subsequent migrator wants to modify
the nvts table then it won't be there anymore.

I think we can skip the complex SQLite case because there's
no harm in having the extra columns.  We will drop SQLite
soon anyway.  So it should just be two of: ALTER TABLE nvts
DROP COLUMN xxx.

Also the ALTER TABLE matches what will be needed in
gvm-migrate-to-postgres.

src/manage_migrators.c
mattmundell: It should work.  But perhaps you ran into some dependency problem
that we don't usually see because the tables are usually created
together.

Would prefer ALTER TABLE, because we usually do it that way, and
only create the table if it is needed by the migrator.

janowagner: It is changed to apply ALTER TABLE now.

src/manage_migrators.c
janowagner: It is now using ALTER TABLE.

src/manage_migrators.c
```diff
@@ -1757,30 +1757,10 @@ migrate_209_to_210 ()
...
-       "  family text,"
-       "  cvss_base text,"
-       "  creation_time integer,"
-       "  modification_time integer,"
-       "  solution_type text,"
-       "  qod integer,"
-       "  qod_type text);");
+  /* Remove the fields "bid" and "xref" from table "nvts". */
+
+  sql ("ALTER TABLE IF EXISTS nvts DROP COLUMN bid CASCADE;");
```
mattmundell: Thanks for changing it Jan.

This needs a sql_is_sqlite3 check because we can't use ALTER TABLE
with SQLite.  For SQLite we can just do nothing.  Columns will vanish
when the user migrates to Postgres.

janowagner: Done.

src/manage_migrators.c
```diff
@@ -1755,12 +1755,19 @@ migrate_209_to_210 ()
       return -1;
     }
 
-  /* Update the database. */
+  /* Do nothing when SQLite is used. During a later
+   * migration the columns will automatically be removed.
+   */ 
+  if (! sql_is_sqlite3 ())
```
mattmundell: Should not have a space after the ! but we will catch that when we enable format checking for this file.

PR #584, opened 2019-06-11
Fix getting ssh port in target_osp_ssh_credential

timopollmeier:
Getting it from the iterator expects a target iterator, not a
credential iterator as actually used in the function, so it returned
the wrong value. Now target_ssh_port is used instead.


src/manage.c
```diff
@@ -4226,7 +4226,8 @@ target_osp_ssh_credential (target_t target)
   if (credential)
     {
       iterator_t iter;
```
mattmundell: I think naming this something more descriptive like "credentials", could have
helped avoid the error.

PR #590, opened 2019-06-14
Produce asciidoc GMP by hacking HTML schema generator

mattmundell:
A hack of the HTML schema generator to produce asciidoc for the GMP docs.

Not intended to be merged ever.

Run something like this to generate the asciidoc and HTML based on the asciidoc:

    gvm-cli socket --sockpath /home/mattm/var/run/gvmd.sock --gmp-username m --gmp-password m --xml '<help format="HTML"/>' | xmlstarlet sel -t -v /help_response/schema/text\(\) | base64 -d > /tmp/a.adoc && a2x -f xhtml /tmp/a.adoc


mattmundell:
Switching doc formats probably won't happen this way anymore, so closing.

PR #595, opened 2019-06-17
Fix log_notice syntax for unsupported database version

cfi-gb:
As reported in the #openvas IRC channel by user "andabata".


cfi-gb:
Backport in #596

PR #599, opened 2019-06-18
Drop long-unsupported '0' to indicate Open Port result

janowagner:
Remove results with nvt = '0' from DB and remove any special handling
or consideration of this case.

The oid '0' was used to indicate a open port detection in very early versions.

Any such results are removed during a migration. It is extremely unlikely that such
results are still in the database. This circumstance of '0' for the  oid isn't supported
in many places since a long time already.

Further, remaining special considerations are finally removed and simplify SQL queries
and XSLT.

Even in the case for some reason a new results get into the database that has nvt = '0',
this '0' will be simply treated like any other oid.   Well, in normal operation mode, this can
not happen, so it is more a theoretical consideration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,7 +117,7 @@ include (CPack)
 
 ## Variables
 
-set (GVMD_DATABASE_VERSION 210)
+set (GVMD_DATABASE_VERSION 211)
 
 set (GVMD_SCAP_DATABASE_VERSION 15)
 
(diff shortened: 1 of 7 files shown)


janowagner:
> I think it's still possible to create these "0" NVT by importing an old report that uses them.
> Should these be converted or rejected or are they out of use for so long that importing them should no longer be considered relevant?
> Besides this removing the special handling for these results looks like a good idea to me.

IMHO this case can be ignored from now on and can be considered irrelevant.

PR #600, opened 2019-06-18
Rename openvas to gvm in xslt namespace.

janowagner:


janowagner:
> The old HTML report format still uses the "openvas" namespace.
> Did you overlook this one or are you intentionally skipping it because we consider it obsolete?

I overlooked it. And it is IMHO also obsolete. However, I applied the changes to this as well.

PR #608, opened 2019-06-19
Remove code for starting a otp scan.

janowagner:
In run_task(), do not call anymore run_otp_task() as fall-though
for non-matching scanner type (not the case anyway).
Instead return a error code in case of an unknown scanner type.

Subsequently remove the now-unsed function run_otp_task().
And further delete all functions that become usused by this.

The diff looks complicated, but actually it is only changing
run_task() and other than that simply remove entire functions.


mattmundell:
Really hard to see what is changing here.

Also, I would expect a much more drastic impact from removing the running
of OTP tasks.  A lot of the complications in the gvmd.c select loop are to
support listening to the OTP scanner while also handling GMP input, if I
remember right.

janowagner:
> Also, I would expect a much more drastic impact from removing the running
> of OTP tasks. A lot of the complications in the gvmd.c select loop are to
> support listening to the OTP scanner while also handling GMP input, if I
> remember right.

Thats true and why I called it "remove code for starting...".
Since this diff was already a bit hard to read, I wanted to have separate PRs
for removing the code that handles a running OTP scan.

timopollmeier:
I've reviewed and merged #623, so I'm closing this one.

PR #610, opened 2019-06-21
Drop SQLite Support.

janowagner:
Using PostgreSQL is made mandatory for configuring, building and using gvmd.

The documentaton is updated accordingly.

The source codes for the specific SQLite database backend are removed.
Also some place where backend-depending SQL syntax was applied for SQLite3.

This patch does not address subsequent simplifications that could be applied
now that only one db backend is supported. Especially the transaction handling
could be improved further and reduce the database locks.


INSTALL.md
```diff
@@ -128,18 +119,146 @@ certificates for scanners, please see also section `Updating Scanner
 Certificates`.
 
 
-Choosing the Database Backend
------------------------------
+Configure PostgreSQL Database Backend
+-------------------------------------
+
+Setting up the PostgresSQL database
```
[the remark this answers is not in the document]
bjoernricks: Markdown used the # to indicate the heading level see https://guides.github.com/features/mastering-markdown/ This is more a rst syntax. So the line should be removed.

mattmundell:
```suggestion
## Setting up the PostgresSQL database
```

INSTALL.md
```diff
@@ -128,18 +119,146 @@ certificates for scanners, please see also section `Updating Scanner
 Certificates`.
 
 
-Choosing the Database Backend
------------------------------
+Configure PostgreSQL Database Backend
```
mattmundell:
```suggestion
# Configure PostgreSQL Database Backend
```

INSTALL.md
```diff
@@ -128,18 +119,146 @@ certificates for scanners, please see also section `Updating Scanner
...
+-------------------------------------
```
mattmundell:
```suggestion
```

janowagner:
I propose to re-work INSTALL.md with a separate PR.
I checked the other INSTALL.md and these all would need the same changes according
to the request (apply markdown syntax).

mattmundell:
> I propose to re-work INSTALL.md with a separate PR.
> I checked the other INSTALL.md and these all would need the same changes according
> to the request (apply markdown syntax).

OK, sounds good.

mattmundell:
> Also, why not do the sql_is_sqlite3 transaction and migrator cases now too?

@janowagner These seem pretty easy.  Let me know if you want me to do them.

PR #613, opened 2019-06-21
Add usage_type for configs and tasks

timopollmeier:
Configs are now marked as either a "scan" config or a "policy" while tasks are marked either "scan" tasks or "audits".

While the tasks and configs currently still behave the same, normal vulnerability assessments can now be handled separately from compliance audits.


src/manage_sql.c
```diff
@@ -36920,15 +36967,22 @@ create_config_from_scanner (const char *scanner_id, const char *name,
     }
   quoted_name = sql_quote (name ?: "");
   quoted_comment = sql_quote (comment ?: "");
+  if (usage_type && strcasecmp (usage_type, "policy") == 0)
+    quoted_usage_type = g_strdup ("policy");
```
mattmundell: Not really quoted ;)

timopollmeier: I've renamed it to "actual_usage_type" in 1bc4ce35b6e18ab855179403400f92d0f607ccaa and also changed it to a const char*, avoiding the allocation and freeing as it's using fixed values.

timopollmeier:
The documentation has been fixed now (395696dd1e960f366554c862b10774b4e5b132c0).

PR #615, opened 2019-06-24
Remove --backup

mattmundell:


mattmundell:
> This looks good to me but I think you should also remove the option from the man page (doc/gvmd.8.xml, generating a new doc/gvmd.8 and gvmd.html)

Good point.  Removed with b082715.

veol:
Why delete the backup option, so how do I backup now

PR #617, opened 2019-06-24
Remove GET_REPORTS type "assets"

mattmundell:


src/manage_sql.c
```diff
@@ -30643,24 +29523,20 @@ print_report_xml_start (report_t report, report_t delta, task_t task,
 
   if (delta)
     {
-      gchar *escaped_host = host ? g_markup_escape_text (host, -1) : NULL;
       gchar *escaped_delta_states = g_markup_escape_text (delta_states, -1);
       g_string_append_printf (filters_extra_buffer,
-                              "<host><ip>%s</ip></host>"
```
mattmundell: This is a mysterious element.  I think it is here by accident.  Parameter "host" applies to the assets report, and "host" was only passed in for the asset cases, so I don't see what it is doing in the delta report output.

mattmundell: A dug a bit deeper, and it looks like this was left over from the transition to named filters.  Before named filters "host" would filter the delta report by host.  But these days filtering by host is done with a filter keyword.

mattmundell:
> The changes look good to me.
> Maybe you could also simplify init_host_prognosis_iterator by removing all the
> pagination and filtering parameters but I don't consider it necessary for this PR.

Good idea, I'll make a new PR.

Issue #611, opened 2019-06-21
gvmd restarts with error: relation “information_schema_catalog_name” does not exist


jake-hoffer:
On 1 of our 8 test scanners (all running the latest stable release of gvm 10), gvmd restarts continuously, throwing the following error into our postgres log file:

    2019-06-19 18:49:13.257 EDT [12032] root@gvmd STATEMENT:  SELECT table_name, column_name,       pg_get_serial_sequence (table_name, column_name)  FROM information_schema.columns  WHERE table_schema = 'public'    AND pg_get_serial_sequence (table_name, column_name)        IS NOT NULL;
    2019-06-19 18:49:58.511 EDT [15421] root@gvmd ERROR:  relation "information_schema_catalog_name" does not exist

We tracked the source query down to `/src/manage_pg.c` line 3284, which contains this query:

    SELECT table_name, column_name, pg_get_serial_sequence (table_name, column_name) FROM information_schema.columns WHERE table_schema = 'public' AND pg_get_serial_sequence (table_name, column_name) IS NOT NULL;

We dug into why this error occurred on this machine, and not on any others, and found that it came down to the way postgres plans the query. The call to `pg_get_serial_sequence` in the `WHERE` clause throws an error if evaluated before the `table_schema = 'public'` condition. In most cases, these conditions are evaluated in the order they appear, and `pg_get_serial_sequence` is only called on things in the `public` schema. In some cases, however, postgres for some reason decides to evaluate `pg_get_serial_sequence` first, and chokes on its inability to identify a schema for which it does not have a search path (ie not `public`).

We rewrote the query into one that always works by using a precomputed CTE to trick postgres' query planner into evaluating the conditions in the desired order:

    WITH serial_sequences AS (SELECT table_name, column_name, pg_get_serial_sequence (table_name, column_name) p FROM information_schema.columns WHERE table_schema = 'public') SELECT * FROM serial_sequences WHERE p IS NOT NULL;

Would someone please double-check our suggestion and consider incorporating this (or similar) into forthcoming releases? Thank you!

asmaack:
After a  "sudo reboot", I'm getting the same/similar issue on one of our scanners. 

gmvd 8.0.2 will not start and throughs the same error:

 

```
[...]

PR #642, opened 2019-07-12
Fix issue #611 

vulnbe:
This PR fixes bug described here https://github.com/greenbone/gvmd/issues/611

- [ ] Tests
- [x] [CHANGELOG](https://github.com/greenbone/gvmd/blob/master/CHANGELOG.md)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Update result diff generation at delta reports [#650](https://github.com/greenbone/gvmd/pull/650)
 
 ### Fixed
-- An PostgreSQL statement order issue [#611](https://github.com/greenbone/gvmd/issues/611) has been addressed [#642](https://github.com/greenbone/gvmd/pull/642)
+- A PostgreSQL statement order issue [#611](https://github.com/greenbone/gvmd/issues/611) has been addressed [#642](https://github.com/greenbone/gvmd/pull/642)
 - Fix iCalendar recurrence and timezone handling [#654](https://github.com/greenbone/gvmd/pull/654)
 - Fix issues with some scheduled tasks by using iCalendar more instead of old period fields [#656](https://github.com/greenbone/gvmd/pull/655)
 - Fix an issue in getting the reports from GMP scanners [#659](https://github.com/greenbone/gvmd/pull/659) [#665](https://github.com/greenbone/gvmd/pull/665)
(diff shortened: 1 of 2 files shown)


timopollmeier:
I don't see why the `table_schema = 'public'` is evaluated before the `pg_get_serial_sequence` call in the `WITH` clause but not the original `SELECT`, so I'm not sure if it's safe to assume this is always the case.

Wouldn't it be safer to only select table_name and column_name (filtered by table_schema) in the `WITH` clause and then use `pg_get_serial_sequence` on those rows?

For example:
```
WITH table_columns AS (SELECT table_name, column_name FROM information_schema.columns WHERE table_schema = 'public')
SELECT *, pg_get_serial_sequence (table_name, column_name) FROM table_columns
WHERE pg_get_serial_sequence (table_name, column_name) IS NOT NULL;
```
or maybe
```
WITH table_columns AS (SELECT table_name, column_name FROM information_schema.columns WHERE table_schema = 'public'),
serial_sequences AS (SELECT *, pg_get_serial_sequence (table_name, column_name) AS p FROM table_columns)
SELECT * FROM serial_sequences WHERE p IS NOT NULL;
```

mattmundell:
I presume that WITH clauses are always evaluated before the selected columns, but I did expect a solution like Timo's first example (because it's clearer that the pg_get_serial_sequence is done later).

Don't mind either way, but would like to see a summary of the explanation from #611 in a comment above the statement.

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Update result diff generation at delta reports [#650](https://github.com/greenbone/gvmd/pull/650)
 
 ### Fixed
+- An PostgreSQL statement order issue [#611](https://github.com/greenbone/gvmd/issues/611) has been addressed [#642](https://github.com/greenbone/gvmd/pull/642)
```
mattmundell:
```suggestion
- A PostgreSQL statement order issue [#611](https://github.com/greenbone/gvmd/issues/611) has been addressed [#642](https://github.com/greenbone/gvmd/pull/642)
```

mattmundell:
Thank you!

asmaack:
Will you merge this into gvmd-8.0 as well?
:-)

mattmundell:
> Will you merge this into gvmd-8.0 as well?
> :-)

PR for this: https://github.com/greenbone/gvmd/pull/691

PR #647, opened 2019-07-21
Don't handle qod_type via tags.

janowagner:
This resolves a unintended redundancy. The qod_type
is handled explicitley in the table "nvts" and also in the "nvt" GMP
object. But it also still occurs in the tags element which does
not make sense. Analog to the timestamps it should have been removed
from there.

This patch uses the new API for the nvti object to add the QoD Type
directly when loading a OSP VT object rather than adding it to
the "tags" element.

Consequently, it will get the qod_type via the new API rather than
parse it from the tags section.

This change has two advantages:

* In column "tags" of table "nvts" the string element
"|qod_type=some-type" is not added anymore.
This saves space and parsing time.

* The qod_type is not parsed from the tags element anymore.
This saves parsing time.

The gains might not be significant though.

The user API is changed at the level of a bug fix:
The "qod_type=" part of the tags should have never been there
and never be used. It was undocumented anyway.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_sql_nvts.c
+++ b/src/manage_sql_nvts.c
@@ -252,7 +252,7 @@ insert_nvt (const nvti_t *nvti)
   quoted_cvss_base = sql_quote (nvti_cvss_base (nvti) ? nvti_cvss_base (nvti) : "");
 
   qod_str = tag_value (nvti_tag (nvti), "qod");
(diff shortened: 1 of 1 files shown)


janowagner:
The changes depend on https://github.com/greenbone/gvm-libs/pull/250

The build tests will fail until that patch is merged.

PR #671, opened 2019-08-08
Check and create default permissions individually (master)

timopollmeier:
The gvmd startup will now check and create the permissions for the
built-in roles and resources individually instead of just looking for
any permissions for each role.

**Checklist**:

- N/A Tests

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -16640,23 +16640,33 @@ check_db_settings ()
  *
  * Caller must ensure args are SQL escaped.
  *
- * @param[in]  role        Role.
+ * @param[in]  role_id     Role.
  * @param[in]  permission  Permission.
  */
 static void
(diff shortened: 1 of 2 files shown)


src/manage_sql.c
```diff
@@ -17595,10 +17613,14 @@ add_permissions_on_globals (const gchar *role_uuid)
 
 /**
  * @brief Ensure the predefined permissions exists.
+ *
+ * @param[in]  force_add   Whether to always try to create permissions.
  */
 static void
-check_db_permissions ()
+check_db_permissions (gboolean force_add)
```
mattmundell: What's the point of having force_add and must_add?  There is only one caller, so they will both always be true.

timopollmeier: I wasn't sure if we want to always do the individual checks, but I've removed the conditions now.

PR #684, opened 2019-08-11
Fix ORPHAN calculation in GET_TICKETS

mattmundell:
**Checklist**:

- [ ] Tests


mattmundell:
> The changes look okay so far but shouldn't the tickets also count as orphaned if the report has been deleted?

Yes good point, but I think that should be a new PR.

PR #697, opened 2019-08-15
Use status text in ticket iterator columns

timopollmeier:
This allows using the status text in filters instead of the internal
numbers that are not visible in GMP.

**Checklist**:

- Tests N/A

### DIFF
--- a/src/manage_sql_tickets.c
+++ b/src/manage_sql_tickets.c
@@ -80,31 +80,6 @@ ticket_status_integer (const char *status)
 }
 
 /**
- * @brief Get ticket status name from DB identifier.
- *
- * @param[in]   status  Status integer.
- *
- * @return Status name.
- */
-static const gchar *
-ticket_status_name (ticket_status_t status)
-{
-  switch (status)
-    {
-      case TICKET_STATUS_OPEN:
-        return "Open";
-      case TICKET_STATUS_FIXED:
-        return "Fixed";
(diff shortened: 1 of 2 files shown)


src/manage_sql_tickets.c
```diff
@@ -157,7 +132,14 @@ ticket_status_name (ticket_status_t status)
    { "host", NULL, KEYWORD_TYPE_STRING },                                     \
    { "location", NULL, KEYWORD_TYPE_STRING },                                 \
    { "solution_type", NULL, KEYWORD_TYPE_STRING },                            \
-   { "status", NULL, KEYWORD_TYPE_STRING },                                   \
+   { "(CASE status"                                                           \
+     " WHEN 0 THEN 'Open'"                                                    \
```
mattmundell:
```suggestion
     " WHEN " TICKET_STATUS_OPEN " THEN 'Open'"                                                    \
```

Would this work?  Would be neater.

timopollmeier: I don't think this will work because `TICKET_STATUS_OPEN` etc. are part of an enum, not preprocessor constants, so they cannot be used like string literals.

A way around this would be defining a function where we can insert the values into the format string of the sql() call. 
Unless there are other place where we could use the function, I don't think this is worth the effort, though.

mattmundell: Oh right, of course.

Another option would be to change the enum to definitions, but no big deal.

PR #704, opened 2019-08-16
Implement report return when starting OSP tasks

mattmundell:
This gets tasks of scanner type GMP working again.

**Checklist**:

- [ ] Tests


src/manage.c
```diff
@@ -4074,20 +4074,24 @@ launch_osp_openvas_task (task_t task, target_t target, const char *scan_id,
...
-  char *report_id, title[128], *error = NULL;
+  char *id, title[128], *error = NULL;
   int rc;
 
   assert (task);
   assert (target);
 
-  if (create_current_report (task, &report_id, TASK_STATUS_REQUESTED))
+  if (report_id == NULL)
+    report_id = &id;
```
timopollmeier: Overwriting the parameter seems inconsistent with how we handle NULL output pointers elsewhere.
I'd prefer to use a local variable for report_id and assigning to the output parameter where appropriate if it the pointer is not NULL.

mattmundell: Used the more usual style in bd7d132.

PR #713, opened 2019-08-22
Cleanout some left over OTP fork handling

mattmundell:
**Checklist**:

- [ ] Tests


mattmundell:
This doesn't update process_gmp and it's caller (the wizards).  And it doesn't remove the now unused forked global variable.  I'll do those in the next PR.

PR #717, opened 2019-08-23
Check if hosts are noticeable in slave scans

timopollmeier:
When hosts are added to the assets from a slave scan, perform the usual
checks of whether each host is alive and has results.

**Checklist**:

- Tests N/A

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -48630,7 +48630,7 @@ update_from_slave (task_t task, entity_t get_report, entity_t *report,
     {
       if (strcmp (entity_name (host_start), "host") == 0)
         {
-          entity_t ip;
+          entity_t ip, end;
           char *uuid;
 
           ip = entity_child (host_start, "ip");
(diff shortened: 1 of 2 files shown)


src/manage_sql.c
```diff
@@ -56371,6 +56380,16 @@ hosts_set_details (report_t report)
...
+       /* Ensure that every report host detail has a corresponding host
+        *  in the assets. */
+       " AND EXISTS (SELECT host"
+       "               FROM host_identifiers"
+       "              WHERE source_id = (SELECT uuid FROM reports"
+       "                            WHERE id = %llu)"
+       "                AND (SELECT name FROM hosts WHERE id = host)"
+       "                      = (SELECT host FROM report_hosts"
+       "                         WHERE id = report_host_details.report_host)"
+       "             LIMIT 1)"
```
mattmundell: I don't think LIMIT is ever considered inside EXISTS, so this line could be left out for neatness.

src/manage_sql.c
```diff
@@ -56371,6 +56380,16 @@ hosts_set_details (report_t report)
        " AND (SELECT value = 'yes' FROM task_preferences"
        "      WHERE task = (SELECT task FROM reports WHERE id = %llu)"
        "      AND name = 'in_assets')"
+       /* Ensure that every report host detail has a corresponding host
+        *  in the assets. */
+       " AND EXISTS (SELECT host"
+       "               FROM host_identifiers"
+       "              WHERE source_id = (SELECT uuid FROM reports"
+       "                            WHERE id = %llu)"
```
```suggestion
       "                                 WHERE id = %llu)"
```

src/manage_sql.c
```diff
@@ -56371,6 +56380,16 @@ hosts_set_details (report_t report)
        " AND (SELECT value = 'yes' FROM task_preferences"
        "      WHERE task = (SELECT task FROM reports WHERE id = %llu)"
        "      AND name = 'in_assets')"
+       /* Ensure that every report host detail has a corresponding host
+        *  in the assets. */
+       " AND EXISTS (SELECT host"
```
mattmundell: No need to specify the column with EXISTS.

src/manage_sql.c
timopollmeier: I've addressed this and the following points in 0435471b341bf9b81a156b6835928c51ea5dfdb8.

PR #719, opened 2019-08-25
Make several NVTI tags internally explicit.  

janowagner:
This change makes the elements "summary", "insight", "detection", "impact" and "affected" explicit columns of the table "nvts".

These elements are already explicit in the vt object incoming via OSP.

For keeping GMP API compatibility, the 5 elements are serialized into the "tag" string.

The API is not modified, the changes are only internally.

This PR requires https://github.com/greenbone/gvm-libs/pull/258

**Checklist**:

- [ ] Tests N/A

### DIFF
--- a/src/manage_sql_nvts.c
+++ b/src/manage_sql_nvts.c
@@ -1128,7 +1128,6 @@ get_tag (entity_t vt)
                                   "%scvss_base_vector=%s",
                                   first ? "" : "|",
                                   entity_text (severity));
-          first = 0;
         }
       else
         g_warning ("%s: no severity", __FUNCTION__);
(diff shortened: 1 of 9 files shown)


src/manage_sql_nvts.c
```diff
@@ -1072,46 +1136,6 @@ get_tag (entity_t vt)
...
     g_warning ("%s: no severities", __FUNCTION__);
 
-  child = entity_child (vt, "detection");
-  if (child)
-    {
-      if (strlen (entity_text (child)))
-        {
-          g_string_append_printf (tag,
-                                  "%svuldetect=%s",
-                                  first ? "" : "|",
```
mattmundell: Because  these cases have been removed, the variable first is no longer used after the assignment on line 1031.  So the assignment is no longer needed, which is causing the build checks to fail.

PR #720, opened 2019-08-26
Check if hosts are noticeable in slave scans (7.0)

timopollmeier:
**Checklist**:

- Tests N/A

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -54467,13 +54467,17 @@ update_from_slave (task_t task, entity_t get_report, entity_t *report,
   current_commit_size = 0;
   while ((host_start = first_entity (hosts)))
     {
-      if (strcmp (entity_name (host_start), "host_start") == 0)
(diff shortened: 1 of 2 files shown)


src/manage_sql.c
```diff
@@ -54467,31 +54467,35 @@ update_from_slave (task_t task, entity_t get_report, entity_t *report,
...
           if (end
               && entity_text (end)
               && strcmp (entity_text (end), "")
               && report_host_noticeable (current_report,
-                                         entity_text (host)))
+                                         entity_text (ip)))
             {
               uuid = report_uuid (current_report);
-              host_notice (entity_text (host), "ip", entity_text (host),
-                           "Report Host", uuid, 1, 1);
```
mattmundell: Looks like a space was lost here.  Might want to put it back if you're trying to make it the same as master.

mattmundell: Oh this was actually consistent with master (but not my slightly out of date master).  Anyway, thanks for fixing.

PR #728, opened 2019-08-29
Add report_id filter keyword for TLS certificates

timopollmeier:
This allows using the filter to select the TLS certificates that were found in a particular report.

**Checklist**:

- Tests N/A

### DIFF
--- a/src/manage_sql_tls_certificates.c
+++ b/src/manage_sql_tls_certificates.c
@@ -197,19 +197,18 @@ tls_certificate_extra_where (const char *filter)
 
   report_id = filter_term_value (filter, "report_id");
 
-  g_message ("report_id=\"%s\"", report_id);
-
   if (report_id)
     {
       gchar *quoted_id;
       quoted_id = sql_quote (report_id);
       g_string_append_printf (ret,
                               " AND"
-                              " (EXISTS (SELECT * FROM"
-                              "  tls_certificate_source_origins AS src_orig"
-                              "  WHERE tls_certificate = tls_certificates.id"
-                              "    AND origin_type = 'Report'"
-                              "    AND origin_id = '%s'))",
+                              " (EXISTS"
+                              "   (SELECT * FROM"
+                              "    tls_certificate_source_origins AS src_orig"
+                              "    WHERE tls_certificate = tls_certificates.id"
+                              "      AND origin_type = 'Report'"
+                              "      AND origin_id = '%s'))",
                               quoted_id);
       g_free (quoted_id);
     }
--- a/src/manage_sql_tls_certificates.c
+++ b/src/manage_sql_tls_certificates.c
@@ -181,6 +181,45 @@ tls_certificate_select_columns ()
 }
 
 /**
+ * @brief Get extra_where string for a TLS certificate iterator or count.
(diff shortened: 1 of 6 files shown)


src/manage_sql_tls_certificates.c
```diff
@@ -180,6 +180,45 @@ tls_certificate_select_columns ()
...
+tls_certificate_extra_where (const char *filter)
+{
+  GString *ret;
+  gchar *report_id;
+
+  ret = g_string_new ("");
+
+  report_id = filter_term_value (filter, "report_id");
+
+  g_message ("report_id=\"%s\"", report_id);
```
mattmundell: Is this line intentional?  Should be g_debug?

src/manage_sql_tls_certificates.c
```diff
@@ -180,6 +180,45 @@ tls_certificate_select_columns ()
...
+
+  if (report_id)
+    {
+      gchar *quoted_id;
+      quoted_id = sql_quote (report_id);
+      g_string_append_printf (ret,
+                              " AND"
+                              " (EXISTS (SELECT * FROM"
+                              "  tls_certificate_source_origins AS src_orig"
```
mattmundell: I know it's a small detail, but I'd really like to keep SQL statement indented under the start of the statement, for consistency.  I'd much rather that the "(SELECT..." went on the line below the EXISTS, if that's necessary to keep the line short (or I would put all the args on the line under g_string_append_printf).

src/manage_sql_tls_certificates.c
```diff
@@ -192,9 +231,33 @@ tls_certificate_count (const get_data_t *get)
 {
   static const char *filter_columns[] = TLS_CERTIFICATE_ITERATOR_FILTER_COLUMNS;
   static column_t columns[] = TLS_CERTIFICATE_ITERATOR_COLUMNS;
+  gchar *filter;
+  char *extra_where;
+  int ret;
+
+  if (get->filt_id && strcmp (get->filt_id, FILT_ID_NONE))
```
mattmundell: Would be great if we could get this filter setup into a function at some point.

src/manage_sql_tls_certificates.c
timopollmeier: This was just some test output I forgot to remove before committing.
I've removed it in 635558d6b5a02a8e99d5cf7edeaf1a6b7525ad22.

src/manage_sql_tls_certificates.c
timopollmeier: I must have overlooked this when I changed it from a count(*) to an EXISTS clause.
I've changed it in 635558d6b5a02a8e99d5cf7edeaf1a6b7525ad22.

src/manage_sql_tls_certificates.c
timopollmeier: I agree, this kind of filter handling is a lot of duplicated code that should be in a function. 
Maybe we could also get rid of the `filter ? filter : get->filter` by updating `get->filter` in `get_init()`.

mattmundell: Yeah, that might be a good idea.

PR #729, opened 2019-08-30
 Make --osp-vt-update optional, use default scanner

timopollmeier:
The OSP VT update socket no longer has to be given with the command line
if the socket path of the "OpenVAS Default" scanner can be used instead.
Also, update_nvt_cache_retry will no longer fork endlessly if the socket
path is still NULL for some reason and the documentation is updated.

**Checklist**:

- Tests N/A

### DIFF
--- a/src/gvmd.c
+++ b/src/gvmd.c
@@ -1795,7 +1795,8 @@ gvmd (int argc, char** argv)
           "<name>" },
         { "osp-vt-update", '\0', 0, G_OPTION_ARG_STRING,
           &osp_vt_update,
-          "Unix socket for OSP NVT update.  Default is to do an OTP update.",
+          "Unix socket for OSP NVT update.  Defaults to the path of the"
+          "'OpenVAS Default' scanner if it is an absolute path.",
           "<scanner-socket>" },
         { "password", '\0', 0, G_OPTION_ARG_STRING,
           &password,
--- a/src/gvmd.c
+++ b/src/gvmd.c
@@ -99,6 +99,7 @@
 #include <gvm/util/serverutils.h>
 
 #include "manage.h"
+#include "manage_sql_nvts.h"
 #include "manage_sql_secinfo.h"
 #include "gmpd.h"
 #include "utils.h"
@@ -291,11 +292,6 @@ static int update_in_progress = 0;
  * @brief Logging parameters, as passed to setup_log_handlers.
  */
 GSList *log_config = NULL;
-
-/**
- * @brief File socket for OSP NVT update.  NULL to update via OTP.
- */
-static gchar *osp_update_socket = NULL;
 
 
 /* Helpers. */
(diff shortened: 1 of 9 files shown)


src/gvmd.c
```diff
@@ -1133,8 +1129,15 @@ update_nvt_cache_retry ()
...
       else if (child_pid == 0)
         {
+          const char *osp_update_socket;
+          osp_update_socket = get_osp_vt_update_socket ();
           if (osp_update_socket)
             exit (update_nvt_cache_osp (osp_update_socket));
+          else
+            {
+              g_warning ("%s: No OSP VT update socket set", __FUNCTION__);
+              exit (-1);
```
mattmundell: I think this should be EXIT_FAILURE because that's what we use everywhere else.

src/gvmd.c
```diff
@@ -2660,6 +2656,31 @@ gvmd (int argc, char** argv)
...
+        {
+          g_debug ("%s: Using OSP VT update socket from default OpenVAS"
+                   " scanner: %s",
+                   __FUNCTION__,
+                   default_socket);
+          set_osp_vt_update_socket (default_socket);
+        }
+      else
+        {
+          g_critical ("%s: No OSP VT update socket found."
```
mattmundell: I'm OK with merging it this way now, but I wonder if it wouldn't be better to check this every time the update is scheduled (and just give a warning if there's no socket path).  Then the user would be able to update the default scanner via GMP, and would be able to start gvmd without needing to have this set up correctly.

src/manage_sql_nvts.c
```diff
@@ -61,6 +61,40 @@ blank_control_chars (char *string)
...
+ */
+static gchar *osp_vt_update_socket = NULL;
+
+/**
+ * @brief Get the current file socket for OSP NVT update.
+ *
+ * @return The path of the file socket for OSP NVT update.
+ */
+const gchar *
+get_osp_vt_update_socket ()
```
mattmundell: To keep in mind for the future: these functions don't contain SQL so should go in a manage_nvts.c file.  I have the idea that the upper layers (gvmd.c, gmpd.c, etc) shouldn't be importing *_sql*.c files, but only a single interface file like manage_tickets.h.

src/gvmd.c
timopollmeier: I've changed this in bdc83ddc70159dd1b722369e19c72e59ff3f43a7.

src/gvmd.c
timopollmeier: Would changing the line to a `g_warning` and removing the following `return EXIT_FAILURE;` produce the behavior you'd prefer?

src/manage_sql_nvts.c
timopollmeier: I was also considering this but I'd prefer to consistently move all the non-SQL helper functions in a separate PR.

src/gvmd.c
mattmundell: It wouldn't be enough, we would also have to try to get the OSP VT update socket in update_nvt_cache_retry if it is not already set.

mattmundell: Let's leave it for now.  Lots of other things to do.

src/manage_sql_nvts.c
mattmundell: Sounds good.

PR #733, opened 2019-09-02
Add CSV Hosts report format columns to match GSA

mattmundell:
**Checklist**:

- [ ] Tests

### DIFF
--- a/src/report_formats/CSV_Hosts/CSV_Hosts.xsl
+++ b/src/report_formats/CSV_Hosts/CSV_Hosts.xsl
@@ -143,12 +143,21 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
         <xsl:text>SSH</xsl:text>
       </xsl:if>
       <xsl:if test="detail[name/text() = 'Auth-SMB-Success']">
+        <xsl:if test="detail[name/text() = 'Auth-SSH-Success']">
+          <xsl:text> </xsl:text>
+        </xsl:if>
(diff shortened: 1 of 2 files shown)


src/report_formats/CSV_Hosts/CSV_Hosts.xsl
```diff
@@ -127,6 +127,30 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
...
+      <xsl:text>,</xsl:text>
+
+      <xsl:variable name="traceroute" select="detail[name/text() = 'traceroute']"/>
+      <xsl:value-of select="string-length($traceroute) - string-length(translate($traceroute, ',', ''))"/>
+
+      <xsl:text>,</xsl:text>
+      <xsl:if test="detail[name/text() = 'Auth-SSH-Success']">
+        <xsl:text>SSH</xsl:text>
+      </xsl:if>
+      <xsl:if test="detail[name/text() = 'Auth-SMB-Success']">
```
timopollmeier: Shouldn't there be some kind of separator like a space between the different methods in case there are multiple successful authentication methods?

mattmundell: I thought "SSHSMB" would be OK, but I guess a space would make them easier to parse.  In GSA it's just the same shield icon, repeated for each one.

mattmundell: Done in 3fdbf41.

Issue #540, opened 2019-05-10
[8.0.0] Unable to import an S/MIME certificate


tuxmaster5000:
### Expected behavior
That I can import an S/MIME certificate for email encryption.

### Current behavior
The error "Erroneous Certificate." will shown.

### Steps to reproduce
1. create new credentials 
2. set type to S/MIME
3. select the file

### GVM versions
**gsa:** (gsad --version)
Greenbone Security Assistant 8.0.0~git

**gvm:** (gvmd --version)
Greenbone Vulnerability Manager 8.0.0

**openvas-scanner:** (openvassd --version)
OpenVAS Scanner 6.0.0

**gvm-libs:**
10.0.0

**openvas-smb:**
1.0.5

### Environment
**Operating system:**
CentoOS 7.6

**Installation method / source:** (packages, source installation)
Git

### Logfiles
gvmd.log:
util gpgme:WARNING:2019-05-10 09h09.06 CEST:13257: gvm_gpg_import_from_string: key_str is not the expected type:  expected: 34, got 35

cfi-gb:
If this was a X.509 cert then https://github.com/greenbone/gvmd/pull/573 (Backport in https://github.com/greenbone/gvmd/pull/551) should have added support for this cert type.

tuxmaster5000:
I have added both patches (gvm-libs and gvmd).
But it will not work. Now I only get other error:

> util gpgme:WARNING:2019-06-03 09h25.42 CEST:10710: gvm_gpg_import_many_types_from_string: Import failed: End of file

For gvmd I have used the patch 551 and 231 for the gvm-lib.

cfi-gb:
https://github.com/greenbone/gvmd/pull/741 might help here. According to @timopollmeier there might be a few additional issues which needs to be sorted out.

tuxmaster5000:
Install gnupg2-smime will allow to install the x509 certificate.
But the mail encryption will fails with:
 d manage:WARNING:2019-09-09 07h26.15 CEST:11060: email_encrypt_smime: encryption failed
md manage:WARNING:2019-09-09 07h26.15 CEST:11060: email: S/MIME encryption failed

tuxmaster5000:
In debug mode of gvmd I can see this:

> util gpgme:WARNING:2019-09-09 07h30.26 CEST:5829: find_email_encryption_key: No suitable key found for foo@foo.foo
[...]

PR #741, opened 2019-09-06
Re-add removed install line. Added additional gpgsm prerequisite.

cfi-gb:
According to @timopollmeier ``gpgsm`` is required for S/MIME support. This might be related to the issue seen in e.g. #540.

I have also re-added the build dependency which seems to got dropped completely in #610 and added a few from the prerequisites mentioned / used in https://github.com/greenbone/openvas/blob/master/INSTALL.md#prerequisites-for-openvas

**Checklist**:

- [ ] Tests

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -30,6 +30,13 @@ Prerequisites for building tests:
 Please see the section "Prerequisites for Optional Features" below additional
 optional prerequisites.
 
+Install prerequisites on Debian GNU/Linux 'Stretch' 9:
+
+    apt-get install gcc cmake libglib2.0-dev libgnutls28-dev libpq-dev postgresql-server-dev-9.6 pkg-config libical-dev
+
+Optional prerequisites on Debian GNU/Linux 'Stretch' 9:
+
+    apt-get install gnutls-bin gpgsm
 
 ## Compiling Greenbone Vulnerability Manager
 


INSTALL.md
```diff
@@ -32,6 +35,13 @@ Prerequisites for building documentation:
 Please see the section "Prerequisites for Optional Features" below additional
 optional prerequisites.
 
+Install prerequisites on Debian GNU/Linux 'Stretch' 9:
+
+    apt-get install gcc cmake libglib2.0-dev libgnutls28-dev libpq-dev postgresql-server-dev-9.6 pkg-config libical-dev
+
+Optional prerequisites on Debian GNU/Linux 'Stretch' 9:
```
mattmundell: This is confusing because there's another section for optional features below.

I think it would be clearer if there was an apt command below each case above.  So, one
command below "Prerequisites for certificate generation", one below
"Prerequisites for S/MIME suport (e.g. email encryption)" etc.

cfi-gb: Indeed. It could make sense to raise an issue at https://github.com/greenbone/templates/issues to have one single base how such an instruction should look like.

Basically every component currently has different install instructions when it comes to prerequisites, quite often they are also just missing mandatory components like libical-dev or gpgsm.

cfi-gb: Raised https://github.com/greenbone/templates/issues/17. Suggestions on how this should look like are welcome.

davidak: How to install the dependencies on debian is quiet special and might be better in an additional file. Figuring that out is a task for people packaging the software for the distros. In best case, a user should only install a package and everything is working as expected. We can link to community maintained packages. See for example: https://repology.org/project/gvmd/versions

The INSTALL.md contains also many informations that belongs in the Usage section or external documentation, but that's a different topic.

cfi-gb: @davidak I have cross-posted this in https://github.com/greenbone/templates/issues/17, its probably the best to handle this in a more broader scope because this is not only affecting gvmd.

cfi-gb:
07c934df0d69bbc90c51f4cf28a61b8a713d7631 has now even added the Debian package name like:

> * xml_split (optional, lowers sync memory usage, Debian package: xml-twig-tools)

cfi-gb:
Replaced by #1242 with a different approach.

PR #743, opened 2019-09-08
Eliminate get_tag() and parse_tags().

janowagner:
The function get_tag() only parses the cvss_base_vector
from the incoming OSP XML for a VT and adds it to the nvti tags
string.

Using the new function nvt_add_tag() makes avoids the knowledge
about the internal syntax (pipe-separated list).

This last tag that is parsed is now located in function nvti_from_vt().

Also this removes the parsing for "cvss_base" inside the tags
string which happened in nvti_from_vt(). Actually it is not being
parsed anyway but rather computed from the vector.

This is done now directly after parsing the cvss_base_vector from
XML inside nvt_from_vt().

An in turn this make it possible to remove parse_tags() function,
another place where the internal knowledge about the pipe-separated
syntax for tags was present.

This requires https://github.com/greenbone/gvm-libs/pull/263

**Checklist**:

- [ ] Tests N/A

### DIFF
--- a/src/manage_sql_nvts.c
+++ b/src/manage_sql_nvts.c
@@ -1298,7 +1298,7 @@ nvti_from_vt (entity_t vt)
           cvss_base = g_strdup_printf ("%.1f", get_cvss_score_from_base_metrics
                                            (entity_text (severity)));
           nvti_set_cvss_base (nvti, cvss_base);
-	  g_free (cvss_base);
+          g_free (cvss_base);
         }
       else
         g_warning ("%s: no severity", __FUNCTION__);
--- a/src/manage_sql_nvts.c
+++ b/src/manage_sql_nvts.c
@@ -36,6 +36,8 @@
 #include <string.h>
 #include <unistd.h>
 
+#include <gvm/base/cvss.h> // for get_cvss_score_from_base_metrics
+
 #include "manage_sql.h"
 #include "manage_sql_nvts.h"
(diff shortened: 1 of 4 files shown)


src/manage_sql_nvts.c
```diff
@@ -36,6 +36,8 @@
 #include <string.h>
 #include <unistd.h>
 
+#include <gvm/base/cvss.h> // for get_cvss_score_from_base_metrics
```
mattmundell:
```suggestion
#include <gvm/base/cvss.h>
```

I don't think this kind of annotation is worth the overhead it introduces.  None of the Manager code does this.

src/manage_sql_nvts.c
```diff
@@ -1289,7 +1291,14 @@ nvti_from_vt (entity_t vt)
                       "cvss_base_v2")
               == 0))
         {
+          gchar * cvss_base;
+
           nvti_add_tag (nvti, "cvss_base_vector", entity_text (severity));
+
+          cvss_base = g_strdup_printf ("%.1f", get_cvss_score_from_base_metrics
+                                           (entity_text (severity)));
```
mattmundell: Looks like an indentation problem here.

mattmundell: Fix is not what I expected :) but let's wait for the clang-format which will surely
change it anyway.

PR #744, opened 2019-09-08
Ensure all NVT preference names are in the new format

mattmundell:
This uses a hardcoded list of NVT preference names to ensure that all
NVT preferences that are in the form OID:TYPE:NAME are converted
to the form OID:ID:TYPE:NAME.

OID:TYPE:NAME was the result of migrate_205_to_206 from #394.  Here
we introduce migrate_219_to_220.

**Checklist**:

- [ ] Tests


mattmundell:
> In my test run there were several timeout entries in nvt_preferences that were not replaced by the migrator.
> 
> The migrator or maybe the part of `manage_update_nvt_cache_osp` that calls `check_preference_names` should rename preferences like `[...]:entry:timeout` to `[...]:0:entry:timeout` (inserting the id `:0`).
> We already assume that the timeout has the id 0 elsewhere, so it should be okay.
> 
> I think updating the timeouts near the start of the migrator would be the best place to prevent warnings about unknown preference names.

Should be solved by b594763.

mattmundell:
And with b594763 there should no longer be unknown preferences (both old or new format) in nvt_preferences after the first sync.

PR #756, opened 2019-09-24
Add sensor relays and OSP sensors

timopollmeier:
This adds a way to replace connections to remote scanners with connections to relays and a new OSP sensor scanner type.

**Checklist**:

- Tests N/A


src/manage_sql.c
```diff
@@ -43755,6 +43755,109 @@ openvas_default_scanner_host ()
...
+    {
+      connection = osp_scanner_relay_connect (host, port, ca_pub);
+    }
+  else
+    {
+      connection = osp_connection_new (host, port, ca_pub, key_pub, key_priv);
+
+      if (connection == NULL)
+        {
+          if (is_unix_socket)
```
mattmundell: These if cases should be the other way round, right?

src/gvmd.c
```diff
@@ -1813,7 +1813,8 @@ gvmd (int argc, char** argv)
           "<number>" },
         { "relay-mapper", '\0', 0, G_OPTION_ARG_FILENAME,
           &relay_mapper,
-          "Excecutable for mapping scanner hosts to relays.",
+          "Executable for mapping scanner hosts to relays."
+          " Will check $PATH for the default. Use empty string to disable.",
```
mattmundell: "Will check $PATH for the default." is confusing to me.  I think it should mention
"gvm-relay-mapper" and it should be clear that $PATH is only checked if
--relay-mapper is not given.

src/manage.c
```diff
@@ -4873,6 +4878,218 @@ run_gmp_slave_task (task_t task, int from, char **report_id,
...
+      return 0;
+    }
+  else
+    {
+      gchar **cmd, *stdout_str, *stderr_str;
+      int exit_code = -1;
+      GError *err = NULL;
+      entity_t relay_entity;
+
+      stdout_str = NULL;
```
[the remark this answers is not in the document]
timopollmeier: This should be fixed in 2efd8bf8c4e63b8f3be70121c4502bcdbe7a8840

src/manage_sql.c
timopollmeier: Yes, the statements for the `if (is_unix_socket)` condition and the `else` should now be the other way with a268ef5ad18f58ef0c8d2a9f3d214c9a818f77f3.

src/gvmd.c
timopollmeier: If changed the documentation of the option in 3f9d7651739052044f2076462b193b00ee39c4b8.

timopollmeier:
In addition to the requested changes I've also added the use of OSP sensors for the `get_system_reports` command.

src/manage.c
```diff
@@ -5605,22 +5605,74 @@ credential_full_type (const char* abbreviation)
...
+get_osp_performance_string (scanner_t scanner, int start, int end,
+                            const char *titles, gchar **performance_str)
+{
+  char *host, *ca_pub, *key_pub, *key_priv;
+  int port;
+  osp_connection_t *connection;
+  osp_get_performance_opts_t opts;
+  gchar *error;
+
+  host = scanner_host (scanner);
```
mattmundell: These need to be freed.

src/manage.c
```diff
@@ -5605,22 +5605,74 @@ credential_full_type (const char* abbreviation)
...
+ * @brief Get system report types from a GMP slave.
  *
  * @param[in]   required_type  Single type to limit types to.
  * @param[out]  types          Types on success.
  * @param[out]  start          Actual start of types, which caller must free.
- * @param[out]  slave_id       ID of GMP slave.
+ * @param[out]  slave          GMP slave.
  *
- * @return 0 if successful, 2 failed to find slave, 3 unused, 4 could not
+ * @return 0 if successful, 2 unused, 3 unused, 4 could not
```
mattmundell: This "unused" business is confusing.  Can we either remove the return, or
describe it better, like "No slave given"?

src/manage.c
```diff
@@ -5778,30 +5828,56 @@ get_system_report_types (const char *required_type, gchar ***start,
...
+        return -1;
+      if (slave == 0)
+        return 2;
+
+      if (slave_type == SCANNER_TYPE_GMP)
+        return get_slave_system_report_types (required_type, start, types,
+                                              slave);
+      else
+        {
+          // Assume OSP scanner
```
mattmundell: Would like the declaration section before the comments, like we usually do.

src/manage.c
timopollmeier: The scanner info strings are now freed with cbd3f828a5845c8f4abe134c2a3e7821e19ba5a4.

src/manage.c
timopollmeier: I've removed the unused return values from the docs and made the functions return -1 if `slave` is 0 (because the check should be done by the caller) in fa06038573cd2d5a8dd6668db77ea32b84aba92d.

src/manage.c
timopollmeier: I've moved the comment in 2486eea576faad720d947181c2b476d357ca14be.

mattmundell: thank you.

PR #757, opened 2019-09-25
Accelerate VT feed update

janowagner:
Instead of a SQL query based on a copy of the nvts table, directly count
the new and modified VTs as they are loaded via OSP.
Use these values to determine about issuing events.

This reduces the update times from about 7 minutes to a few seconds.

**Checklist**:

- [ ] Tests N/A

### DIFF
--- a/src/manage_sql_nvts.c
+++ b/src/manage_sql_nvts.c
@@ -1000,7 +1000,7 @@ check_for_updated_nvts ()
  * @brief Set the NVT update check time in the meta table.
  */
 static void
-set_nvts_check_time ()
+set_nvts_check_time (int count_new, int count_modified)
 {
   if (sql_int ("SELECT NOT EXISTS (SELECT * FROM meta"
                "                   WHERE name = 'nvts_check_time')"))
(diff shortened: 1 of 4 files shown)


src/manage_sql_nvts.c
```diff
@@ -1302,6 +1302,14 @@ update_nvts_from_vts (entity_t *get_vts_response,
   entity_t vts, vt;
   entities_t children;
   GList *preferences;
+  int count_modified_vts = 0;
```
mattmundell: Would like to see these variables initialised after the declaration section, like
the other variables in the block.  And would like the two int variables on one
line like the entity_t variables.

src/manage_sql_nvts.c
```diff
@@ -1302,6 +1302,14 @@ update_nvts_from_vts (entity_t *get_vts_response,
...
+  int count_new_vts = 0;
+  time_t feed_version_epoch = 0;
+  struct tm tm;
+
+  memset (&tm, 0, sizeof (struct tm));
+  strptime (nvts_feed_version(), "%Y%m%d%H%M%S", &tm);
```
mattmundell: There's a space missing after version :)

PR #759, opened 2019-09-27
Allow "0" or empty credential_id in create_scanner

timopollmeier:
Previously the `create_scanner` command only accepted the `credential_id` missing completely
to set no credential. Now it also accepts "" and "0".
Also, the task config checks did not recognize the new scanner type.

**Checklist**:

- [] Tests N/A


bjoernricks:
Could you please update the API docs too?

PR #766, opened 2019-10-01
Get scanner preferences from OSP-OpenVAS scanner

timopollmeier:
The preferences of the OpenVAS scanner are expected to be in the
nvt_preferences table, so they have to be added there.

**Checklist**:

- Tests N/A

### DIFF
--- a/src/manage_sql_nvts.c
+++ b/src/manage_sql_nvts.c
@@ -1477,27 +1477,6 @@ manage_update_nvt_cache_osp (const gchar *update_socket)
       update_nvts_from_vts (&vts, scanner_feed_version);
       free_entity (vts);
 
-      /* Tell the main process to update its NVTi cache. */
-      sql ("UPDATE %s.meta SET value = 1 WHERE name = 'update_nvti_cache';",
-           sql_schema ());
-
-      g_info ("Updating VTs in database ... done (%i VTs).",
-              sql_int ("SELECT count (*) FROM nvts;"));
-
-      if (sql_int ("SELECT coalesce ((SELECT CAST (value AS INTEGER)"
-                   "                  FROM meta"
-                   "                  WHERE name = 'checked_preferences'),"
-                   "                 0);")
-          == 0)
-        {
-          check_preference_names ("config_preferences");
-          check_preference_names ("config_preferences_trash");
-
-          sql ("INSERT INTO meta (name, value)"
-               " VALUES ('checked_preferences', 1)"
-               " ON CONFLICT (name) DO UPDATE SET value = EXCLUDED.value;");
-        }
-
       /* Update scanner preferences */
       connection = osp_connection_new (update_socket, 0, NULL, NULL, NULL);
       if (!connection)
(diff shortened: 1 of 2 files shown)


src/manage_sql_nvts.c
```diff
@@ -1415,6 +1415,7 @@ manage_update_nvt_cache_osp (const gchar *update_socket)
 {
   osp_connection_t *connection;
   gchar *db_feed_version, *scanner_feed_version;
+  GSList *scanner_prefs;
```
mattmundell: This is only used in one block so it should be declared in that block.

src/manage_sql_nvts.c
```diff
@@ -1496,6 +1497,65 @@ manage_update_nvt_cache_osp (const gchar *update_socket)
...
+        {
+          GString *prefs_sql;
+          GSList *point;
+          int first;
+
+          point = scanner_prefs;
+          first = 1;
+
+          osp_connection_close (connection);
+          prefs_sql = g_string_new ("INSERT INTO nvt_preferences (name, value)"
```
mattmundell: I think this needs to handle the case where the scanner preference exists already.

src/manage_sql_nvts.c
```diff
@@ -1496,6 +1497,65 @@ manage_update_nvt_cache_osp (const gchar *update_socket)
                " VALUES ('checked_preferences', 1)"
                " ON CONFLICT (name) DO UPDATE SET value = EXCLUDED.value;");
         }
+
```
mattmundell: I think this should come before telling the main process to update its NVT cache.

src/manage_sql_nvts.c
timopollmeier: I've moved the variable in daa969520825068f563c0a03752b730ba7e0c4ab.

src/manage_sql_nvts.c
timopollmeier: Okay, I've changed it in 188e38bf9ea95129cc036d5c185823e9b25f8c8d.

src/manage_sql_nvts.c
```diff
@@ -1496,6 +1497,65 @@ manage_update_nvt_cache_osp (const gchar *update_socket)
...
+          osp_connection_close (connection);
+          prefs_sql = g_string_new ("INSERT INTO nvt_preferences (name, value)"
+                                    " VALUES");
+          while (point)
+            {
+              osp_param_t *param;
+              gchar *quoted_name, *quoted_value;
+
+              param = point->data;
+              quoted_name = sql_quote (osp_param_name (param));
```
jjnicola: Here you should use the param_id instead of the param_name

src/manage_sql_nvts.c
timopollmeier: This should be handled by the `ON CONFLICT (name) ...` clause and the I've run the enclosing if block multiple times without getting any duplicates.

src/manage_sql_nvts.c
timopollmeier: I've changed it in bf4b700804dd4c35c02609d54cdd5dd18869adf4.

src/manage_sql_nvts.c
mattmundell: Oh I missed that, nice.

PR #775, opened 2019-10-07
Extra TLS certificate info, rename last_collected to last_seen (master)

timopollmeier:
The sources of TLS certificates now contain the asset UUID of the host.
If the origin is a report they will also contain the report timestamp
and the task.
Also, the `last_collected` element and filter column are renamed to `last_seen`.

**Checklist**:

- Tests N/A


src/manage_sql_tls_certificates.c
```diff
@@ -1647,3 +1648,79 @@ add_tls_certificates_from_report_host (report_host_t report_host,
 
   return 0;
 }
+
+gchar *
+tls_certificate_origin_extra_xml (const char *origin_type,
```
mattmundell: The idea of the code organisation is that the "manage" layer does the data stuff,
and the "gmp" layer does all the GMP stuff (i.e. anything XML).  So getting the data
here is good, but generating the XML should be done in the GMP layer.

timopollmeier: I was expecting more special SQL when I started this function but I agree that it makes more sense in the GMP layer file, so I changed it in dc7a201f203faf217c8e395c842123b0cc3f50a3.

mattmundell: Now it feels like the data code in the function is in the wrong place, but it will have to do.

I think we should move away from using the resource types in the GMP layer.  So no report_t task_t etc and no find_* functions.  That's what's making me feel like this code is doing work on the data.

CHANGELOG.md
```diff
@@ -21,7 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- Added TLS certificates as a new resource type [#585](https://github.com/greenbone/gvmd/pull/585) [#663](https://github.com/greenbone/gvmd/pull/663) [#673](https://github.com/greenbone/gvmd/pull/673) [#674](https://github.com/greenbone/gvmd/pull/674) [#689](https://github.com/greenbone/gvmd/pull/689) [#695](https://github.com/greenbone/gvmd/pull/695) [#703](https://github.com/greenbone/gvmd/pull/703) [#728](https://github.com/greenbone/gvmd/pull/728) [#732](https://github.com/greenbone/gvmd/pull/732) [#750](https://github.com/greenbone/gvmd/pull/750) [#752](https://github.com/greenbone/gvmd/pull/752) [#775](https://github.com/greenbone/gvmd/pull/774)
```
mattmundell: Actually, the link on 775 is 774 on this.

PR #777, opened 2019-10-08
Use UTC in iCalendar strings (9.0)

timopollmeier:
To simplify the timezone handling, times like DTSTART are now
always converted to UTC and old schedules are migrated to this
new format.

**Checklist**:

- Tests N/A

### DIFF
--- a/src/manage_migrators.c
+++ b/src/manage_migrators.c
@@ -1481,6 +1481,88 @@ migrate_219_to_220 ()
   return 0;
 }
 
+/**
+ * @brief Convert iCalendar strings of schedules to new format for version 221.
+ *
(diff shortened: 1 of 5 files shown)


src/manage_migrators.c
```diff
@@ -1481,6 +1481,88 @@ migrate_219_to_220 ()
   return 0;
 }
 
+/**
+ * @brief Convert iCalendar strings of schedules to new format for version 221.
+ *
+ * @param[in]  trash  Whether to convert the trash table.
+ */
+static void convert_schedules_221 (gboolean trash)
```
mattmundell:
```suggestion
static void
convert_schedules_221 (gboolean trash)
```

PR #780, opened 2019-10-09
Return only one setting when setting_id is given (master)

timopollmeier:
The get_settings GMP command will now only return the setting value that
applies instead of both the user's and the global one.

**Checklist**:

- Tests N/A

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -58624,11 +58624,15 @@ init_setting_iterator (iterator_t *iterator, const char *uuid,
                    "SELECT %s"
                    " FROM settings"
                    " WHERE uuid = '%s'"
-                   " AND " ACL_GLOBAL_OR_USER_OWNS ()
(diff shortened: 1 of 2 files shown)


src/manage_sql.c
```diff
@@ -58624,11 +58624,15 @@ init_setting_iterator (iterator_t *iterator, const char *uuid,
                    "SELECT %s"
                    " FROM settings"
                    " WHERE uuid = '%s'"
-                   " AND " ACL_GLOBAL_OR_USER_OWNS ()
-                   /* Force the user's setting to come before the default. */
-                   " ORDER BY coalesce (owner, 0) DESC;",
+                   " AND (owner = (SELECT id FROM users WHERE uuid = '%s')"
```
mattmundell: Would be nice to not do ACL stuff manually in here, but at least it's consistent with the other branch of the 'if'.

PR #784, opened 2019-10-10
If the nvt preference is "file" type, encode it into Base64 format.

jjnicola:
Also the ssh key sent for authenticated scans will be encoded.

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -3785,10 +3785,12 @@ target_osp_ssh_credential (target_t target)
       if (strcmp (type, "usk") == 0)
         {
           const char *private_key = credential_iterator_private_key (&iter);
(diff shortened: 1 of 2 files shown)


src/manage.c
```diff
@@ -4091,6 +4092,8 @@ launch_osp_openvas_task (task_t task, target_t target, const char *scan_id,
               osp_value = g_strdup (split_value[0]);
               g_strfreev (split_value);
             }
+          else if (strcmp (type, "file") == 0)
+            osp_value = g_base64_encode ((guchar*) value, strlen(value));
```
mattmundell:
```suggestion
            osp_value = g_base64_encode ((guchar*) value, strlen (value));
```

PR #787, opened 2019-10-10
Add an option to log SQL timing info

mattmundell:
**Checklist**:

- [ ] Tests


mattmundell:
This was useful to understand how our logs compare to pg_stat_statements, but I don't think I'll use it again, and it add extra code, so closing.

PR #795, opened 2019-10-14
Count results while printing them, when ignoring pagination

mattmundell:
This skips some counting SQL by counting the results as they are being added
to the response XML.

This is possible because when we ignore_pagination we are getting all the
filtered results anyway, so we can do both the page and filtered counting as we
go through the results.

On my basic test report of 5000 results this saved about 212ms of a 1344ms
report, or 12%.  To be clear: that's the time spent in SQL for the GET_REPORTS
call made by the GSA Report page.

**Checklist**:

- [ ] Tests


src/manage_sql.c
```diff
@@ -30081,6 +30133,18 @@ print_report_xml_start (report_t report, report_t delta, task_t task,
...
+      if (count_filtered)
+        {
+          debugs = f_debugs;
+          holes = f_holes;
+          infos = f_infos;
+          logs = f_logs;
+          warnings = f_warnings;
+          false_positives = f_false_positives;
+          filtered_result_count = debugs + holes + infos + logs + warnings
+                                  + false_positives;
```
timopollmeier: Why are you overwriting the full counts with the filtered ones here instead of just adding the filtered counts for `filtered_result_count`?
Earlier you already get them with `report_counts_id_full (...)` and while the total is still correct after this the counts per level are not if there are any filter conditions limiting the results selection.

src/manage_sql.c
```diff
@@ -29483,6 +29483,8 @@ print_report_xml_start (report_t report, report_t delta, task_t task,
              "</delta>");
     }
 
+  count_filtered = (delta == 0 && ignore_pagination && get->details);
```
timopollmeier: This could perhaps also be used in any case where the number of rows is unlimited via a `rows=-1` in the filter and the "Max Rows" limit is disabled.
If "Max Rows" is enabled it could also be compared against the unfiltered total or some other upper bound that's quick to query to see if using the number of returned results could give the wrong count.

src/manage_sql.c
mattmundell: Oh I misunderstood what the report_counts_id part was doing.  Should be right with 
299fefa.  Thanks for catching.  Really want to clean this code up, but need to backport these PRs to gvmd 8.

src/manage_sql.c
mattmundell: Sounds like a good idea.  Would like to leave it for a later PR though.

PR #802, opened 2019-10-15
Update gvm-portnames-update

apilynx:
When issueing the command 
`gvm-portnames-update service-names-port-numbers.xml` 
to update iana service names we get:

> [e] Error: openvassd is not in the path, could not determine the Manager directory.

Simple name change (openvassd -> openvas). Openvassd was the scanner's name in previous versions of gvm.

**Checklist**:

* N/A Tests
* [x]  [CHANGELOG](https://github.com/greenbone/gvmd/blob/master/CHANGELOG.md) Entry


cfi-gb:
Seems this was never backported to the gvm-9.0 branch, did this in #1166

PR #812, opened 2019-10-16
Increase GMP version from 9.0 to 20.04

janowagner:
**Checklist**:

- [ ] Tests N/A


timopollmeier:
We have decided to keep the extra zero in version numbers wherever possible because the 20.04 format has already been used in various places referring to the next major release.

bjoernricks:
Shouldn't be derive the GMP version from PROJECT_VERSION_STRING now?

mattmundell:
> Shouldn't be derive the GMP version from PROJECT_VERSION_STRING now?

I merged so that we at least have the right version now.  Not against PROJECT_VERSION_STRING but it needs some consideration because we use the version to decide which CHANGEs to include in the output.

PR #814, opened 2019-10-17
Count all results directly for total_result_count

mattmundell:
**Checklist**:

- [ ] Tests


mattmundell:
This is very neat, but does not have any affect on performance and there may be some cases that I haven't thought of, so I'm closing.

Could come up again in future, especially if we want to move away from caching.

PR #817, opened 2019-10-17
Get REFs in result iterator, instead of in separate queries

mattmundell:
This gets the cert_bund and dfn_cert REF names from the result iterator,
instead of getting them using dedicated queries for each result.

With my large basic report of 50000 results the load time of the GSA report
page went from 72s to 55s for a saving of 20%.  While the total SQL time
has not changed much (maybe 5% faster) there are far fewer queries being
sent to the db, so we're saving on query transmission and setup time.

**Checklist**:

- [ ] Tests


mattmundell:
> For now, there's just one small issue I noticed: Results without any refs of a certain CERT now have a single ref of that CERT type without any id. To me this looks like an issue with generating the arrays.

Should be fixed now.

mattmundell:
> I think returning an empty array instead of NULL would be simpler and more like other cases where we iterate over some lists. Is there a particular reason you did it this way?
> 
> It's not a major issue for me, so I'll leave it up to you if you want to change this again or merge it as it is.

That's a good idea.  NULL means that the column was NULL.  Changed in https://github.com/greenbone/gvmd/pull/817/commits/5276ec28c01dd33b7364a09ec9f17e13b2c31f44.

PR #826, opened 2019-10-23
Get refs in the NVT SQL when updating nvti_cache

mattmundell:
This improves the gvmd startup time by about 10s.  Previously we ran a separate SQL statement for each NVT, to retrieve the refs.

Slow gvmd startup time really frustrates me during development.

I printed and diffed nvti_cache to confirm that it was the same before and after the patch.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -15489,10 +15489,17 @@ update_nvti_cache ()
 
   nvti_cache = nvtis_new ();
 
+  /* Because there are many NVTs and many refs it's slow to query the refs
+   * for each NVT.  So this query gets the NVTs and their refs at the same
+   * time.
+   *
+   * The NVT data is duplicated in the result of the query when there are
+   * multiple refs for an NVT, but the loop below uses nvtis_lookup to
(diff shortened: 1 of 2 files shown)


src/manage_sql.c
```diff
@@ -15490,37 +15490,41 @@ update_nvti_cache ()
   nvti_cache = nvtis_new ();
 
   init_iterator (&nvts,
-                 "SELECT oid, name, family, cvss_base, tag, solution, solution_type FROM nvts;");
+                 "SELECT nvts.oid, nvts.name, nvts.family, nvts.cvss_base,"
+                 "       nvts.tag, nvts.solution, nvts.solution_type,"
+                 /*      7 */
```
timopollmeier: I think this comment should be removed or replaced with one that briefly explains how the iterator combines the normal NVT info with the refs and why.

mattmundell: Good point, done in https://github.com/greenbone/gvmd/pull/826/commits/2e484026827f76d3e4954780a3cae4d2fce6c8aa.

PR #829, opened 2019-10-24
Add a script to migrate old exported configs

mattmundell:
**Checklist**:

- [ ] Tests N/A


mattmundell:
> The overall idea looks good but I wonder if it wouldn't be better if the preferences list were in an external file, with another script (or new part of the main one) to generate a new list from GMP or OSP.

I kept it to a single file so that it would be easy for people like support to use the script.

Why would they preference list need to be updated?  The pref names should not change until the old gvm versions are retired, if I understand right.  Perhaps an update would be needed when new preferences are added in the feed?

timopollmeier:
> I kept it to a single file so that it would be easy for people like support to use the script.
> 
> Why would they preference list need to be updated? The pref names should not change until the old gvm versions are retired, if I understand right. Perhaps an update would be needed when new preferences are added in the feed?

You're right, the names should not change for now, but as new NVTs are added to the feed, so are new preferences that have to be migrated to include an id.

bjoernricks:
>You're right, the names should not change for now, but as new NVTs are added to the feed, so are >new preferences that have to be migrated to include an id.
@cfi-gb can this ever happen? IMHO new nvts should not use name based preferences at all. Otherwise the feature has wasted a lot of our time!

timopollmeier:
@bjoernricks : New NVTs will have an id that is used instead of the name, but only in versions of gvmd that support it.

cfi-gb:
^ confirmed, as long as GOS 5.0 (GVM-10) is supported new VTs still need to use name based preferences (and are in addition using an id as a preparation for newer GVM versions). Such older GOS/GVM versions don't have any knowledge about the id based preferences.

mattmundell:
> The overall idea looks good but I wonder if it wouldn't be better if the preferences list were in an external file, with another script (or new part of the main one) to generate a new list from GMP or OSP.

I've added a new script, create-gvm-migrate-config, that generates gvm-migrate-config.

This way gvm-migrate-config is remains a single file, and can be updated to match a gvmd db at any time.

PR #835, opened 2019-10-25
Add host_id filter for tls_certificates

timopollmeier:
**Checklist**:

- Tests N/A

### DIFF
--- a/src/manage_sql_tls_certificates.c
+++ b/src/manage_sql_tls_certificates.c
@@ -205,7 +205,7 @@ tls_certificate_extra_where (const char *filter)
       quoted_id = sql_quote (host_id);
       g_string_append_printf
          (ret,
-          "AND (tls_certificates.id IN"
+          " AND (tls_certificates.id IN"
           " (WITH host_idents AS"
           "   (SELECT source_id AS ident_report_id, value AS ident_ip"
           "      FROM host_identifiers"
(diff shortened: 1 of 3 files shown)


src/manage_sql_tls_certificates.c
```diff
@@ -192,12 +192,41 @@ gchar *
...
+          "     WHERE host = (SELECT id FROM hosts"
+          "                   WHERE uuid='%s')"
+          "       AND name = 'ip')"
+          "  SELECT tls_certificate"
+          "    FROM tls_certificate_sources AS sources"
+          "    JOIN tls_certificate_origins AS origins"
+          "      ON origins.id = sources.origin"
+          "    JOIN tls_certificate_locations AS locations"
+          "      ON locations.id = sources.location"
+          "    WHERE origins.origin_id"
```
mattmundell: I'm struggling to get my head around this part.  It seem to me like you're selecting all tls_certificates that have any of the reports in host_idents as origin and at the same time any of the ip's in host_identifiers as location.  Doesn't that mean you will select a cert that report_a found at ip_2 and report_b found at ip_1, when the host_identifier was ip_1 by report_a?

timopollmeier: You're right, the conditions for the report id and the IP address need to apply to the same host_idents entry simultaneously.

timopollmeier: I've replaced the WHERE clause with another JOIN that has to match both the IP and the report id at the same time in 3c1ab3bae11b514110fe699d286701ee3f778aa3.

mattmundell: Ya, makes more sense to me now.

timopollmeier:
> Does the new filter term need to be added to the GMP docs?

Thanks for reminding me. I've added it in abb8828004aec2fe030ff6825143f19ee6802730.

PR #845, opened 2019-10-30
Allow use of public key auth in SCP alert (9.0)

timopollmeier:
The alert now not only works with password but also SSH key credentials.
Also, the error messages now also include the first lines of the sshpass / scp stderr output.

**Checklist**:

- Tests N/S

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -10177,6 +10177,66 @@ alert_script_exec (const char *alert_id, const char *command_args,
 }
 
 /**
+ * @brief Write data to a file for use by an alert script.
+ *
+ * @param[in]  directory      Base directory to create the file in
+ * @param[in]  filename       Filename without directory
+ * @param[in]  content        The file content
+ * @param[in]  content_size   Size of the file content
+ * @param[in]  description    Short file description for error messages
+ * @param[out] file_path      Return location of combined file path
+ *
+ * @return 0 success, -1 error
+ */
+static int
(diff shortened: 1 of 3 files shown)


src/manage_sql.c
```diff
@@ -10176,6 +10176,66 @@ alert_script_exec (const char *alert_id, const char *command_args,
...
+                       const char *content, gsize content_size,
+                       const char *description, gchar **file_path)
+{
+  gchar *path;
+  GError *error;
+
+  if (file_path)
+    *file_path = NULL;
+
+  /* Setup certificate file */
```
mattmundell: I guess this comment is for an old version because this function is not specifically about certs.

timopollmeier: Yes, this is an oversight from copying the code which I'll fix.

src/manage_sql.c
```diff
@@ -10176,6 +10176,66 @@ alert_script_exec (const char *alert_id, const char *command_args,
...
+    {
+      g_warning ("%s: Failed to write %s to file: %s",
+                 __FUNCTION__,
+                 description ? description : "extra data",
+                 error->message);
+      g_free (path);
+      return -1;
+    }
+
+  if (geteuid () == 0)
```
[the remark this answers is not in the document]
timopollmeier: The function is meant to handle files not covered by alert_script_exec.
For SCP alert_script_exec handles the password file, but not the private key, which has to be created and have its owner changed separately.
I'll add some comments to clarify this.

PR #847, opened 2019-10-30
Remove 1.3.6.1.4.1.25623.1.0.90011 from Discovery config (9.0)

timopollmeier:
The NVT is removed from the built-in "Discovery" config because it is
not available in every feed and only adds info already covered by other
NVTs.

**Checklist**:

- Tests N/A

### DIFF
--- a/src/manage_config_discovery.c
+++ b/src/manage_config_discovery.c
@@ -985,5 +985,12 @@ check_config_discovery (const char *uuid)
                             "yes",
                             TRUE);
 
+  sql ("DELETE FROM nvt_selectors"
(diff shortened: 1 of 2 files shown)


src/manage_config_discovery.c
```diff
@@ -985,5 +985,12 @@ check_config_discovery (const char *uuid)
                             "yes",
                             TRUE);
 
+  sql ("DELETE FROM nvt_selectors"
+       " WHERE family_or_nvt = '1.3.6.1.4.1.25623.1.0.90011'"
```
mattmundell: Not sure why github thinks the ' is a syntax error.

PR #848, opened 2019-10-31
Add tests of utils.c functions gvm_usleep and gvm_sleep

mattmundell:
**Checklist**:

- [x] Tests

### DIFF
--- a/src/utils_tests.c
+++ b/src/utils_tests.c
@@ -41,34 +41,22 @@ Ensure (utils, gvm_usleep_sleep_for_1)
 
 Ensure (utils, gvm_sleep_sleep_for_0)
 {
-  int (*prevent_inlining) (unsigned int);
-
-  prevent_inlining = gvm_sleep;
-  assert_that (prevent_inlining (0), is_equal_to (0));
+  assert_that (gvm_sleep (0), is_equal_to (0));
 }
 
 Ensure (utils, gvm_sleep_sleep_for_1ms)
 {
-  int (*prevent_inlining) (unsigned int);
-
-  prevent_inlining = gvm_sleep;
-  assert_that (prevent_inlining (0.001), is_equal_to (0));
+  assert_that (gvm_sleep (0.001), is_equal_to (0));
 }
 
 Ensure (utils, gvm_sleep_sleep_for_1us)
 {
-  int (*prevent_inlining) (unsigned int);
-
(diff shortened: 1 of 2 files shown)


src/utils_tests.c
```diff
@@ -25,6 +25,40 @@ Describe (utils);
...
+/* gvm_sleep */
+
+Ensure (utils, gvm_sleep_sleep_for_0)
+{
+  assert_that (gvm_usleep (0), is_equal_to (0));
+}
+
+Ensure (utils, gvm_sleep_sleep_for_1ms)
+{
+  assert_that (gvm_usleep (0.001), is_equal_to (0));
```
timopollmeier: Since gvm_sleep takes an integer parameter, giving it floating point values less than 1 doesn't really make sense.

src/utils_tests.c
```diff
@@ -25,6 +25,40 @@ Describe (utils);
...
+Ensure (utils, gvm_usleep_sleep_for_1)
+{
+  assert_that (gvm_usleep (1), is_equal_to (0));
+}
+
+/* gvm_sleep */
+
+Ensure (utils, gvm_sleep_sleep_for_0)
+{
+  assert_that (gvm_usleep (0), is_equal_to (0));
```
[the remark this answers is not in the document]
timopollmeier: Never mind, this seems to be resovled with the commit to prevent inlining.

src/utils_tests.c
mattmundell: As in your previous comment, I was calling the wrong function.

Didn't need to prevent inlining either.  Reverted in https://github.com/greenbone/gvmd/pull/848/commits/049d2fda9868f2baf08026225f762ee3e2bfacc4.

timopollmeier: I see what you're trying to do here, but both the gvm_usleep and the gvm_sleep function take an unsigned integer, so all the calls will still be effectively the same as `gvm_sleep (0)`.

In my opinion a more interesting test case would be something like `gvm_sleep (2)` and checking the system time if the sleep lasted at least 2 seconds.

mattmundell: Oh right I see what you mean.  Switched in https://github.com/greenbone/gvmd/pull/848/commits/049d2fda9868f2baf08026225f762ee3e2bfacc4.

PR #866, opened 2019-11-14
Extend command line options for managing scanners (master)

timopollmeier:
This is a port of #815, which extends the command line options `--get-scanners`, `--create-scanner` and `--modify-scanner` to provide more information and accept more scanner types.

**Checklist**:

- N/A Tests


bjoernricks:
@timopollmeier i really would like to use the term Greenbone-Sensor. IMHO OSP is only an implementation detail.

bjoernricks:
But I am fine with changing this in another PR.

PR #869, opened 2019-11-15
Allow resuming OSPd-based OpenVAS tasks (9.0)

timopollmeier:
The new scanner type previously always started a new scan and report.
With this PR the last stopped or interrupted report is reused and only unfinished hosts are rescanned, as it was done before switching to OSP.

**Checklist**:

- N/A Tests


src/alert_methods/SCP/alert.orig
```diff
@@ -0,0 +1,90 @@
...
+# Escalator method script: SCP.
+
+USERNAME=$1
+HOST=$2
+DEST=$3
+KNOWN_HOSTS=$4
+PRIVATE_KEY_FILE=$5
+PASSWORD_FILE=$6
+REPORT_FILE=$7
+<<<<<<< Updated upstream
```
mattmundell: Looks like this entire file was accidentally added.

timopollmeier: Yes, I've removed it in c421becab2ad86d506365fe36f3d4fa1d363aa7c

PR #872, opened 2019-11-20
Require Postgres 9.6 as a minimum (9.0)

mattmundell:
This makes Postgres 9.6 the minimum required version, to simplify
the performance oriented code that uses ON CONFLICT.

**Checklist**:

- [ ] Tests


mattmundell:
> It would be nice if we could check the postgres version via cmake too. Not sure if that is possible at the moment.

I hacked something in https://github.com/greenbone/gvmd/pull/872/commits/3fef338e1c35bb0f21d4e55e8393e9dc38308745.  Doesn't seem to be a clean way.

PR #890, opened 2019-12-04
Chunk CPE INSERTs, to reduce memory usage (9.0)

mattmundell:
**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_sql_secinfo.c
+++ b/src/manage_sql_secinfo.c
@@ -2009,6 +2009,7 @@ typedef struct
   GString *statement;      ///< Current statemet.
   int current_chunk_size;  ///< Number of rows in current statement.
   int max_chunk_size;      ///< Max number of rows per INSERT.
+  gchar *close_sql;        ///< SQL to close the statement.
 } inserts_t;
 
 /**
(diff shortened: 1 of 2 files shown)


src/manage_sql_secinfo.c
```diff
@@ -1995,6 +2000,137 @@ update_cert_bund_advisories (int last_cert_update)
...
+
+/**
+ * @brief Check size of current statement.
+ *
+ * @param[in]  inserts  Insert buffer.
+ *
+ * @return Whether this is the first value in the statement.
+ */
+static int
+inserts_check_size (inserts_t *inserts)
```
timopollmeier: As with `inserts_statement_close`, the function name should reflect that this is for CPEs or have a way to define the start of the new statement elsewhere. (I think the struct would be a good place to ensure it's consistent.)

src/manage_sql_secinfo.c
```diff
@@ -2245,29 +2377,13 @@ update_scap_cpes (int last_scap_update)
...
-           "     title = EXCLUDED.title,"
-           "     creation_time = EXCLUDED.creation_time,"
-           "     modification_time = EXCLUDED.modification_time,"
-           "     status = EXCLUDED.status,"
-           "     deprecated_by_id = EXCLUDED.deprecated_by_id,"
-           "     nvd_id = EXCLUDED.nvd_id",
-           inserts->str);
-      g_string_free (inserts, TRUE);
-    }
+  inserts_run (&inserts);
```
timopollmeier: I haven't made a comparison but wouldn't it be better to reduce the transaction size, not just the size of the statements within a single transaction?

src/manage_sql_secinfo.c
mattmundell: Yes, good idea.  I was just trying to be quick.  I've extended inserts_init so the caller can specify the wrapper SQL.

src/manage_sql_secinfo.c
mattmundell: It's an option now that the statements size is smaller, but it's very fast now (perhaps faster than it was with a single statement) and the Postgres memory usage looks to be very small (I didn't even see it in the top list, whereas it was number 1 at 20% of memory before).

timopollmeier: Okay, I guess if it's still a problem you can still try to optimize it further in another PR.

PR #898, opened 2019-12-10
Setup general task preferences to launch an osp openvas task.

jjnicola:
When a task is started, it sends now the max_checks, max_hosts, source_iface and hosts_ordering options to the scanner.

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -4072,6 +4072,7 @@ launch_osp_openvas_task (task_t task, target_t target, const char *scan_id,
   GHashTable *vts_hash_table;
   osp_credential_t *ssh_credential, *smb_credential, *esxi_credential;
   osp_credential_t *snmp_credential;
+  gchar *max_checks, *max_hosts, *source_iface, *hosts_ordering;
   GHashTable *scanner_options;
   int ret;
   config_t config;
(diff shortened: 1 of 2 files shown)


src/manage.c
```diff
@@ -4162,6 +4163,29 @@ launch_osp_openvas_task (task_t task, target_t target, const char *scan_id,
         }
     }
 
+  /* Setup general task preferences */
+  max_checks = task_preference_value (task, "max_checks");
+  g_hash_table_insert (scanner_options,
+                       g_strdup ("max_checks"),
+                       max_checks ? max_checks : MAX_CHECKS_DEFAULT);
```
mattmundell: MAX_CHECKS_DEFAULT should be g_strdup'ed because it will be g_freed when the hashtable is destroyed.

src/manage.c
```diff
@@ -4162,6 +4163,29 @@ launch_osp_openvas_task (task_t task, target_t target, const char *scan_id,
...
+
+  max_hosts = task_preference_value (task, "max_hosts");
+  g_hash_table_insert (scanner_options,
+                       g_strdup ("max_hosts"),
+                       max_hosts ? max_hosts : MAX_HOSTS_DEFAULT);
```
mattmundell: Same here.

src/manage.c
```diff
@@ -4162,6 +4163,25 @@ launch_osp_openvas_task (task_t task, target_t target, const char *scan_id,
         }
     }
 
+  /* Setup general task preferences */
+  max_checks = task_preference_value (task, "max_checks");
+  g_hash_table_insert (scanner_options, g_strdup ("max_checks"),
+                       max_checks ? max_checks : strdup (MAX_CHECKS_DEFAULT));
```
mattmundell: Why strdup?  max_checks is a gchar * so g_strdup will work, no?

PR #901, opened 2019-12-11
Split CPE file to reduce sync memory usage (9.0)

mattmundell:
This reduces the maximum amount of memory used by the CPE update during the SCAP sync, by parsing 40Mb chunks of the CPE dictionary at a time, instead of parsing the whole 162Mb dictionary.

My gvmd sync process is now maxing at about 3.5% of 16GB where it was 9% before.

gvmd creates the chunks itself, in a temp directory, using system to run a shell command.  Not the best, but g_spawn_sync has trouble here because we're already handling child signals (because we've forked a process to run the sync).  Also the XML wrapper replacements done by the shell command would be tedious to do in C.

This also add details to the proctitle during sync to make it easier to see where memory is spiking.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_sql_secinfo.c
+++ b/src/manage_sql_secinfo.c
@@ -171,6 +171,95 @@ increment_transaction_size (int* current_size)
     }
 }
 
+/**
+ * @brief Split a file.
+ *
+ * @param[in]  path  Path to file.
+ * @param[in]  size  Approx size of split files.
+ *
+ * @return Temp dir holding split files.
+ */
+static const gchar *
+split_xml_file (gchar *path, const gchar *size)
+{
+  int ret;
+  static gchar dir[] = "/tmp/gvmd-split-xml-file-XXXXXX";
+  gchar *previous_dir, *command;
+
+  if (mkdtemp (dir) == NULL)
+    {
+      g_warning ("%s: Failed to make temp dir: %s",
+                 __func__,
+                 strerror (errno));
+      return NULL;
+    }
+
+  previous_dir = getcwd (NULL, 0);
+  if (previous_dir == NULL)
+    {
+      g_warning ("%s: Failed to getcwd: %s",
+                 __FUNCTION__,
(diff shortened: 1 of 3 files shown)


src/manage_sql_secinfo.c
```diff
@@ -171,6 +171,95 @@ increment_transaction_size (int* current_size)
...
+    }
+
+  if (gvm_file_copy (path, "split.xml") == FALSE)
+    {
+      g_free (previous_dir);
+      return NULL;
+    }
+
+  command = g_strdup_printf
+             ("xml_split -s%s split.xml"
```
timopollmeier: The package containing the `xml_split` executable should be added to the installation prerequisites.

src/manage_sql_secinfo.c
```diff
@@ -171,6 +171,95 @@ increment_transaction_size (int* current_size)
...
+/**
+ * @brief Split a file.
+ *
+ * @param[in]  path  Path to file.
+ * @param[in]  size  Approx size of split files.
+ *
+ * @return Temp dir holding split files.
+ */
+static const gchar *
+split_xml_file (gchar *path, const gchar *size)
```
timopollmeier: Because the executed command contains some parts specific to the CPE dictionary, I think this should either be made more generic or renamed to clarify that it's for the CPEs.

src/manage_sql_secinfo.c
```diff
@@ -171,6 +171,95 @@ increment_transaction_size (int* current_size)
...
+      g_free (previous_dir);
+      return NULL;
+    }
+
+  if (gvm_file_copy (path, "split.xml") == FALSE)
+    {
+      g_free (previous_dir);
+      return NULL;
+    }
+
```
timopollmeier: It would be nice to have a short description of the different files produced in the split command:
What do the intial `split-*.xml` files contain? What should the head and tail consist of?

PR #913, opened 2019-12-18
Add prerequisite

qha:
Document dependency on gpgme.

**Checklist**:
- [ ] Tests n/a
- [ ] [CHANGELOG](https://github.com/greenbone/gvmd/blob/master/CHANGELOG.md) Entry n/a

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -22,6 +22,7 @@ Prerequisites:
 * sqlite3 library >= 3.8.3 or PostgreSQL database
 * pkg-config
 * libical >= 1.0.0
+* libgpgme
 
 Prerequisites for certificate generation:
 * GnuTLS certtool


qha:
I realise now that i hadn't gotten as far as i thought, perhaps a specific version of gpgme is required...

cfi-gb:
According to @timopollmeier in #741 the "gpgsm" is required, wondering if there is really also a requirement for the gpgme library?

qha:
[src/lsc_crypt.c](../blob/master/src/lsc_crypt.c) uses types gpgme_ctx_t, gpgme_key_t and various other things that sound like they would be provided by gpgme.
[src/CMakeLists.txt](../blob/master/src/CMakeLists.txt) looks for gpgme.

The latter is the main reason for my pull request.

Also trying to build against a packaged gvm with another gpgme version installed than the packager used causes the build to fail (thus my [earlier comment](#issuecomment-567093515)).

cfi-gb:
Not quite sure but gvmd has a hard dependency on gvm-libs in its [CMakeLists.txt](https://github.com/greenbone/gvmd/blob/master/src/CMakeLists.txt#L30-L33) as well as in the [INSTALL.md](https://github.com/greenbone/gvmd/blob/master/INSTALL.md):

> libgvm_base, libgvm_util, libgvm_osp, libgvm_gmp >= 20.4

This gvm-libs seems to be already covering the following requirement (note the libgpgme vs. gpgme) in its CMakeLists.txt:

> libgpgme >= 1.1.2

which could be applied here. The initial gpgsm i had assumed seems to be unrelated to this.

INSTALL.md
```diff
@@ -22,6 +22,7 @@ Prerequisites:
 * sqlite3 library >= 3.8.3 or PostgreSQL database
 * pkg-config
 * libical >= 1.0.0
+* gpgme
```
mattmundell: Thanks for the patch.

```suggestion
* libgpgme
```

I think this is better, at least to match gvm-libs.

I don't think we need a version number until we add one to the cmake check.

qha: Thanks for your suggestion! I have amended the commit.

qha: > I have amended the commit.

That seems to have been a bad idea, unfortunately by the time i realised that i had already discarded the only copy of the original commit i had access to from home over the holidays.

Sorry if this causes you extra work.

mattmundell: No problem, thanks.

PR #922, opened 2019-12-31
Remove agents

mattmundell:
**Checklist**:

- [ ] Tests


mattmundell:
Not sure what happened with the GMP doc diff.  Let me know if it's too hard to follow and I'll break it up.

src/manage_migrators.c
```diff
@@ -1633,6 +1633,77 @@ migrate_222_to_223 ()
...
+    {
+      sql_rollback ();
+      return -1;
+    }
+
+  /* Update the database. */
+
+  /* Agents were removed entirely. */
+
+  sql ("DELETE FROM tags"
```
timopollmeier: Deleting tags that are still referenced by `tag_resources` will violate the foreign key constraint. 

The simplest solution would be to just remove this `DELETE` because it would only cover tags that have no `resource_type` themselves or wrong ones.

If you still want to delete all the tags referencing agents, you would have to collect them before removing the `tag_resources` rows and then delete them later. 
Also, I think the condition should be ` WHERE id IN (SELECT ...)` instead of ` WHERE id = (SELECT ...)`.

mattmundell: I removed the DELETE in https://github.com/greenbone/gvmd/pull/922/commits/904e6ed17f253966c0f87a3e2d9cbc6a1abaf9c2.

And yes, you're right about the IN.

PR #928, opened 2020-01-07
Use temp tables to speed up migrate_213_to_214 (master)

timopollmeier:
This is a port of #911 to master.

Creating indexed, temporary tables makes it quicker to access host
details that are slow to get repeatedly for every TLS certificate,
presumably because of the "LIKE" conditions for the name or value field.

**Checklist**:

- Tests N/A


mattmundell:
Replaced by https://github.com/greenbone/gvmd/pull/930.

PR #923, opened 2020-01-01
Remove GMP COMMANDS

mattmundell:
**Checklist**:

- [ ] Tests
- [x] [CHANGELOG](https://github.com/greenbone/gvmd/blob/master/CHANGELOG.md) Entry

PR #931, opened 2020-01-09
Import predefined configs from feed at startup

mattmundell:
This startup check adds any new configs from the feed, but does not update existing configs.

**Checklist**:

- [ ] Tests


mattmundell:
Needs https://github.com/greenbone/gvmd/pull/923 first.

mattmundell:
This is ready for review after https://github.com/greenbone/gvmd/pull/923.

There are still some additions required but I think they should wait because this PR is already big:
 - handle config prefs that do not yet exist in the NVT db
 - allow --value="" for --modify-setting of "Import Feed Owner/Roles"
 - implement usage_type (currently just assumes "scan")

PR #938, opened 2020-01-15
Ignore Feed User's permissions when syncing from feed

mattmundell:
**Checklist**:

- [ ] Tests


src/manage_sql.c
```diff
@@ -49768,6 +49769,36 @@ create_permission (const char *name_arg, const char *comment,
...
+ * @param[in]   subject_id      UUID of subject.
+ * @param[out]  permission      Permission.
+ *
+ * @return 0 success, 2 failed to find subject, 3 failed to find resource,
+ *         5 error in resource, 6 error in subject, 7 error in name,
+ *         8 permission on permission, 9 permission does not accept resource,
+ *         99 permission denied, -1 internal error.
+ */
+int
+create_permission_no_acl (const char *name_arg, const char *comment,
```
mattmundell: Technically create_feed_config_permissions could have used create_permission_internal, but I thought it was better to match the naming used in the previous commit.  And perhaps exporting create_permission_internal is not the best and should be improved later anyway.

PR #946, opened 2020-01-16
Move port list management to a new file

mattmundell:
Almost all the port list internals are isolated in manage_sql_port_lists.c now.  More will happen when we add loading of port lists from the feed.

**Checklist**:

- [ ] Tests


src/manage_sql_port_lists.c
```diff
@@ -2334,6 +2334,25 @@ inherit_port_lists (user_t user, user_t inheritor)
        inheritor, user);
 }
 
+/**
+ * @brief Delete all port lists owner by a user.
```
jjnicola: Just a typo. owner -> owned

PR #947, opened 2020-01-17
Add target's alive test method before starting a scan

jjnicola:
Depends on pr greenbone/gvm-libs#312

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -4071,6 +4071,7 @@ launch_osp_openvas_task (task_t task, target_t target, const char *scan_id,
 {
   osp_connection_t *connection;
   char *hosts_str, *ports_str, *exclude_hosts_str, *finished_hosts_str;
+  int alive_test;
   osp_target_t *osp_target;
   GSList *osp_targets, *vts;
   GHashTable *vts_hash_table;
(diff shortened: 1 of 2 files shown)


src/manage.c
```diff
@@ -4103,6 +4104,10 @@ launch_osp_openvas_task (task_t task, target_t target, const char *scan_id,
   hosts_str = target_hosts (target);
   ports_str = target_port_range (target);
   exclude_hosts_str = target_exclude_hosts (target);
+
+  if (target_alive_tests (target) >= 0)
```
mattmundell: alive_test should be init'd if condition is false.

src/manage.c
```diff
@@ -4071,6 +4071,7 @@ launch_osp_openvas_task (task_t task, target_t target, const char *scan_id,
 {
   osp_connection_t *connection;
   char *hosts_str, *ports_str, *exclude_hosts_str, *finished_hosts_str;
+  int alive_test = 0;
```
mattmundell: For consistency, please keep initialisations out of the declaration block in gvmd.  Like how config and connection are done below.

I know some places in gvmd do it, but I've been trying to prevent new code from doing it because it got messy.

jjnicola:
@mattmundell Thanks for the reviews!

PR #948, opened 2020-01-21
Set run status only after getting OSP scan

timopollmeier:
This addresses issues with the start time being undefined while the
status is already set to "Running", e.g. during sensor scans.

**Checklist**:
- [] Tests N/A


src/manage.c
```diff
@@ -3560,6 +3560,7 @@ handle_osp_scan (task_t task, report_t report, const char *scan_id)
   char *host, *ca_pub, *key_pub, *key_priv;
   int rc, port;
   scanner_t scanner;
+  gboolean started = FALSE;
```
mattmundell: I know I'm nitpicking but could you init this in the group below for consistency.

timopollmeier: Done in 2d17b0139e45b59a70ca5aeca86412e15e86db79.

PR #949, opened 2020-01-21
Fix get_system_reports for GMP scanners (9.0)

timopollmeier:
The get_system_report_types function did not fetch the scanner type so
it erroneously assumed it was handling an OSP scanner instead.

**Checklist**:
- [] Tests N/A

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -6216,7 +6216,6 @@ get_system_report_types (const char *required_type, gchar ***start,
       scanner_type_t slave_type;
 
       slave = 0;
-      slave_type = SCANNER_TYPE_NONE;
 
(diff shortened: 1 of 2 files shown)


src/manage.c
```diff
@@ -6223,6 +6223,7 @@ get_system_report_types (const char *required_type, gchar ***start,
       if (slave == 0)
         return 2;
 
+      slave_type = scanner_type (slave);
```
mattmundell: Build is failing because it's being initialised twice.

timopollmeier: I've removed the redundant init before this in 6c22de6f528caac56dcb97d1deb7be3de6a70009.

PR #989, opened 2020-02-21
Move report format dirs when inheriting user

mattmundell:
**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_sql_report_formats.c
+++ b/src/manage_sql_report_formats.c
@@ -3919,8 +3919,9 @@ empty_trashcan_report_formats ()
 /**
  * @brief Change ownership of report formats, for user deletion.
  *
- * @param[in]  user       Current owner.
- * @param[in]  inheritor  New owner.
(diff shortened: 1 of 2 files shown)


src/manage_sql_report_formats.c
```diff
@@ -3916,6 +3916,59 @@ empty_trashcan_report_formats ()
...
+    }
+
+  inheritor_id = user_uuid (inheritor);
+  if (inheritor_id == NULL)
+    {
+      g_warning ("%s: inheritor_id NULL, skipping report format dir", __func__);
+      return;
+    }
+
+  old_dir = g_build_filename (GVMD_STATE_DIR,
```
mattmundell: This will be wrong for old predefined report formats.  Their owner is changed in migrate_predefined_report_formats, but their files will still be in the old special dir (and not in the owner's report format dir).  I'll fix this in a new PR.

PR #991, opened 2020-02-24
Print UUIDs in --get-users when --verbose given

mattmundell:
**Checklist**:

- [ ] Tests


mattmundell:
> Not sure if --verbose is really required but I guess you just wanted to not break existing scripts relying on the old/current output of --get-users

Yes, I would prefer without --verbose, but I don't want trouble.

PR #995, opened 2020-02-25
Always run manage_option_cleanup

mattmundell:
Every option handler must run manage_option_setup at the beginning, and manage_option_cleanup when it exits.  This adds a few missing cases.

It's a minor issue because the process is about to exit, but better to have it consistent.

**Checklist**:

- [ ] Tests


mattmundell:
> Could you add some extra information why it is necessary to always run manage_option_cleanup?

Updated description, hope that clears it up.

PR #996, opened 2020-02-25
Remove header comments

mattmundell:
We do not use this type of comment for all include statements, and they add
maintenance overhead without adding much benefit, so better to remove them.
The reference for gvm/base/networking.h was out of date, for instance.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage.h
+++ b/src/manage.h
@@ -33,15 +33,13 @@
 #include <glib.h>
 #include <gnutls/gnutls.h>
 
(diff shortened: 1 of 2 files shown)


mattmundell:
> I've used such comments in gsad as a hint why this include was added. Many includes in gsad c files have been unnecessary in the past and it is difficult to spot when they can be removed.

I understand there are some situations where it can help, but I don't want to maintain this in gvmd.  I think it's simpler to remove the include and try compile.

PR #998, opened 2020-03-03
Add --rebuild and --update

mattmundell:
This adds back the --rebuild and --update options.

These should not be required in normal operation, and are added only to force an update/rebuild when errors have resulted in mismatches between the gvmd and ospd-openvas NVTs.

--update will get all NVTs from the scanner, and update them.

--rebuild will do the same, but will remove all the gvmd NVTs first.

Note that these are different from gvmd's automatic sync of NVTs, which will only consider NVTs that have been created/modified since the last sync.

**Checklist**:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Automatically load predefined report formats from the feed [#968](https://github.com/greenbone/gvmd/pull/968) [#970](https://github.com/greenbone/gvmd/pull/970)
 - Print UUIDs in --get-users when --verbose given [#991](https://github.com/greenbone/gvmd/pull/991)
 - Add --get-roles [#992](https://github.com/greenbone/gvmd/pull/992)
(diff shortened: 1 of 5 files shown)


bjoernricks:
Shouldn't we just use one command? I don't know why we should not clean up the db before requesting all nvts. I am really sure we always want to clear the db if the command is issues in GOS.

mattmundell:
> The changes look okay so far, but I think rebuilding the NVTs should also rebuild the
> `result_nvts` table (maybe using the `cleanup_result_nvts` function) while the update option could also run the cleanup.

I'm not sure why rebuilding result_nvts would be necessary.  result_nvts associates IDs with all the NVT OIDs that occurr in results, to improve performance, regardless of whether the OIDs appear in the nvts table.

And we already have cleanup_result_nvts, so the user could call that after --rebuild/update if they needed to.

mattmundell:
> Shouldn't we just use one command? I don't know why we should not clean up the db before requesting all nvts. I am really sure we always want to clear the db if the command is issues in GOS.

Done in a369c3a.

bjoernricks:
Thanks a lot!

PR #1003, opened 2020-03-06
Improve handling of removed NVT prefs

mattmundell:
**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_sql_nvts.c
+++ b/src/manage_sql_nvts.c
@@ -1819,6 +1819,7 @@ update_or_rebuild (int update)
   if (update == 0)
     {
       sql ("TRUNCATE nvts;");
+      sql ("TRUNCATE nvt_preferences;");
       set_nvts_feed_version ("0");
     }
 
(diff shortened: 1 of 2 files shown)


src/manage_sql_nvts.c
```diff
@@ -1437,19 +1436,21 @@ update_nvts_from_vts (entity_t *get_vts_response,
 
       insert_nvt (nvti);
 
+      preferences = NULL;
       if (update_preferences_from_vt (vt, nvti_oid (nvti), &preferences))
         {
           sql_rollback ();
           return;
         }
+      sql ("DELETE FROM nvt_preferences WHERE name LIKE '%s:%%';",
```
timopollmeier: Would it help with the performance of the first update to check outside the loop if there are any preferences and skip this `DELETE` if that's the case? Otherwise it would be run without any effect for every NVT.

mattmundell: Usually this only considers NVTs added/updated since the last sync, so I don't think the DELETEs will be a problem.  Also not sure if there's an easy SQL statement to check this.

But something to keep in mind if we see performance problems.

mattmundell:
> If we now allow removing NVT preferences I also wonder if we should check for and warn about config
> preferences that are no longer valid.

Hmm, yeah, I'd have to look at that.  Will merge this in the mean time.

PR #1022, opened 2020-03-23
Use nvti_qod instead of the old nvti_get_tag()

jjnicola:
Depends on greenbone/gvm-libs#321

### DIFF
--- a/src/manage_sql_nvts.c
+++ b/src/manage_sql_nvts.c
@@ -296,15 +296,13 @@ insert_nvt (const nvti_t *nvti)
 
   quoted_cvss_base = sql_quote (nvti_cvss_base (nvti) ? nvti_cvss_base (nvti) : "");
 
-  qod_str = nvti_get_tag (nvti, "qod");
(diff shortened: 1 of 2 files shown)


src/manage_sql_nvts.c
```diff
@@ -1238,8 +1236,15 @@ nvti_from_vt (entity_t vt)
...
   if (detection)
     {
+      const gchar *qod;
+
       nvti_set_detection (nvti, entity_text (detection));
-      nvti_set_qod_type (nvti, entity_attribute (detection, "qod_type"));
+
+      qod = entity_attribute (detection, "qod");
+      if (qod == NULL)
+        nvti_set_qod_type (nvti, entity_attribute (detection, "qod_type"));
```
mattmundell: You don't need to set qod_type when qod is given?

jjnicola: It is one or the other. There should not be both at the same time. OSPD also considers one of them, not both.

PR #1024, opened 2020-03-25
Perform integrity check of VTs after updates

timopollmeier:
If OSPd-OpenVAS sends a sha256_hash attribute in the vts element, the
SHA-256 hash is also calculated for the NVTs in the database and
compared after updating.

**Checklist**:

- Tests N/A

### DIFF
--- a/src/gvmd.c
+++ b/src/gvmd.c
@@ -1075,7 +1075,8 @@ handle_sigabrt_simple (int signal)
  *
  * @param[in]  update_socket  UNIX socket for contacting openvas-ospd.
  *
- * @return 0 success.
+ * @return 0 success, -1 error, 1 VT integrity check failed,
+ *         2 scanner still loading.
  */
 static int
 update_nvt_cache_osp (const gchar *update_socket)
--- a/src/gvmd.c
+++ b/src/gvmd.c
@@ -1119,7 +1119,22 @@ update_nvt_cache_retry ()
           const char *osp_update_socket;
           osp_update_socket = get_osp_vt_update_socket ();
           if (osp_update_socket)
-            exit (update_nvt_cache_osp (osp_update_socket));
+            {
+              int ret;
+
+              ret = update_nvt_cache_osp (osp_update_socket);
+              if (ret)
+                {
+                  g_message ("Rebuilding NVTs because integrity check failed");
+                  ret = update_or_rebuild_nvts (0);
+                  if (ret)
+                    g_warning ("%s: rebuild failed", __func__);
+                  else
+                    g_message ("%s: rebuild successful", __func__);
+                }
+
+              exit (ret);
+            }
           else
             {
               g_warning ("%s: No OSP VT update socket set", __func__);
@@ -2252,7 +2267,10 @@ gvmd (int argc, char** argv)
       ret = manage_rebuild (log_config, database);
       log_config_free ();
       if (ret)
-        return EXIT_FAILURE;
+        {
+          printf ("Failed to rebuild NVT cache.\n");
+          return EXIT_FAILURE;
+        }
       return EXIT_SUCCESS;
     }
 
--- a/src/gvmd.c
+++ b/src/gvmd.c
@@ -1075,8 +1075,7 @@ handle_sigabrt_simple (int signal)
  *
  * @param[in]  update_socket  UNIX socket for contacting openvas-ospd.
  *
- * @return 0 success, -1 error, 1 VT integrity check failed,
- *         2 scanner still loading.
+ * @return 0 success, -1 error, 1 VT integrity check failed.
  */
 static int
 update_nvt_cache_osp (const gchar *update_socket)
@@ -1123,7 +1122,7 @@ update_nvt_cache_retry ()
               int ret;
 
               ret = update_nvt_cache_osp (osp_update_socket);
-              if (ret)
+              if (ret == 1)
                 {
                   g_message ("Rebuilding NVTs because integrity check failed");
                   ret = update_or_rebuild_nvts (0);
--- a/src/manage.c
+++ b/src/manage.c
@@ -8597,8 +8597,7 @@ gvm_migrate_secinfo (int feed_type)
  *
  * @param[in]  update_socket  Socket to use to contact ospd-openvas scanner.
  *
- * @return 0 success, -1 error, 1 VT integrity check failed,
- *         2 scanner still loading.
+ * @return 0 success, -1 error, 1 VT integrity check failed.
  */
 int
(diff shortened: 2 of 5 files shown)


src/manage_sql_nvts.c
```diff
@@ -1751,9 +1792,10 @@ manage_update_nvt_cache_osp (const gchar *update_socket)
       g_info ("OSP service has newer VT status (version %s) than in database (version %s, %i VTs). Starting update ...",
               scanner_feed_version, db_feed_version, sql_int ("SELECT count (*) FROM nvts;"));
 
-      if (update_nvt_cache_osp (update_socket, db_feed_version,
-                                scanner_feed_version))
-        return -1;
+      ret = update_nvt_cache_osp (update_socket, db_feed_version,
```
mattmundell: Just return directly?

src/manage_sql_nvts.c
```diff
@@ -1455,6 +1461,35 @@ update_nvts_from_vts (entity_t *get_vts_response,
...
+                      "                      'sha256'),"
+                      "               'hex')"
+                      " FROM str;");
+
+      if (strcmp (osp_vt_hash, db_vts_hash ? db_vts_hash : ""))
+        {
+          g_warning ("%s: SHA-256 hash of the VTs in the database does not"
+                     " match the one from the scanner.",
+                     __func__);
+          g_debug ("%s: Hash from OSP: %s, from database: %s",
```
mattmundell: Any reason for not putting this in the warning?  Could be useful to support.

src/gvmd.c
```diff
@@ -1075,7 +1075,8 @@ handle_sigabrt_simple (int signal)
  *
  * @param[in]  update_socket  UNIX socket for contacting openvas-ospd.
  *
- * @return 0 success.
+ * @return 0 success, -1 error, 1 VT integrity check failed,
+ *         2 scanner still loading.
```
mattmundell: This is not actually true, we don't get this 2 return anymore.

src/manage.c
```diff
@@ -8597,7 +8597,8 @@ gvm_migrate_secinfo (int feed_type)
  *
  * @param[in]  update_socket  Socket to use to contact ospd-openvas scanner.
  *
- * @return 0 success, -1 error, 2 scanner still loading.
+ * @return 0 success, -1 error, 1 VT integrity check failed,
+ *         2 scanner still loading.
```
mattmundell: Same here.

src/manage_sql_nvts.c
```diff
@@ -1700,7 +1738,8 @@ update_nvt_cache_osp (const gchar *update_socket, gchar *db_feed_version,
  *
  * @param[in]  update_socket  Socket to use to contact ospd-openvas scanner.
  *
- * @return 0 success, -1 error, 2 scanner still loading.
+ * @return 0 success, -1 error, 1 VT integrity check failed,
+ *         2 scanner still loading.
```
mattmundell: And same here.

mattmundell: I know manage_update_nvt_cache_osp doc has this return but it's not true there either, and I don't see it happening anytime soon, so I don't think we should spread that.

src/manage_sql_nvts.c
```diff
@@ -1455,6 +1461,35 @@ update_nvts_from_vts (entity_t *get_vts_response,
           count_new_vts, count_modified_vts);
 
   sql_commit ();
+
+  if (osp_vt_hash && strcmp (osp_vt_hash, ""))
+    {
+      char *db_vts_hash;
+
+      db_vts_hash
```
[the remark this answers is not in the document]
timopollmeier: Should be fixed in aaa7768402a4371ccf407ea221647144105817d4

src/manage_sql_nvts.c
```diff
@@ -1390,9 +1393,12 @@ update_nvts_from_vts (entity_t *get_vts_response,
   if (vts == NULL)
     {
       g_warning ("%s: VTS missing", __func__);
-      return;
+      return -1;
     }
 
+  osp_vt_hash
```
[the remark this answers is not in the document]
timopollmeier: Yes, I've changed it in aaa7768402a4371ccf407ea221647144105817d4

src/manage_sql_nvts.c
timopollmeier: This was left from an earlier approach where the return code was handled here.
Changed in aaa7768402a4371ccf407ea221647144105817d4

src/manage_sql_nvts.c
timopollmeier: I thought this may only be useful for debugging but I don't mind combining the two, so
I've done it in aaa7768402a4371ccf407ea221647144105817d4

src/manage_sql_nvts.c
timopollmeier: I think @jjnicola recently changed OSPd(-OpenVAS) to return an error if it is still loading
but it would be cleaner if we handled this in a separate PR that also considers other uses of OSP.

src/manage_sql_nvts.c
```diff
@@ -1477,13 +1476,15 @@ update_nvts_from_vts (entity_t *get_vts_response,
...
       if (strcmp (osp_vt_hash, db_vts_hash ? db_vts_hash : ""))
         {
-          g_warning ("%s: SHA-256 hash of the VTs in the database does not"
-                     " match the one from the scanner.",
-                     __func__);
-          g_debug ("%s: Hash from OSP: %s, from database: %s",
-                   __func__, osp_vt_hash, db_vts_hash);
+          g_warning ("%s: SHA-256 hash of the VTs in the database (%s)"
+                     " does not match the one from the scanner (%s).",
+                     __func__, osp_vt_hash, db_vts_hash);
```
mattmundell: Thanks.  The hash args should be swapped.

src/manage_sql_nvts.c
mattmundell: Ah, OK so they could have stayed.  Thanks.

src/manage_sql_nvts.c
timopollmeier: I fixed this in 99cbf9483b36a6da684a66a729c0914ba8b1b531

timopollmeier:
> Looks good to me. You can find the PR for including the preferences here [greenbone/ospd#236](https://github.com/greenbone/ospd/pull/236) but i guess it needs some additional work in ospd-openvas too for activating the calculation.

I've changed the string used to calculate the hash to include the preferences in
f46a338c28b06e60125b200509f1fe19689ad476.

PR #1041, opened 2020-04-09
Remove unused port names facility

mattmundell:
This was used by the OTP code in gvmd.

**Checklist**:

- [ ] Tests


mattmundell:
> This looks okay to me but I'm not sure if this should be removed or if the port names should be added to the OSPd-OpenVAS results.

I don't think we should add the port names to the OSP results.

This was changing "port" table "results" from "20/tcp" to "20/tcp (IANA: ftp-data)".  This is not a good place to store the IANA name.  Plus, I'm not sure gvmd should be providing this service at all.  GSA could even just keep a list of IANA names internally, it's not very large.

And we've released GVM 6 already which does not use OTP, so it already does not provide IANA names.

mattmundell:
> This is not a good place to store the IANA name.

An example of where this is already problematic: when applying overrides we are comparing the override's port with the result's port.  We compare directly with = in SQL so the string must be identical.  If the override is "20/tcp" and the result is "20/tcp (IANA: ftp-data)" then this fails.  It's too much complexity to try keep these in sync.  We could change the comparison, but also adds complexity.

PR #1045, opened 2020-04-14
Improve validation of note and override ports

mattmundell:
This is the port validation used by GMP CREATE/MODIFY_NOTE/OVERRIDE.

**Checklist**:

- [ ] Tests


src/manage_sql.c
```diff
@@ -31081,39 +31081,40 @@ validate_port (const char *port)
...
 
   if (!port)
     return 1;
 
-  if (g_str_has_prefix (port, "cpe:"))
-    return 0;
-
-  if (strncmp ("general/", port, 8) == 0)
-    return 0;
+  /* "cpe:abc", "general/tcp", "20/upd"
```
timopollmeier, line 31099:
```suggestion
  /* "cpe:abc", "general/tcp", "20/udp"
```

src/manage_sql.c
```diff
@@ -31081,39 +31081,40 @@ validate_port (const char *port)
...
+   *
+   * The , and ; is to stop users from entering lists of ports.
+   * CPE doesn't use them because seems like they're valid in CPEs. */
```
timopollmeier, line 31102: This could be a bit clearer that it's referring to the restriction not containing `,` and `;` for CPEs.

Also, whether the characters are allowed without percent escaping depends on the version of the specification.
If I understand it correctly, 2.2 (the version we currently use) requires them to be escaped.
(Chapter 5.4. in https://cpe.mitre.org/specification/2.2/cpe-specification_2.2.pdf)

src/manage_sql.c
mattmundell: Fixed in 917206a.

src/manage_sql.c
mattmundell: Improved in 5ce9e79.

I don't think we need to worry about escaping at this point.  Ultimately we could use a full CPE regex (saw one at https://csrc.nist.gov/schema/cpe/2.3/cpe-naming_2.3.xsd) but it might prove too restrictive (eg won't allow for errors in the shipped CPEs or mixed CPE versions).  Anyway, a topic for a future PR, I'd say.

PR #1048, opened 2020-04-16
Fix version field of OVAL definitions (8.0)

timopollmeier:
The update_ovaldef_xml function now adds the actual version number to
the SQL where it mistakenly added the address of the version string.

**Checklist**:
- [ ] Tests N/A


timopollmeier:
I didn't like using the string as it is but with the validation it should be safe,
so I'll create a backport of #874 instead.

timopollmeier:
New PR: #1053

PR #1050, opened 2020-04-16
Add --rebuild-scap option (9.0)

timopollmeier:
This option allows rebuilding SCAP data of a given type.
Currently only OVAL definitions are supported.

**Checklist**:
- Tests N/A

### DIFF
--- a/src/manage_sql_secinfo.c
+++ b/src/manage_sql_secinfo.c
@@ -5201,6 +5201,10 @@ rebuild_scap (const char *type)
   int lockfile;
 
   ret = open_secinfo_lockfile ("gvm-sync-scap", &lockfile);
+  if (ret == 1)
+    return 2;
+  else if (ret)
+    return -1;
 
   if (strcasecmp (type, "ovaldefs") == 0
       || strcasecmp (type, "ovaldef") == 0)
@@ -5238,9 +5242,7 @@ rebuild_scap (const char *type)
  * @param[in]  database    Location of manage database.
  * @param[in]  type        The type of SCAP info to rebuild.
 
- * @return 0 success, -1 error, -2 main database is wrong version,
- *         -3 main database needs to be initialised from server,
- *         -5 sync currently running.
+ * @return 0 success, -1 error.
  */
 int
 manage_rebuild_scap (GSList *log_config, const gchar *database,
@@ -5252,7 +5254,7 @@ manage_rebuild_scap (GSList *log_config, const gchar *database,
 
   ret = manage_option_setup (log_config, database);
   if (ret)
-    return ret;
+    return -1;
 
   if (manage_update_scap_db_init ())
     goto fail;
@@ -5277,6 +5279,11 @@ manage_rebuild_scap (GSList *log_config, const gchar *database,
   if (ret == 1)
     {
       printf ("Type must be 'ovaldefs'.\n");
+      goto fail;
+    }
+  else if (ret == 2)
+    {
+      printf ("SCAP sync is currently running.\n");
       goto fail;
     }
   else if (ret)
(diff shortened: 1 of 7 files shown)


src/manage_sql_secinfo.c
```diff
@@ -5123,6 +5123,128 @@ manage_sync_scap (sigset_t *sigmask_current)
...
+  int updated_scap_ovaldefs, updated_scap_cpes, updated_scap_cves;
+
+  // updated_scap_ovaldefs = 0; // initial value currently unused
+  updated_scap_cpes = 0;
+  updated_scap_cves = 0;
+
+  if (strcasecmp (type, "ovaldefs") == 0
+      || strcasecmp (type, "ovaldef") == 0)
+    {
+      g_debug ("%s: rebuilding ovaldefs", __FUNCTION__);
```
mattmundell, line 5145:
```suggestion
      g_debug ("%s: rebuilding ovaldefs", __func__);
```

src/manage_sql_secinfo.c
```diff
@@ -5123,6 +5123,128 @@ manage_sync_scap (sigset_t *sigmask_current)
...
+    goto fail;
+
+  if (manage_scap_db_exists ())
+    {
+      if (check_scap_db_version ())
+        goto fail;
+    }
+  else
+    {
+      g_info ("%s: Initializing SCAP database", __FUNCTION__);
```
mattmundell, line 5220:
```suggestion
      g_info ("%s: Initializing SCAP database", __func__);
```

src/manage_sql_secinfo.c
```diff
@@ -5123,6 +5123,128 @@ manage_sync_scap (sigset_t *sigmask_current)
...
+
+      if (manage_db_init ("scap"))
+        {
+          g_warning ("%s: Could not initialize SCAP database", __FUNCTION__);
```
mattmundell, line 5224:
```suggestion
          g_warning ("%s: Could not initialize SCAP database", __func__);
```

src/manage_sql_secinfo.c
```diff
@@ -5123,6 +5123,128 @@ manage_sync_scap (sigset_t *sigmask_current)
...
+      g_debug ("%s: rebuilding ovaldefs", __FUNCTION__);
+      sql ("DELETE FROM ovalfiles");
+      sql ("DELETE FROM affected_ovaldefs");
+      sql ("DELETE FROM ovaldefs");
+
+      updated_scap_ovaldefs = update_scap_ovaldefs (0, 0 /* Feed data. */);
+      if (updated_scap_ovaldefs == -1)
+        return -1;
+
+      switch (update_scap_ovaldefs (0, 1 /* Private data. */))
```
mattmundell, line 5154: I don't like that these update functions are called here and in sync_scap.  It means there are two places to keep track of what happens during sync, and also that the syncs behave slightly differently.  This version doesn't set proctitle or call manage_update_scap_db_init/cleanup.

src/manage_sql_secinfo.c
```diff
@@ -5135,7 +5135,7 @@ rebuild_scap (const char *type)
 {
   int updated_scap_ovaldefs, updated_scap_cpes, updated_scap_cves;
 
-  updated_scap_ovaldefs = 0;
+  // updated_scap_ovaldefs = 0; // initial value currently unused
```
mattmundell, line 5138: I know this might be used in the future but I think you should remove the line for now because it's confusing.  The initial value is always set below.

timopollmeier: This has been made obsolete by the refactoring to get rid of the duplication and inconsistencies you mentioned in your other comment.

src/manage_sql_secinfo.c
timopollmeier: I've refactored some things in b813f30fbeaefffd42208150f466fdf2b06e7446 to address this.

mattmundell: Thank you.  Very neat now.

PR #1054, opened 2020-04-21
Updated the list of current Local Security Checks for the Auto-FP feature

cfi-gb:
This updates the hard-coded list of Local Security Checks used by the Auto-FP to the current used Local Security Checks families. The "Privilege escalation" family was misplaces in the list and was removed as well.

**Checklist**:

- [ ] Tests


mattmundell:
Could you give some explanation of why Privilege Escalation is no longer needed?

I'm pretty sure I was explicitly asked to include it when I added the feature, but I don't know why it was required.

cfi-gb:
"Privilege escalation" isn't a "LSC checks only" category. There might be a few LSCs placed in this family but if this is a criterion then we basically would need to add most of the existing families to this list as well.

PR #1073, opened 2020-05-08
Give feed lock a gvmd specific name.

janowagner:
This is to prevent clashes with ospd-openvas which
uses the very same name. In cases where both use the same
directory for lock files, ospd-openvas will be blocked
and not serve OSP.

**Checklist**:

- [ ] Tests N/A


janowagner:
The issue was solved by the flock patches, so closing.

PR #1075, opened 2020-05-08
Adding --extra-host-identifiers command line argument to fix register_host_detail function

befoulad:
The **manage_report_host_details** function in manage_sql.c has several if clauses that compares host identifier names to a set of hard-coded names such as "MAC", "OS" and "ip". This prevents NVT plugins which rely on register_host_detail function to set host identifier values (such as asset serial number, firmware, etc) for names other than those hard-coded ones. This PR adds a new command line parameter (--extra-host-idents) that allows the user to define a comma-separated list of identifier names to extend the hard-coded values in register_host_detail function, hence enabling NVT plugins to call register_host_detail NASL function as expected.

**Checklist**:
- [ ] Tests

### DIFF
--- a/src/gvmd.c
+++ b/src/gvmd.c
@@ -1716,6 +1716,7 @@ main (int argc, char **argv)
   static gchar *role = NULL;
   static gchar *disable = NULL;
   static gchar *value = NULL;
+  static gchar *extra_host_idents = "";
   GError *error = NULL;
   lockfile_t lockfile_checking, lockfile_serving;
   GOptionContext *option_context;
@@ -1954,6 +1955,10 @@ main (int argc, char **argv)
       &print_version,
       "Print version and exit.",
       NULL },
+    { "extra-host-idents", '\0', 0, G_OPTION_ARG_STRING,
+      &extra_host_idents,
+     "comma-separated list of extra host identifier names",
+     "<id_1,id_2>" },
     { NULL }};
 
   /* Set locale based on environment variables. */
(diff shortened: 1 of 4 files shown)


befoulad:
@janowagner @mattmundell could you please help with reviewing this PR?

mattmundell:
> @janowagner @mattmundell could you please help with reviewing this PR?

Thanks for the contribution.  I'll try find some time tomorrow.  A little worried about side effects.

Also, from a quick look this PR looks low quality: formatting is not consistent, no addition to identifier_name.

befoulad:
> Also, from a quick look this PR looks low quality: formatting is not consistent, no addition to identifier_name.
Thank you. Could you please provide me example of formatting issue so I can improve it.

bjoernricks:
And please use `--extra-host-identifiers` as the name for the argument.

timopollmeier:
> Thank you. Could you please provide me example of formatting issue so I can improve it.

The blocks of if statements and loops should be indented one more level. Also, there should be spaces between function names, condition and loop keywords and the following parentheses:

For example
```
    if(extra_host_idents)
    {
       manage_set_extra_host_idents(extra_host_idents);
    }
```
should be written like this:
```
    if (extra_host_idents)
      {
         manage_set_extra_host_idents (extra_host_idents);
      }
```

Also, the new entry in `option_entries` is missing a space in the last two lines.

befoulad:
@mattmundell , thanks for pointing out the "identifier_name" function. I added is_extra_host_identifier check to that function and tested report exporting that I believed could hit identifier_name. Please let me know any other concerns you may have regarding the side effects.

befoulad:
@bjoernricks Could you please let me know if this PR would be merged or need further work?

bjoernricks:
@timopollmeier @mattmundell could you decide about this pr?

PR #1087, opened 2020-05-14
First replacement of db functions with plain plpgsql

rrenkert:
This replaces the first C based database functions with plain plpgsql functions.

The following functions are migrated till now:
* level_max_severity
* level_min_severity
* valid_db_resource_type

- [ ] Tests


src/manage_pg.c
```diff
@@ -212,17 +212,79 @@ manage_create_sql_functions ()
...
-       " LANGUAGE C;",
-       GVM_LIB_INSTALL_DIR);
+  sql ("CREATE OR REPLACE FUNCTION level_max_severity (lvl text, cls text)"
+       "RETURNS double precision AS $$"
+       "DECLARE"
+       "  v double precision;"
+       "BEGIN"
+       "  CASE"
+       "    WHEN lvl = 'Log' THEN"
+       "      v := 0.0;" // SEVERITY_LOG
```
mattmundell, line 222: How about using the actual definition, in case they change?  Something like
```suggestion
       "      v := " G_STRINGIFY (SEVERITY_LOG) ";"
```

src/manage_pg.c
```diff
@@ -212,17 +212,79 @@ manage_create_sql_functions ()
        " LANGUAGE C;",
        GVM_LIB_INSTALL_DIR);
 
-  sql ("CREATE OR REPLACE FUNCTION level_max_severity (text, text)"
-       " RETURNS double precision"
-       " AS '%s/libgvm-pg-server', 'sql_level_max_severity'"
-       " LANGUAGE C;",
-       GVM_LIB_INSTALL_DIR);
+  sql ("CREATE OR REPLACE FUNCTION level_max_severity (lvl text, cls text)"
```
mattmundell, line 215: It would be good to make it clear that this must be kept the same as the C version of level_max_severity.

src/manage_pg.c
```diff
@@ -307,10 +307,43 @@ manage_create_sql_functions ()
        GVM_LIB_INSTALL_DIR);
 
   sql ("CREATE OR REPLACE FUNCTION valid_db_resource_type (text)"
```
mattmundell, line 309: Here also, it would be good to make it clear that it must match the C valid_db_resource_type.

Even better would be to have them defined in only one place, maybe using macros, but I guess that gets a bit complex.

rrenkert: Yes, it must be in sync with the C code. I already thought about it and my idea was to use only the database ones. Since there are not as many calls to these three functions it would be less effort than writing macros. What do you think? Do you see performance impacts?

src/manage_pg.c
rrenkert: Done that in commit 01a6cba86195ac958d504a3805b51940cadcf9e1

src/manage_pg.c
mattmundell: Looking a bit closer at valid_db_resource_type, the SQL function is only used by the SQL resource_name.  The SQL resource_name is only used by the permission iterator.  I think we should simply drop the call to valid_resource_type in SQL resource_name, and remove the function.  It's unlikely (maybe even impossible) that we get an invalid resource type in SQL resource_name.  And if we do it will just be a SQL error because the SELECT will use an invalid table.

mattmundell: About always using SQL for level_max/min_severity, these are used by severity_data_level_counts, which is used by report_counts_id_full.  Result counting is a performance hotspot, but severity_data_level_counts is only called once per report so I think it will not have a significant perfomance impact.

However, it's ugly, and does not fit in line with reducing the number of queries to the db.

I think duplicating the code is the neater option, but will not block you if you prefer using an SQL function.

mattmundell: And to be clear, I"m also happy with just duplicating the code (with a clear comment), to avoid complex macros.

rrenkert: I just commited the suggested changes. Are the comments clear enough?

By the way, I understand that you want to keep the queries to the db as low as possible!

mattmundell: Yes, that's clear, thank you.

PR #1094, opened 2020-05-18
Handle PENDING status for launched scan but still not started by ospd.

jjnicola:
Depends on PR greenbone/gvm-libs#336

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -4006,9 +4006,10 @@ prepare_osp_scan_for_resume (task_t task, const char *scan_id, char **error)
         }
     }
   else if (status == OSP_SCAN_STATUS_RUNNING
+           || status == OSP_SCAN_STATUS_PENDING
            || status == OSP_SCAN_STATUS_FINISHED)
     {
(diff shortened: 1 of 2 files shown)


src/manage.c
```diff
@@ -3630,7 +3630,8 @@ handle_osp_scan (task_t task, report_t report, const char *scan_id)
                   rc = 0;
                   break;
                 }
-              else if (osp_scan_status == OSP_SCAN_STATUS_RUNNING
+              else if ((osp_scan_status == OSP_SCAN_STATUS_RUNNING
+                        || osp_scan_status == OSP_SCAN_STATUS_RUNNING)
```
mattmundell, line 3634:
```suggestion
                        || osp_scan_status == OSP_SCAN_STATUS_PENDING)
```

jjnicola: :open_mouth:

mattmundell:
Also, there's another case of OSP_SCAN_STATUS_RUNNING in prepare_osp_scan_for_resume.  Is that relevant?

jjnicola:
PENDING status does not need to be handled. It is the same case as for INIT, which is ignored as well.

PR #1097, opened 2020-05-18
Handle removed CPEs and CVEs in SCAP sync

mattmundell:
**Checklist**:

- [ ] Tests


mattmundell:
> This looks okay overall but unless I'm missing something this doesn't remove affected products that were removed from CVEs that are still in the XML.

This is done by the changes to insert_cve_products in https://github.com/greenbone/gvmd/pull/1097/commits/2879642669a07553a942f065ec9a2e7ad80a6461.

It collects all affected products for the CVE in initial_affected, then removes the CPEs from init_affected as they are encountered.  At the end it removes any that remain.

Or do you mean something else?

mattmundell:
> I also wonder if using a temporary table or other data structure instead of buffering an array literal wouldn't be a better idea, especially when the CPEs from CVEs could have some redundancy.

I wanted to avoid the SQL queries that would be required for a temporary table.  Apparently the ANY(array...) will be converted to a temporary table by Postgres anyway.

I don't think duplicates will be a problem.  The list should be small.  (Admittedly a small list mean using SQL queries to fill a temporary table may not be a problem.)

PR #1101, opened 2020-05-20
Add 'all' case to --rebuild-scap

mattmundell:
**Checklist**:

- [ ] Tests


src/manage_sql_secinfo.c
```diff
@@ -4999,7 +4999,17 @@ rebuild_scap (const char *type)
...
+      sql ("UPDATE scap.meta SET value = '-1' WHERE name = 'last_update';");
+      ret = update_scap (FALSE,  /* ignore_last_scap_update */
+                         TRUE,   /* update_cpes */
+                         TRUE,   /* update_cves */
+                         TRUE    /* update_ovaldefs */);
+      if (ret == 1)
+        ret = 2;
+    }
+  else if (strcasecmp (type, "ovaldefs") == 0
       || strcasecmp (type, "ovaldef") == 0)
```
timopollmeier, line 5024:
```suggestion
           || strcasecmp (type, "ovaldef") == 0)
```

src/manage_sql_secinfo.c
```diff
@@ -4999,7 +4999,17 @@ rebuild_scap (const char *type)
   else if (ret)
     return -1;
 
-  if (strcasecmp (type, "ovaldefs") == 0
+  if (strcasecmp (type, "all") == 0)
+    {
+      sql ("UPDATE scap.meta SET value = '-1' WHERE name = 'last_update';");
```
timopollmeier, line 5015: I would prefer it if the `ignore_last_scap_update` could be used, but I also remember that it would require some extra changes. Do you think those changes would be too much?

mattmundell: Is https://github.com/greenbone/gvmd/pull/1101/commits/99a0b925207f87c2467243fcf702e38d85c19e2f what you have in mind?

timopollmeier: Having an explicit parameter for a full rebuild looks good but it should work without setting the `last_update` entry here. IMHO that should either no longer be required or more of a reaction to the rebuild request.

mattmundell: Oh I forgot to remove the UPDATE.  Done in e22834d.

PR #1102, opened 2020-05-21
Add TZID to DTSTART when storing icalendar

mattmundell:
**Checklist**:

- [ ] Tests


mattmundell:
> One of the biggest sources of problems in the scheduling was doing timezone adjustments and I think by adding this string manipulation you are adding another one to keep track of.

I expected as much.  Will try do it properly.  And maybe I can remove some of the schedule fields to simplify things.

PR #1113, opened 2020-06-02
Handle QUEUED task status for queued scans

jjnicola:


src/manage.c
```diff
@@ -3614,11 +3615,13 @@ handle_osp_scan (task_t task, report_t report, const char *scan_id)
               osp_scan_status = get_osp_scan_status (scan_id, host, port,
                                                      ca_pub, key_pub, key_priv);
 
-              if (osp_scan_status == OSP_SCAN_STATUS_PENDING)
+              if (osp_scan_status == OSP_SCAN_STATUS_PENDING
+                  && pending_status_updated == FALSE)
```
mattmundell: I think it would be clearer if pending_status_updated was checked in a new ````if```` inside the osp_scan_status ````if````.  Because if status is Pending then we don't need to run the ````else```` branch.

src/manage.c
```diff
@@ -3608,7 +3614,18 @@ handle_osp_scan (task_t task, report_t report, const char *scan_id)
 
               osp_scan_status = get_osp_scan_status (scan_id, host, port,
                                                      ca_pub, key_pub, key_priv);
-              if (progress >= 0 && progress < 100
+
+              if (osp_scan_status == OSP_SCAN_STATUS_QUEUED)
+                {
+                  if(queued_status_updated == FALSE)
```
mattmundell, line 3620:
```suggestion
                  if (queued_status_updated == FALSE)
```

src/manage.c
```diff
@@ -1576,6 +1576,8 @@ run_status_name (task_status_t status)
 
       case TASK_STATUS_RUNNING:          return "Running";
 
+      case TASK_STATUS_QUEUED:          return "Queued";
```
mattmundell, line 1579:
```suggestion
      case TASK_STATUS_QUEUED:           return "Queued";
```

src/manage.c
```diff
@@ -1611,6 +1613,8 @@ run_status_name_internal (task_status_t status)
...
```
mattmundell, line 1616:
```suggestion
      case TASK_STATUS_QUEUED:           return "Queued";
```

PR #1116, opened 2020-06-03
Remove classic schedule elements from GMP

mattmundell:
**Checklist**:

- [ ] Tests


mattmundell:
> I'm okay with removing most of the old elements but I wonder if we should keep the info when the schedule is next run in `get_schedules` just because that information cannot simply be extracted from the iCalendar but also requires some time calculations that can get pretty complex depending on the recurrence rules used.

Thanks, I"m going to merge because we don't help the GMP user build the ical, so I think it's OK that we don't help them decipher the ical.  GSA is already doing these calculations.  If it's a problem we can add it back later.

PR #1135, opened 2020-06-15
Use ical version of schedule GMP in wizards

mattmundell:
**Checklist**:

- [ ] Tests


src/wizards/modify_task.xml
```diff
@@ -156,6 +162,15 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             </xsl:if>
 
             <xsl:if test="(/wizard/params/param[name='reschedule']/value = 1) and (/wizard/params/param[name='start_year']/value != '') and (/wizard/params/param[name='start_month']/value != '') and (/wizard/params/param[name='start_day']/value != '') and (/wizard/params/param[name='start_hour']/value != '') and (/wizard/params/param[name='start_minute']/value != '')">
+              <xsl:variable name="start">
+                <xsl:value-of select="substring(string('100' + /wizard/params/param[name='start_year']/value), 2)"/>
```
timopollmeier, line 166: I think you could use the [`format-number()` function](https://www.w3schools.com/xml/func_formatnumber.asp) to format these numbers. Also, I'd like it better if the year was kept at 4 digits.

src/wizards/modify_task.xml
```diff
@@ -164,6 +179,21 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+                  <xsl:call-template name="newline"/>
+                  <xsl:text>DTSTART:</xsl:text>
+                  <xsl:value-of select="$start"/>
+                  <xsl:call-template name="newline"/>
+                  <xsl:text>END:VEVENT</xsl:text>
+                  <xsl:call-template name="newline"/>
+                  <xsl:text>END:VCALENDAR</xsl:text>
+                  <xsl:call-template name="newline"/>
+                </icalendar>
                 <first_time>
```
timopollmeier, line 197: It looks like the old `<first_time>` element could be removed here as well.

src/wizards/quick_task.xml
```diff
@@ -195,6 +201,15 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           </xsl:if>
 
           <xsl:if test="(/wizard/params/param[name='auto_start']/value = 1) and (/wizard/params/param[name='start_year']/value != '') and (/wizard/params/param[name='start_month']/value != '') and (/wizard/params/param[name='start_day']/value != '') and (/wizard/params/param[name='start_hour']/value != '') and (/wizard/params/param[name='start_minute']/value != '')">
+            <xsl:variable name="start">
+              <xsl:value-of select="substring(string('100' + /wizard/params/param[name='start_year']/value), 2)"/>
```
timopollmeier, line 205: Same as in modify_task.xsl with the number formatting.

src/wizards/modify_task.xml
```diff
@@ -156,6 +162,15 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
 
             <xsl:if test="(/wizard/params/param[name='reschedule']/value = 1) and (/wizard/params/param[name='start_year']/value != '') and (/wizard/params/param[name='start_month']/value != '') and (/wizard/params/param[name='start_day']/value != '') and (/wizard/params/param[name='start_hour']/value != '') and (/wizard/params/param[name='start_minute']/value != '')">
+              <xsl:variable name="start">
+                <xsl:value-of select="substring(string('100' + /wizard/params/param[name='start_year']/value), 2)"/>
+                <xsl:value-of select="substring(string('100' + /wizard/params/param[name='start_month']/value), 2)"/>
+                <xsl:value-of select="substring(string('100' + /wizard/params/param[name='start_day']/value), 2)"/>
+                <xsl:text>T</xsl:text>
+                <xsl:value-of select="substring(string('100' + /wizard/params/param[name='start_hour']/value), 2)"/>
+                <xsl:value-of select="substring(string('100' + /wizard/params/param[name='start_minute']/value), 2)"/>
+                <xsl:text>00Z</xsl:text>
```
timopollmeier, line 172: I think users would expect the time to be entered as local time according the chosen timezone instead of UTC.

src/wizards/modify_task.xml
mattmundell: Done in 83660e5.

src/wizards/modify_task.xml
mattmundell: Done in 7e96e38.

src/wizards/quick_task.xml
mattmundell: Also in 7e96e38.

src/wizards/modify_task.xml
mattmundell: The user actually provides the timezone.  Adjusted to use that timezone in 86c3adb.

PR #1134, opened 2020-06-15
Create SCAP constraints at end of CSV sync

mattmundell:
This takes my CVS SCAP sync from 60s to 20s.

**Checklist**:

- [ ] Tests
- [ ] [CHANGELOG](https://github.com/greenbone/gvmd/blob/master/CHANGELOG.md) Entry

PR #1137, opened 2020-06-16
Init SCAP directly before XML sync, in case CSV sync fails

mattmundell:
**Checklist**:

- [ ] Tests


mattmundell:
Waiting for https://github.com/greenbone/gvmd/pull/1134.

PR #1143, opened 2020-06-19
XML escape solution element in get_nvt_xml

timopollmeier:
The text in the element was only escaped properly in results, but not
the get_nvts GMP command.

**Checklist**:

- Tests N/A


janowagner:
+1, I can confirm it resolves my issue with the "General" family.

PR #1179, opened 2020-07-08
Include all source in utils-test build

mattmundell:
This solves a linker error on newer systems.  Probably due to utils.h
now including xmlutils.h, and the way gvm-libs is built.  Easier
to just include everything like the other tests, than to figure out
the details.

The build fail doesn't show up in CI, presumably because it's an older system.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -180,7 +180,24 @@ add_test (gmp-tickets-test gmp-tickets-test)
 
 add_executable (utils-test
                 EXCLUDE_FROM_ALL
-                utils_tests.c)
+                utils_tests.c
+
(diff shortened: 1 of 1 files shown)


mattmundell:
> I'm not sure about linking all the source files:
> On the one hand the utils should work with minimal dependencies to keep the Postgres extension small.
> On the other hand ensuring this could be considered outside the scope of building the tests.

Yes, it's not perfect.  Just the quickest way to get them building.

mattmundell:
> Okay, I don't really insist on keeping the linking minimal for testing. I just wanted to point it out as something to consider.

Thanks.

PR #1192, opened 2020-07-13
Quote identifiers in SQL functions using EXECUTE (20.08)

timopollmeier:
PL/pgSQL functions that build EXECUTE statements with schema, relation
or column names passed as parameters should quote them to ensure they
are always interpreted as identifiers.

**Checklist**:

- [ ] Tests


mattmundell:
Could you give some examples of what cases this is catching?

Wouldn't it be better to check the text before passing it to EXECUTE?

And try_exclusive_lock is only called once.  How about renaming it try_exclusive_lock_reports?

timopollmeier:
> Could you give some examples of what cases this is catching?
> 
This is meant to catch any case where the function parameters are not referring to schemas, tables or columns as intended, but I was thinking of ones where they are maliciously manipulated.

> Wouldn't it be better to check the text before passing it to EXECUTE?
> 
I guess that would make sense as an addition check. I was thinking of the flexibility PostgreSQL offers for the identifier names, but it may not be too complex if we just allow the cases we actually use.

> And try_exclusive_lock is only called once. How about renaming it try_exclusive_lock_reports?

I didn't want to specialize the function too much, but if no use case for other tables has shown up, I could change the function to only lock the reports table.

PR #1221, opened 2020-08-04
Ignore case when comparing disabled commands

mattmundell:
This was introduced in befde8cbbd1de47a9f2350f9f7c4d7a0cef65053 in https://github.com/greenbone/gvmd/pull/807.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_acl.c
+++ b/src/manage_acl.c
@@ -73,6 +73,29 @@ user_may_internal (const char *operation)
 }
 
 /**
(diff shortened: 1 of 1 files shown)


mattmundell:
> This looks okay overall but I think the function name `strv_strcasecmp` looks like it may return something non-zero if all the strings in the array are different from the search string.

Good point.  Done in b82a876.

PR #1224, opened 2020-08-05
Add missing min_qod to FILTERS in GET_VULNS response

mattmundell:
**Checklist**:

- [ ] Tests


mattmundell:
> I think the vulns should get a separate case where only min_qod is added to the filter because apply_overrides has no effect.

Done in https://github.com/greenbone/gvmd/pull/1224/commits/6a1ad1170bce08467fb5272850e4ab05e94333d1.

PR #1229, opened 2020-08-07
Updated outdated links and project name in docs.

cfi-gb:
This was missed during the transition from OpenVAS to GVM and from wald.intevation.de to github.

The final manpage can be verified with something like e.g.:

```
man doc/greenbone-certdata-sync.8
```


cfi-gb:
Backports to the other active branches in #1230 and #1231

PR #1232, opened 2020-08-07
[20.08] Additional tweaks to the docs / manpages:

cfi-gb:
- Update references to newer / renamed / deprecated manpages
- Updated links and texts in URLs
- Use a consistent header for the MORE INFORMATION part.

To have a consistent manpage / doc across all main modules: https://github.com/greenbone/gsa/pull/2369

Additional notes:

1. gvm related man pages / docs don't need to link to openvas.org so i have replaced this link with greenbone.net similar to already done in GSA
2. There is no gvm-cli manpage, have replaced it with the other links which makes more sense


cfi-gb:
Backport to the gvmd-9.0 branch in #1233

Haven't backported it to the gvmd-8.0 branch due to the differences in ospd-openvas not existing there, please let me know if i should do a manual backport of these changes.

PR #1241, opened 2020-08-10
Add prerequisites for S/MIME suport to INSTALL.md

cfi-gb:
As #741 doesn't move forward we should have at least the text based prerequisite mentioned in the INSTALL.md

According to @timopollmeier gpgsm is required for S/MIME support. This might be related to the issue seen in e.g. #540.

**Checklist**:

- [ ] Tests

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -19,6 +19,9 @@ Prerequisites:
 Prerequisites for certificate generation:
 * GnuTLS certtool
 
+Prerequisites for S/MIME suport (e.g. email encryption):
+* GNU privacy guard - S/MIME version
+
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -639,6 +639,8 @@ Prerequisites for Tipping Point alert:
 Prerequisites for key generation on systems with low entropy:
 * haveged (or a similar tool)
 
+Prerequisites for S/MIME suport (e.g. email encryption):
```
cfi-gb, line 642: suport -> support

cfi-gb: Fix in #1257

PR #1252, opened 2020-08-14
Do not set beta version implicitly

ArnoStiefvater:
Use PROJECT_BETA_RELEASE variable instead of
setting the beta release implicitly when no patch
version is provided.

**Checklist**:

- [ ] Tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,8 +19,6 @@ cmake_minimum_required (VERSION 3.0)
 
 message ("-- Configuring Greenbone Vulnerability Manager...")
 
-# VERSION: Set patch version for stable releases, e.g. "9.0.0",
-#          unset patch version for prereleases, e.g. "9.0"
 project (gvm
          VERSION 21.4
          LANGUAGES C)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -61,12 +59,21 @@ else (${PROJECT_VERSION_MINOR_LENGTH} GREATER 1)
   set (PROJECT_VERSION_MINOR_STRING "0${PROJECT_VERSION_MINOR}")
 endif (${PROJECT_VERSION_MINOR_LENGTH} GREATER 1)
 
-# If a patch version is not defined, assume a pre-release
-if (DEFINED PROJECT_VERSION_PATCH AND NOT PROJECT_VERSION_PATCH STREQUAL "")
+# Set beta version if this is a beta release series,
+# unset (put value 0) if this is a stable release series.
+if (NOT PROJECT_BETA_RELEASE)
+  set (PROJECT_BETA_RELEASE 1)
+endif (NOT PROJECT_BETA_RELEASE)
```
bjoernricks, line 66:
```suggestion
set (PROJECT_BETA_RELEASE 1)
```

bjoernricks: Not need for the if clause here. We either set it or not.

PR #1256, opened 2020-08-14
Remove zero padding for version strings

ArnoStiefvater:
**Checklist**:

- [ ] Tests


mattmundell:
> There are hardcoded versions in the CHANGE elements in src/schema_formats/XML/GMP.xml.in.
> 
> These should be changed too.

Actually the 20.08 do not change, so it's correct, merging.

PR #1260, opened 2020-08-17
 Move EXE credential generation to a Python script

timopollmeier:
Creating the EXE credential in an external script using a template makes
the NSIS script generation easier to read, maintain and customize.

**Checklist**:

- [ ] Tests


tools/gvm-lsc-exe-creator
```diff
@@ -0,0 +1,66 @@
...
+                  "__PASSWORD__" : password,
+                  "__OUTPUT_PATH__" : args.output_path,
+                }
+nsis_script = template.safe_substitute(substitutions)
+nsis_script_path = os.path.join(args.temp_dir, "script.nsis")
+with open(nsis_script_path, "w") as nsis_script_file:
+    nsis_script_file.write(nsis_script)
+
+run = subprocess.run(["makensis", nsis_script_path])
+sys,exit(run.returncode)
```
bjoernricks, line 66: There is typo. Needs to be sys.exit

bjoernricks: And the code should be put into a function that gets called instead

timopollmeier: These should be fixed in #1262

PR #1264, opened 2020-08-20
Fix SCP alert authentication and logging (20.08)

timopollmeier:
The BatchMode=yes option for scp is removed to make sshpass work and the
prompt expected by sshpass is adjusted for public key auth.

Carriage return characters are removed from the log output as they
can make journalctl interpret it as blob data.

(This was originally fixed in gvmd-9.0 with #972 but not ported to the 20.08 branch)

**Checklist**:

- [ ] Tests


bjoernricks (approved):
Not that I understand the changes completely but we need this fix asap.

PR #1266, opened 2020-08-20
Check feed status without acquiring lock

timopollmeier:
The feed versions / timestamps are now checked without acquiring the lock on the lockfile first so the check does not appear as an update in progress.
This check and handling the actual updates is also done in a forked process now so the scanner taking longer to respond with the feed version does not block the main process.

**Checklist**:

- [ ] Tests


mattmundell (approved):
Pity that things get more complex, but seems sensible.

PR #1275, opened 2020-08-26
Send NULL as extra_where_single in init_result_get_iterator

mattmundell:
**What**:

Prevents filters from having an effect when specifying a result_id in
GET_RESULTS.

**Why**:

This error was introduced in c9dc67a01ed0f9b381172634d4514ccd2bd83ef2 (from https://github.com/greenbone/gvmd/pull/1076) which
introduced the extra_where_single arg.  Before that there was no way to
send extra args to init_get_iterator2_with when getting a single specified
resource, which is equivalent to sending NULL now.

**How**:

Response from each of these should include result, even if result has min_qod < 70:
 - `2o m m '<get_results result_id="e89920a5-9af3-47af-975c-cb38cccdea69" filter="min_qod=70"/>'`
 - `2o m m '<get_results result_id="e89920a5-9af3-47af-975c-cb38cccdea69" filter="min_qod=0"/>'`
 - `2o m m '<get_results result_id="e89920a5-9af3-47af-975c-cb38cccdea69"/>'`

**Checklist**:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix delta sorting for unusual filter sort terms [#1249](https://github.com/greenbone/gvmd/pull/1249)
 - Fix SCP alert authentication and logging [#1264](https://github.com/greenbone/gvmd/pull/1264)
 - Set file mode creation mask for feed lock handling [#1265](https://github.com/greenbone/gvmd/pull/1265)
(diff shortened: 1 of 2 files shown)


timopollmeier (changes requested):
While this fixes the issue with the min_qod, I think this will also make the command ignore the `apply_overrides` keyword.
I think we should selectively filter out the `min_qod` or set it to 0 for the `extra_where_single` clause instead.

timopollmeier:
I've implemented a different solution that still applies other special keywords like `apply_overrides` in #1276.

PR #1276, opened 2020-08-27
Ignore min_qod when getting single results by UUID

timopollmeier:
**What**:

Prevents results being hidden by implicit or explicit min_qod filters when selecting results by UUID.

This is an alternative solution to the issue in #1275 that also allows using other filter options like `apply_overrides=1`.

**Why**:

This error making get_results return no result at all was introduced in c9dc67a01ed0f9b381172634d4514ccd2bd83ef2.
Before this special filter options like min_qod were simply ignored by
init_get_iterator2_with when getting single results, ensuring the result
would be visible.

**How**:

Response from GMP commands like these should include the result, even if result has min_qod < 70:
 - `<get_results result_id="e89920a5-9af3-47af-975c-cb38cccdea69" filter="min_qod=70"/>`
 - `<get_results result_id="e89920a5-9af3-47af-975c-cb38cccdea69" filter="min_qod=0"/>`
 - `<get_results result_id="e89920a5-9af3-47af-975c-cb38cccdea69"/>`

**Checklist**:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix delta sorting for unusual filter sort terms [#1249](https://github.com/greenbone/gvmd/pull/1249)
 - Fix SCP alert authentication and logging [#1264](https://github.com/greenbone/gvmd/pull/1264)
 - Set file mode creation mask for feed lock handling [#1265](https://github.com/greenbone/gvmd/pull/1265)
+- Ignore min_qod when getting single results by UUID [#1276](http://github.com/greenbone/gvmd/pull/1276)
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


bjoernricks (approved):
Looks good to me. If requesting a single results the filter will always use `min_qod=0` independent of the filter from the user.

mattmundell:
Doesn't this change the behaviour of GET_RESULTS from gvmd <20.08 though?

In gvmd <20.08 there was no extra_where_single arg for init_get_iterator2_with, so none of the filtering would reach the SQL when an id was given.

This was consistent with the other types, that ignore the filter when an id is given, for example

`<get_targets target_id="47ab4d8a-eba3-11ea-9fb4-13d9aeb88196" filter="name=abc"/>`

does not return an empty response when the target's name is 'def'.

PR #1278, opened 2020-08-31
Fix alternative options for radio type preferences when exporting a scan config

jjnicola:
**What**:
Add alternative option for radio type nvt's preferences

**Why**:
When exporting, the alternative options were not added to the xml file.

**How**:
Importing a scan config and checking the NVT's preferences, search for an radio type pref. Only the default value was shown. Now the alternative values are shown as well.

**Checklist**:

- [ ] Tests N/A

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -8674,13 +8674,6 @@ buffer_config_preference_xml (GString *buffer, iterator_t *prefs,
       char *pos = strchr (value, ';');
       if (pos) *pos = '\0';
       buffer_xml_append_printf (buffer, "<value>%s</value>", value);
-      while (pos)
-        {
-          char *pos2 = strchr (++pos, ';');
-          if (pos2) *pos2 = '\0';
-          buffer_xml_append_printf (buffer, "<alt>%s</alt>", pos);
-          pos = pos2;
-        }
     }
   else if (value
            && type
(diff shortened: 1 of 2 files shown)


src/gmp.c
```diff
@@ -8674,13 +8674,6 @@ buffer_config_preference_xml (GString *buffer, iterator_t *prefs,
       char *pos = strchr (value, ';');
       if (pos) *pos = '\0';
       buffer_xml_append_printf (buffer, "<value>%s</value>", value);
-      while (pos)
```
mattmundell, line 8677: Surely this hasn't been broken since 2010?  I think this worked because OTP sent the alts with the value.  Has ospd-openvas stopped doing that?

mattmundell (changes requested):

src/gmp.c
mattmundell: Hmm, I think somewhere along the line we lost track of keeping the alts in the preference value in the XML.  For example in the feed config XML one radio value includes the alts but the other one does not.

And gvmd was not checking for the alts when importing XML configs, so the values ended up in dbs without alts.  My current db has both.

I think using using default_value which comes from the nvts table is a good solution.

mattmundell (approved):

PR #1281, opened 2020-08-31
Remove DROP from vulns creation

mattmundell:
This DROP was added in 01baac22a, which changed the view to a UNION with
explicit columns.  It's likely that the DROP was needed because CREATE OR
REPLACE does not support changing the columns.

But that was 2017 and the columns have not changed since, so anyone
upgrading from the last version will have the same columns.

If someone needs to change the columns they can add a migrator.

**What**:

Remove DROP from vulns creation.

**Why**:

We have reports of "ERROR: relation "vulns" does not exist" that are hard to reproduce, and this may help.

**How**:

Restart gvmd.

**Checklist**:

- [ ] Tests


bjoernricks (approved):
Looks legit to me

PR #1285, opened 2020-09-01
Reduce severity classes

janowagner:
The list of supported severity classes is reduced by removing last remains
of "classic" (deprecated since v8.0) and by removing "bsi" which is actually the same as "nist".

This cleans code and UI, without removing functionality.

- [ ] Tests N/A


mattmundell (approved):
This could be confusing to users who have BSI as their "Severity Class" setting.  But I guess you're going to remove that in subsequent PRs.

janowagner:
> This could be confusing to users who have BSI as their "Severity Class" setting. But I guess you're going to remove that in subsequent PRs.

The only confusion can happen in "My Settings", which is addressed by https://github.com/greenbone/gsa/pull/2434 .
The actual classification is identical.

mattmundell:
> The only confusion can happen in "My Settings", which is addressed by greenbone/gsa#2434 .

Nice.  So they will still appear in GMP GET_SETTINGS.  How about removing the BSI settings from the db?

janowagner:
> Nice. So they will still appear in GMP GET_SETTINGS. How about removing the BSI settings from the db?

on my agenda, but related to another aspect.

PR #1288, opened 2020-09-05
Reduce severity classes

janowagner:
The support for multiple severity class ranges is removed.
There is now only one classification scheme (the one previously
called "nist"). Actually there was only one other classification scheme called
"pci-dss". It wasn't used or needed.

With applying a single default scheme code and database operations can
be simplified.

Basically all code elements are removed, including a migrator
that removes any such setting from the database.

The "report" XML element does not include the severity classification with
its range details (as element ""severity_class) anymore and the "authenticate_response"
XML does not include the "severity" element anymore.

**Checklist**:

- [ ] Tests N/A


src/manage_pg.c
```diff
@@ -1443,12 +1443,10 @@ manage_create_sql_functions ()
            "                                              text,"
            "                                              text)"
            " RETURNS boolean AS $$"
-           "              (CASE lower ($2)"
+           "              (SELECT CASE lower ($2)"
            "               WHEN 'high'"
            "               THEN $1 >= 7"
            "                    AND $1 <= 10"
-           "               WHEN 'medium'"
```
mattmundell, line 1476: Missing medium now?

src/manage_pg.c
```diff
@@ -1443,12 +1443,10 @@ manage_create_sql_functions ()
            "                                              text,"
            "                                              text)"
            " RETURNS boolean AS $$"
-           "              (CASE lower ($2)"
+           "              (SELECT CASE lower ($2)"
```
mattmundell, line 1445:
```suggestion
           "  SELECT CASE lower ($2)"
```

src/manage_pg.c
```diff
@@ -1443,12 +1443,10 @@ manage_create_sql_functions ()
...
            "               WHEN 'high'"
```
mattmundell, line 1446:
```suggestion
           "         WHEN 'high'"
```
etc

src/manage_pg.c
```diff
@@ -1462,10 +1462,20 @@ manage_create_sql_functions ()
       sql ("CREATE OR REPLACE FUNCTION severity_in_level (double precision,"
            "                                              text)"
            " RETURNS boolean AS $$"
-           " SELECT severity_in_level"
-           "         ($1,"
-           "          $2,"
-           "          'dummy')"
+           "              (SELECT CASE lower ($2)"
```
mattmundell, line 1445: Indenting as before.

mattmundell (changes requested):

mattmundell:
Also, it's missing the changes to the GMP doc.

src/manage_pg.c
janowagner: Fixed, this was a mistake.

mattmundell (approved):

PR #1291, opened 2020-09-07
Skip DB check in helpers when main process is running

mattmundell:
**What**:

Skip DB check in helpers (--create-user, etc) when main process is running.

**Why**:

The create_tables in the db check was causing deadlock.

**How**:

1 `tail -f gvmd.log | grep -i "deadlock detected"`
2 `while true; do gvmd --modify-setting 78eceaec-3385-11ea-b237-28d24461215b --value d1ab52ec-a88a-4ef3-99d5-57123f88a95c; done`
3 In GSA refresh the results page.
4 Without patch should see tail output after 1 or 2 refreshes.

**Checklist**:

- [ ] Tests


bjoernricks (dismissed):
Looks legit

PR #1295, opened 2020-09-11
Correct FROM of some queries to the right schema

chris-x86-64:
**What**:

Relation `cert_bund_advs` does not exist, however, the table resides in a schema called `cert`, hence it should be  `cert.cert_bund_advs`. Same goes with `cert_bund_cves` as well.
With this PR, `FROM cert_` will be replaced with `FROM cert.cert_` .

**Why**:

When I tried to `EXPLAIN ANALYZE` a query, I got the following error:
```sql
ERROR:  relation "cert_bund_advs" does not exist
LINE 1: ..., nvts.tag, (ARRAY (SELECT name::text        FROM cert_bund_...
```
I looked into the source code and found that the table belongs to schema `cert`.
Proof: https://github.com/greenbone/gvmd/blob/ea8242eab4dd17c73cc8bc2cc2fa27bccd34279d/src/manage_pg.c#L3070-L3083

With my changes, these queries should now point to the right table in the right schema.
I also suspect that this is could be the cause of a performance issue reported in a third-party Docker deployment. https://github.com/Secure-Compliance-Solutions-LLC/GVM-Docker/issues/54

**How**:

I did a basic `sed` substitution.
```
git grep -l 'FROM cert_' | xargs sed -i 's/FROM cert_/FROM cert.cert_/g'
```

Not verified yet as I haven't taken time to prepare a build environment (which what makes this PR a draft).
I'm planning on adding tests later.

**Checklist**:

- [ ] Tests
- [x] [CHANGELOG](https://github.com/greenbone/gvmd/blob/master/CHANGELOG.md) Entry 4be00c0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 - Use GMP version with leading zero for feed dirs [#1287](https://github.com/greenbone/gvmd/pull/1287)
+- Correct FROM of some queries to the right schema [#1295](https://github.com/greenbone/gvmd/pull/1295)
 
 ### Removed
 - Reduce Severity Classes [#1285](https://github.com/greenbone/gvmd/pull/1285)
--- a/src/manage_pg.c
+++ b/src/manage_pg.c
@@ -3133,7 +3133,7 @@ manage_db_init (const gchar *name)
       sql ("CREATE OR REPLACE FUNCTION cert.cert_delete_bund_adv ()"
            " RETURNS TRIGGER AS $$"
            " BEGIN"
-           "   DELETE FROM cert_bund_cves where adv_id = old.id;"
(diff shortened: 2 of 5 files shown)


mattmundell:
You could also set the search_path like gvmd does, then your EXPLAIN ANALYZE would work.

chris-x86-64:
Thanks, it did work and I got the results from EXPLAIN ANALYZE.

```sql
gvmd=> set search_path to "$user",public,cert;
SET
gvmd=> explain analyze ...
```

I overlooked the `search_path` configurations in src/manage_pg.c, sorry. https://github.com/greenbone/gvmd/blob/ea8242eab4dd17c73cc8bc2cc2fa27bccd34279d/src/manage_pg.c#L3035-L3039

mattmundell:
I'm closing because this is getting old, it's possible to solve with search_path, and some cases have already been done in #1343.

By the way, I originally did not do this to allow switching databases, similar to the --database option for the main database.

PR #1300, opened 2020-09-16
Drop autofp

janowagner:
**What**:

Remove the functionality "autofp" (Auto False Positives).
This is a filter functionality for scan results using CVEs of confirmed
applied vendor fixes to eliminate same CVE alerts from less reliable tests (like
remote banner tests).
Associated with this the "auto_type" is removed, which is the new
severity type "False Positive" in case autofp applies to the result.

**Why**:

The QoD is capable to filter our less reliable results where wanted.
This method shows a much better performance as well, and actually
is applied by default already with a vlaue of 70%.

**How**:

The handling of the filter element "autofp" is removed as well as
the corresponding SQL query extensions and the returned result
extensions.
**Checklist**:

- [ ] Tests N/A


src/manage_sql.c
```diff
@@ -23583,7 +23583,7 @@ report_severity_data (report_t report, const char *host,
   iterator_t results;
 
   gchar *filter;
-  int apply_overrides, autofp;
+  int apply_overrides ;
```
```suggestion
  int apply_overrides;
```

mattmundell (changes requested):

src/manage_sql.c
```diff
@@ -21250,23 +21247,20 @@ where_levels_auto (const char *levels, const char *new_severity_sql,
...
-                                                   (SEVERITY_FP) ")"
-                                  "      OR %s = 1)",
-                                  auto_type_sql,
-                                  new_severity_sql,
-                                  auto_type_sql);
+                                  " AND (("
+                                  "       %s"
+                                  "       = " G_STRINGIFY
+                                               (SEVERITY_FP) ")"
+                                  " )",
```
```suggestion
                                               (SEVERITY_FP) "))",
```

src/manage_sql.c
```diff
@@ -21250,23 +21247,20 @@ where_levels_auto (const char *levels, const char *new_severity_sql,
...
         }
       else
         g_string_append_printf (levels_sql,
                                 " OR %s"
                                 "    = " G_STRINGIFY
                                           (SEVERITY_FP) "))"
-                                " OR %s = 1)",
-                                new_severity_sql,
-                                auto_type_sql);
+                                " )",
```
mattmundell, line 21262: Also would be nicer as ")))", on line 21261.

src/manage_sql.c
```diff
@@ -21800,19 +21800,13 @@ init_result_get_iterator_severity (iterator_t* iterator, const get_data_t *get,
...
   gchar *filter;
-  int autofp, apply_overrides, dynamic_severity;
+  int apply_overrides, dynamic_severity;
   gchar *extra_tables, *extra_where, *owned_clause, *with_clause;
   gchar *with_clauses;
   char *user_id;
 
   assert (report);
 
-  columns[0].select
```
mattmundell: This adjusts the columns, but I don't see any adjustment in the columns expected by the caller (report_severity_data).

src/manage_sql.c
janowagner: Fixed.

mattmundell (changes requested):

src/manage_sql.c
mattmundell: Background: init_result_get_iterator_severity is a special iterator used only for counting the results.  It avoids all the columns used by the general result iterators, so does not benefit from the changes to RESULT_ITERATOR_COLUMNS* and result_iterator_severity etc.

mattmundell (approved):

PR #1301, opened 2020-09-16
Add permission check on host in OS host count

mattmundell:
**What**:

Add a permission check on the hosts in the host count of the OS asset iterator.

**Why**:

The count was including hosts that the user did not have access to.

**How**:

Scan a new host with a new OS with user 1.
Scan same host with user 2.
Check GSA > Assets > OSs: count should be 1 not 2.

**Checklist**:

- [ ] Tests


timopollmeier (changes requested):
This seems to work for the most part but it seems to break the "Operating Systems by Vulnerability Score" chart for me.
I think it's probably because the replacement of the extra WITH clause by extra_tables isn't applied (correctly) to the `get_aggregates` command.

mattmundell:
> This seems to work for the most part but it seems to break the "Operating Systems by Vulnerability Score" chart for me.
> I think it's probably because the replacement of the extra WITH clause by extra_tables isn't applied (correctly) to the `get_aggregates` command.

Should work now.

timopollmeier (approved):

PR #1306, opened 2020-09-29
Fix create_credential for snmpv3.

jjnicola:
**What**:
Consider the case in which a snmp credential is created for version 3 only,
where the community is not requiered.

**Why**:
The credential was not created successfully for snmp v3 if a community was not provided. 

**How**:
Create credentials for a snmp v3 target without community and check that the login is successful.

**Checklist**:

- [ ] Tests N/A

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -33948,6 +33948,7 @@ create_credential (const char* name, const char* comment, const char* login,
   gchar *generated_private_key;
   credential_t new_credential;
   int auto_generate, allow_insecure_int;
+  int using_snmp_v3;
   int ret;
 
   assert (name && strlen (name) > 0);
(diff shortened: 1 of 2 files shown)


src/manage_sql.c
```diff
@@ -34209,7 +34211,7 @@ create_credential (const char* name, const char* comment, const char* login,
     }
 
   /* SNMP passwords */
-  if (community)
+  if (community || using_snmp_v3)
```
timopollmeier, line 34214: The `lsc_crypt_encrypt` function call setting the `community` within the `if (community || using_snmp_v3)` block should handle the community parameter either as as a separate case that only sets password and privacy_password or by using an empty string for `community` is if it is `NULL`.

timopollmeier (changes requested):

src/manage_sql.c
timopollmeier: Sorry, I've mixed this up with the entry **name** being NULL as a terminator, so this should be okay.

timopollmeier (approved):

PR #1316, opened 2020-10-06
Drop severity type "debug"

janowagner:
**What**:

Any code to handle the message type "Debug Message" and the severity type "Debug" is removed.

**Why**:

This removes unneeded code and thus simplifies code. Also, this potentially makes code faster
as it removes useless conditionals from result loops.

The creation of Debug Messages was removed with the release of OpenVAS-5 in March 2012.

**How**:

Testing the changes was about testing the scan results keep the same.
Basically it is dead code with no influence other than consuming CPU cycles.
 

**Checklist**:

- [ ] Tests N/A


src/manage_migrators.c
```diff
@@ -2391,6 +2391,40 @@ migrate_234_to_235 ()
...
+    {
+      sql_rollback ();
+      return -1;
+    }
+
+  /* Update the database. */
+
+  /* Replace any result type "Debug Message" by "Error Message". */
+
+  sql ("UPDATE results SET type = 'Error Message' WHERE type = 'Debug Message';");
```
mattmundell, line 2416: Technically I guess this should set severity to -3.0.  But realistically I don't think there are any debug messages out there.

mattmundell (approved):

PR #1320, opened 2020-10-06
Move GET iterator ordering selection into FROM clause

mattmundell:
**What**:

Performance experiment for GET_RESULTS.

Moves the CASE and "SELECT FROM result_new_severities" from the ORDER BY clause into the FROM clause.

Needs https://github.com/greenbone/gvmd/pull/1319.

**Why**:

**How**:

**Checklist**:

- [ ] Tests


mattmundell:
Doesn't have the speedup I expected.

PR #1327, opened 2020-10-14
add user parameter for db connection

nichtsfrei:
**What**:

under some circumstances the db username is not the same as the local
user. To support that usecase the parameter db-user is added.

**Why**:

To enable socket connection into an container with `trust` instead of `peer` configuration.

**How**:

Start a docker container with with a installed postgresql, mount `-v /tmp/test:/var/run/postgresql:z` , change peer to trust in it via:
```
 RUN sed -i 's/peer/trust/' /etc/postgresql/11/main/pg_hba.conf
```
start a migration with a postgres known username and db:
```
/usr/local/sbin/gvmd --db-host=/tmp/test/ -d gvmd --db-user gvm -m
```

**Checklist**:

- [ ] Tests


mattmundell (changes requested):
Looks OK.

Could you add a change log entry, please?

bjoernricks (approved):

PR #1334, opened 2020-10-22
Delete TLS certificate sources when deleting users

timopollmeier:
**What**:
The delete_tls_certificates_user function now also removes the referenced
rows from tls_certificate_sources, tls_certificate_locations and
tls_certificate_origins.

**Why**:
This caused foreign key constraint violations when deleting a user owning
any TLS certificates.

**How**:
Tested this by creating a new user, importing a report with TLS certificates and then trying to delete the user without an inheritor.
With the patch this should work, whereas it caused an error like this before:
```
sql_exec_internal: PQexec failed: ERROR:  update or delete on table "tls_certificates" violates foreign key constraint "tls_certificate_sources_tls_certificate_fkey" on table "tls_certificate_sources"
DETAIL:  Key (id)=(806) is still referenced from table "tls_certificate_sources".
```

**Checklist**:

- [ ] Tests


bjoernricks (approved):
Just a question besides the PR is looking good to me, why aren't we using `ON DELETE CASCADE` more often especially for something like this?

timopollmeier:
In older versions using SQLite we did not use any constraints, so the default style was to remove referenced rows explicitly.
Doing this should also serve as a reminder to remove referenced rows that do not use a foreign key constraint.
Perhaps this could be changed if we use foreign keys more consistently.

PR #1341, opened 2020-10-29
Handle some cases of missing CVE feed info

mattmundell:
**What**:

Handle missing scores in the feed XML and the alternate name for cvss3:vector.

**Why**:

SCAP sync was failing.

**How**:

`gvmd --rebuild-scap=all` should work.

Note that sync will also fail because some of the v2 CVSS base entries are missing vectors in the feed.  This needs to be fixed on the feed side.  Let me know if you need hack code to do it by hand.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_sql_secinfo.c
+++ b/src/manage_sql_secinfo.c
@@ -47,6 +47,7 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+#include <gvm/base/cvss.h>
 #include <gvm/base/proctitle.h>
 #include <gvm/util/fileutils.h>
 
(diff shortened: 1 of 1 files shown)


src/manage_sql_secinfo.c
```diff
@@ -2786,35 +2787,35 @@ insert_cve_from_entry (element_t entry, element_t last_modified,
...
-          g_free (id);
-          return -1;
-        }
+      double base;
 
       cvss_vector = element_child (base_metrics,
                                    cvss_is_v3 ? "cvss3:vector-string"
                                               : "cvss:vector-string");
+      if (cvss_is_v3 && (cvss_vector == NULL))
+        /* At least nvdcve-2.0-2004.xml uses this format. */
```
timopollmeier, line 2801: This is actually about an upcoming change in the feed that will affect all CVE XML files, not just the 2004 one.

src/manage_sql_secinfo.c
```diff
@@ -2786,35 +2787,35 @@ insert_cve_from_entry (element_t entry, element_t last_modified,
...
           g_free (id);
           return -1;
         }
-    }
 
-  if (score == NULL)
-    score_int = 0;
-  else
-    score_int = round (atof (element_text (score)) * 10);
+      base = get_cvss_score_from_base_metrics (element_text (cvss_vector));
```
timopollmeier, line 2810: I'm not sure if we should always calculate the score from the metrics or just use that as a fallback.

timopollmeier (changes requested):
Using the old element name as a fallback looks like a good idea but there are some small thing I'd change.

mattmundell:
Closing because this will be covered by changes in the feed.  Better to catch errors on the feed side I think.

PR #1346, opened 2020-11-04
Add `allow_simult_ips_same_host` field for targets

timopollmeier:
**What**:
This adds the `allow_simult_ips_same_host` column to the targets and corresponding elements to the GMP commands `create_target`, `get_targets` and `modify_target`.
The value is sent to the scanner as a scanner preference.

**Why**:
This allows defining the scanner preference added in https://github.com/greenbone/openvas/pull/604 as part of the target in case scanning multiple IP addresses of a single host causes problem

**How**:
I've checked the GMP responses of the `get_targets` command by inspecting the HTTP responses for the targets list and trashcan in GSA.

For the other two GMP commands I've used gvm-cli with GMP commands like the following with the values "0", "1" and "123":
```xml
<modify_target target_id="88c4b9e5-5b1c-4a0b-8b7c-b2b098676526">
  <allow_simult_ips_same_host>0</allow_simult_ips_same_host>
</modify_target>
```
```xml
<create_target>
  <allow_simult_ips_same_host>0</allow_simult_ips_same_host>
  <name>localhost test</name>
  <hosts>localhost</hosts>
  <port_list id="c7e03b6c-3bbe-11e1-a057-406186ea4fc5"/>
</create_target>
```

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -30823,7 +30823,8 @@ copy_target (const char* name, const char* comment, const char *target_id,
 
   ret = copy_resource ("target", name, comment, target_id,
                        "hosts, exclude_hosts, port_list, reverse_lookup_only,"
-                       " reverse_lookup_unify, alive_test",
+                       " reverse_lookup_unify, alive_test,"
+                       " allow_simult_ips_same_host",
                        1, new_target, &old_target);
   if (ret)
     return ret;
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -30995,6 +30995,7 @@ delete_target (const char *target_id, int ultimate)
  * @param[in]   reverse_lookup_only   Scanner preference reverse_lookup_only.
  * @param[in]   reverse_lookup_unify  Scanner preference reverse_lookup_unify.
  * @param[in]   alive_tests     Alive tests.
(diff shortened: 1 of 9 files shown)


src/manage_sql.c
```diff
@@ -32117,6 +32117,20 @@ target_reverse_lookup_unify (target_t target)
...
 
+/**
+ * @brief Return the allow_simult_ips_same_host value of a target.
+ *
+ * @param[in]  target  Target.
+ *
+ * @return The allow_simult_ips_same_host value if available, else NULL.
+ */
+char*
+target_allow_simult_ips_same_host (target_t target)
```
mattmundell, line 32128: Not a big deal, but wouldn't returning an int be cleaner?  It's an integer column in the db, and being compared to 0 by the caller.

mattmundell (approved):

src/manage_sql.c
timopollmeier: I've considered using int but then I've done it this way mainly for consistency with the similar target options reverse_lookup_only and reverse_lookup_unify. Perhaps we should change all of them in a separate PR.

timopollmeier:
Note: The field has been renamed to `allow_simultaneous_ips` in #1396

PR #1348, opened 2020-11-05
Replace sql_severity_matches_ov with a plain plpgsql function.

s-l-teichmann:
**What**:
The function severity_matches_ov() used in some SQL statements is currently
implemented in C. There is no need to do so as it converts pretty simple
to plpgsql.

**Why**:

Reducing the level of dependency to custom C functions in the database.

**How**:

The original C function is rewritten to only use plpgsql.

**Checklist**:

N/A

### DIFF
--- a/src/manage_pg.c
+++ b/src/manage_pg.c
@@ -288,13 +288,18 @@ manage_create_sql_functions ()
        " LANGUAGE C;",
        GVM_LIB_INSTALL_DIR);
 
-  sql ("CREATE OR REPLACE FUNCTION severity_matches_ov (double precision,"
-       "                                                double precision)"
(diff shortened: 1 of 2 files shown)


mattmundell (approved):

mattmundell:
By the way "How" is for "How did you test it".  It's a bad heading.

bjoernricks:
> By the way "How" is for "How did you test it". It's a bad heading.

Yes indeed. We should change the headline.

bjoernricks:
Does this change need an additional migration?

mattmundell:
> Does this change need an additional migration?

No I don't think so.  The new version will just replace the old one.

PR #1349, opened 2020-11-06
GVM_CERTIFICATE_SAN missing $

andreasbrett:
**What**:

**Why**:
- env var GVM_CERTIFICATE_SAN was missing leading dollar sign
- SAN settings were never put into the template file therefore creating CSRs without SAN information

**How**:
Verified/tested through creating CSR, getting it signed by CA and using it in GVM.

**Checklist**:
- [x] Tests
- [] [CHANGELOG](https://github.com/greenbone/gvmd/blob/master/CHANGELOG.md) Entry N/A

### DIFF
--- a/tools/gvm-manage-certs.in
+++ b/tools/gvm-manage-certs.in
@@ -360,7 +360,7 @@ create_certificate ()
     fi
     if [ -n "$GVM_CA_CERTIFICATE_SAN" ]
     then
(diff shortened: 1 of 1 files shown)


cfi-gb:
Ref to the PR where this functionality was introduced: #1246

timopollmeier (approved):

PR #1337, opened 2020-10-27
Switch to dynamic-only result severities (query time)

mattmundell:
**What**:

Test of removing static severity.

This is "query-time" result severity.  The severity is calculated from the NVT when the result is queried.  Compare "sync-time" severity at https://github.com/greenbone/gvmd/pull/1356.

**Why**:

**How**:

**Checklist**:

- [ ] Tests
- [ ] [CHANGELOG](https://github.com/greenbone/gvmd/blob/master/CHANGELOG.md) Entry

PR #1356, opened 2020-11-17
Switch to dynamic-only result severities (sync time)

mattmundell:
**What**:

Test of keeping the severities in the results table, but updating them when the NVTs change.

This is "sync-time" dynamic severity.  The severity of the results are adjusted when the NVTs are synced.  Compare "query-time" severity at https://github.com/greenbone/gvmd/pull/1337.

**Why**:

**How did you test it**:

**Checklist**:

- [ ] Tests


mattmundell:
This is the better option.  It's faster than https://github.com/greenbone/gvmd/pull/1337 and the same speed as static severity.

Closing because master has diverged.  Better for someone else to do this from scratch.

PR #1359, opened 2020-11-20
Get origin and date for VT severities from OSP

timopollmeier:
**What**:

When getting the VTs via OSP, the origin and date elements are now also
read and stored in the database.

**Why**:

This is part of our plan to add extended severity information.

**How did you test it**:

Tested by running `gvmd --rebuild` with an up to date scanner and NVTs feed and then checking the response to a `get_info` GMP command getting the details of an NVT that has the new fields, like `1.3.6.1.4.1.25623.1.0.117034`.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_sql_nvts.c
+++ b/src/manage_sql_nvts.c
@@ -1374,16 +1374,29 @@ nvti_from_vt (entity_t vt)
             }
           else
             {
+              entity_t origin, severity_date;
               double cvss_base_dbl;
               gchar * cvss_base;
+              time_t parsed_severity_date;
 
               cvss_base_dbl
                 = get_cvss_score_from_base_metrics (entity_text (value));
 
+              origin
+                = entity_child (severity, "origin");
+              severity_date
+                = entity_child (severity, "date");
+              
+              if (severity_date)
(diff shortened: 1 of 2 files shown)


src/manage_sql_nvts.c
```diff
@@ -1374,16 +1374,29 @@ nvti_from_vt (entity_t vt)
...
+              origin
+                = entity_child (severity, "origin");
+              severity_date
+                = entity_child (severity, "date");
+              
+              if (severity_date)
+                parsed_severity_date = strtol (entity_text (severity_date),
+                                               NULL, 10);
+              else
+                parsed_severity_date = nvti_modification_time (nvti);
```
jjnicola, line 1394: This should use the nvti creation date.

jjnicola (changes requested):

src/manage_sql_nvts.c
timopollmeier: Changed in aa5b0edc00aed24ae3007c6d44f07a1586ebe5b2.

jjnicola (approved):

PR #1361, opened 2020-11-23
Speed up result counting

mattmundell:
**What**:

Speed up result counting iterator.

2 parts:
1. Use result_new_severities_dynamic/result_new_severities_static instead of result_new_severities when the dynamic state is known.
2. Use LATERAL to save having the severity calculation repeated in the WHERE clause (where it was not accessing the WITH optimisations).

Test case was an `--optimize rebuild-report-cache` that took 32s on master.

With 1 the test case dropped to 23s, with 1 and 2 it dropped to under 4s.

**Why**:

Immediate problem: Result counting slows down the report upload used to generate big dbs.

**How did you test it**:

1. `time gvmd --optimize rebuild-report-cache` on master (32s)
2. `psql gvmd -c "SELECT report, user, severity, count, override, end_time, min_qod from report_counts;" > /tmp/rc_master`
3. `time gvmd --optimize rebuild-report-cache` with patch (4s)
4. `psql gvmd -c "SELECT report, user, severity, count, override, end_time, min_qod from report_counts;" | sort > /tmp/rc_pr`
5. `sort /tmp/rc_master | diff - /tmp/rc_pr` (same)

**Checklist**:

- [ ] Tests


timopollmeier (approved):
I can confirm that it speeds up the cache rebuild a lot and uploading reports feels faster as well.

PR #1372, opened 2020-12-07
Fix rsync errors due to IP blocking.

cfi-gb:
**What**:

Replaces #326. I have also updated this PR so that:
- the sleep is only done for the GCF and not for the GSF
- a 5 seconds sleep is used according to https://community.greenbone.net/t/connection-refused-111-greenbone-nvt-sync/7457/13
- a short sleep before is_feed_current is used as well

Also see **Why** below

**Why**:

From #326:

> The synchronization scripts do two rsyncs in quick succession, this leads to frequent IP blocking by the rsync server.
>
> Fixed by adding a short sleep before the second (synchronization) rsync.

See various ongoing topics at the community portal like e.g.:

https://community.greenbone.net/t/connection-refused-111-greenbone-nvt-sync/7457

or the related PR for openvas here: greenbone/openvas/pull/335

**How did you test it**:

No testing done as it just adds an additional sleep.

**Checklist**:

- [ ] Tests

### DIFF
--- a/tools/greenbone-feed-sync.in
+++ b/tools/greenbone-feed-sync.in
@@ -377,6 +377,9 @@ is_feed_current () {
     fi
     remove_tmp_key
   else
+    # Sleep for five seconds (a previous feed might have been synced a few seconds before) to prevent
+    # IP blocking due to network equipment in between keeping the previous connection too long open.
+    sleep 5
     log_notice "No Greenbone Security Feed access key found, falling back to Greenbone Community Feed"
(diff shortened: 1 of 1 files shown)


cfi-gb:
@greenbone/gvm-dev I would like to ask for a timely review of this because this is an ongoing issue with repeated postings at the community portal which could be easily prevented by adding a short sleep.

mattmundell (approved):

mattmundell (approved):

PR #1383, opened 2020-12-21
Expect report format scripts to exit with code 0

timopollmeier:
**What**:
To detect more errors, run_report_format_script will handle any non-zero
exit code from the generate script as a failure.

**Why**:
Being stricter about the return codes makes it easier to detect if errors occur running the report format scripts.

**How did you test it**:
I modified a report format `generate` script to return a non-zero exit code and loaded a report in GSA using it.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_sql_report_formats.c
+++ b/src/manage_sql_report_formats.c
@@ -3451,9 +3451,8 @@ run_report_format_script (gchar *report_format_id,
                 }
 
               ret = system (command);
-              /* Ignore the shell command exit status, because we've not
-                * specified what it must be in the past. */
-              if (ret == -1)
+              /* Report scripts should return 0 since version 21.04 */
+              if (ret)
                 {
                   g_warning ("%s (child):"
                               " system failed with ret %i, %i, %s",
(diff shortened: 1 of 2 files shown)


src/manage_sql_report_formats.c
```diff
@@ -3551,9 +3550,8 @@ run_report_format_script (gchar *report_format_id,
       /* Just run the command as the current user. */
 
       ret = system (command);
-      /* Ignore the shell command exit status, because we've not
-        * specified what it must be in the past. */
-      if (ret == -1)
+      /* Report scripts should return 0 since version 21.04 */
+      if (ret)
```
mattmundell, line 3554: The return from system is either -1 or a "wait status".  Should you be checking for -1 and using WEXITSTATUS?

mattmundell (changes requested):

src/manage_sql_report_formats.c
timopollmeier: I've changed this in e96859a84babecf7861ec7e1ef693b60333a0357

mattmundell (approved):

PR #1391, opened 2021-01-15
Replace g_file_test with gvm-libs file tests (20.08)

timopollmeier:
**What**:
This replaces the `g_file_test` calls with the functions gvm_file_exists
gvm_file_is_executable and gvm_file_is_readable, which were added
in greenbone/gvm-libs#422.

**Why**:
The new functions use the effective UID/GID when checking if a file is
accessible so gvmd can be used with the file flags to set UID and GID.

**How did you test it**:
* Running gvmd with the `--relay-mapper` option with
  * a file that does not exists
  * a file the user has no permissions for
  * a file the user can access but that is not executable
  * an executable file the user can access
* Exporting a report

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -9297,7 +9297,7 @@ alert_script_exec (const char *alert_id, const char *command_args,
 
   script = g_build_filename (script_dir, "alert", NULL);
 
-  if (!g_file_test (script, G_FILE_TEST_EXISTS))
+  if (! gvm_file_is_readable (script))
     {
       g_warning ("%s: Failed to find alert script: %s",
            __func__,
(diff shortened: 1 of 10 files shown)


src/manage_sql.c
```diff
@@ -9297,7 +9297,7 @@ alert_script_exec (const char *alert_id, const char *command_args,
 
   script = g_build_filename (script_dir, "alert", NULL);
 
-  if (!g_file_test (script, G_FILE_TEST_EXISTS))
+  if (! gvm_file_is_readable (script))
```
mattmundell, line 9300: Not sure about the space, everywhere else uses !call.

mattmundell (approved):

src/manage_sql.c
timopollmeier: Changed this in 2ffceab.

mattmundell (approved):

PR #1395, opened 2021-01-20
Improve "Failed to find..." messages (20.08)

timopollmeier:
**What**:
This fixes the following issues with the "Failed to find..." messages.

1. If a SecInfo entry cannot be found by ID or name, the error message will
say it cannot find the given ID or name instead of a message that the
type could not be found.
2. XML escaping is applied to the error messages.
3. Error messages if a filter given by `filt_id` cannot be found now say "filter"
 instead of using the type of the `get_` command.

**Why**:
1. The previous message was misleading and should only be used if
an invalid `info_type` was used.
2. When requesting a name or id with quote marks and possibly other
 reserved characters, the response would be invalid.
3. The message was wrong and misleading.

**How did you test it**:
1. Tested by requesting a nonexistent CPE via GSA.
2. Tested by requesting a an id containing a quote mark via gvm-cli.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -13112,8 +13112,13 @@ handle_get_info (gmp_parser_t *gmp_parser, GError **error)
       switch (ret)
         {
         case 1:
-          if (send_find_error_to_client ("get_info", "type",
-                                         get_info_data->type,
+          if (send_find_error_to_client ("get_info",
+                                         get_info_data->name
+                                          ? "name"
+                                          : "ID",
+                                         get_info_data->name
+                                          ? get_info_data->name
+                                          : get_info_data->get.id,
                                          gmp_parser))
             {
               error_send_to_client (error);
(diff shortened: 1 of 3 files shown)


src/gmp.c
```diff
@@ -13112,8 +13112,13 @@ handle_get_info (gmp_parser_t *gmp_parser, GError **error)
       switch (ret)
         {
         case 1:
-          if (send_find_error_to_client ("get_info", "type",
-                                         get_info_data->type,
+          if (send_find_error_to_client ("get_info",
+                                         get_info_data->name
+                                          ? "name"
```
mattmundell, line 13117: I think this would break if name/id contained a ".

It was safe for type because it's checked above.

mattmundell (changes requested):

src/gmp.c
timopollmeier: I think this is a general issue with `send_find_error_to_client` that was only avoided so far because of the id validation in GSA, so I'll change that function to XML escape the input.

timopollmeier: The issue should be addressed with 9504567.

mattmundell (approved):

PR #1400, opened 2021-02-01
Deprecate pg

s-l-teichmann:
**What**:
Remove functions from gvmd which are now in the PostgreSQL extension.

**Why**:

The functions moved to another place.

**How did you test it**:

**Checklist**:
- [ ] Tests N/A


timopollmeier (approved):
The calls to create or replace the functions need to be removed as well, but I'll add this in a separate PR.

PR #1405, opened 2021-02-03
add <detection></detection> on report import

nichtsfrei:
**What**:
When importing a report with <detection> and even if it has the correct
nvt as a host_detail as well as an result within the report the
detection block was missing when importing via create_report.

To add those sections two new entries within host_details will be added
when a <detection> got found within a result.

**Why**:

**How did you test it**:

To test it you need to have a report with
- result/detection
in it, the source_oid of those detection must be also available as result nvt oid entry.

With that given you need to import that report into an container task and then download the report via gsa.

**Checklist**:

- [ ] Tests


src/gmp.c
```diff
@@ -7219,6 +7234,7 @@ gmp_xml_handle_start_element (/* unused */ GMarkupParseContext* context,
             create_report_data->host_ends = make_array ();
             create_report_data->host_starts = make_array ();
             create_report_data->results = make_array ();
+			create_report_data->result_detection = make_array ();
```
y0urself, line 7237:
```suggestion
             create_report_data->result_detection = make_array ();
```

src/gmp.c
```diff
@@ -7375,23 +7390,58 @@ gmp_xml_handle_start_element (/* unused */ GMarkupParseContext* context,
...
+          set_client_state (CLIENT_CREATE_REPORT_RR_RESULTS_RESULT_DETECTION);
         ELSE_READ_OVER;
+      case CLIENT_CREATE_REPORT_RR_RESULTS_RESULT_DETECTION:
+        if (strcasecmp ("RESULT", element_name) == 0)
+          {
+            set_client_state (
+              CLIENT_CREATE_REPORT_RR_RESULTS_RESULT_DETECTION_RESULT);
+          }
+       ELSE_READ_OVER; 
+      case CLIENT_CREATE_REPORT_RR_RESULTS_RESULT_DETECTION_RESULT:
```
y0urself, line 7419:
```suggestion
     case CLIENT_CREATE_REPORT_RR_RESULTS_RESULT_DETECTION_RESULT:
```

src/gmp.c
```diff
@@ -18527,6 +18577,144 @@ handle_modify_config (gmp_parser_t *gmp_parser, GError **error)
...
+      else if (strcasecmp (create_report_data->result_threat, "Low") == 0)
+        create_report_data->result_severity = strdup ("2.0");
+      else if (strcasecmp (create_report_data->result_threat, "Log") == 0)
+        create_report_data->result_severity = strdup ("0.0");
+      else if (strcasecmp (create_report_data->result_threat, "False Positive")
+               == 0)
+        create_report_data->result_severity = strdup ("-1.0");
+      else
+        create_report_data->result_severity = strdup ("");
+    }
```
y0urself, line 18632: Is it common practice to not use `{}` for one line conditions in gvmd? I remember this is veeery error-prone ...  Remembering the SSL Bug by apple https://blog.codecentric.de/en/2014/02/curly-braces/

bjoernricks: Personally I would prefer always using braces even for one line conditions. In the past the gvmd developers avoided the braces in their coding style recommendations.

nichtsfrei: Although I usually also prefer braces, in this `else if` chain construct there isn't that much benefit since it would break compilation with a second line except in the else construct but a
```
if (create_report_data->result_threat == NULL)
{
 ...
}
else
{
  if (...)
  {

  }
 ...
```
would be much harder to read.

y0urself: Huh? So

```C
if (create_report_data->result_threat == NULL)
{
 ...
}
else if (...)
{
  ...
}
 ...
```

wouldn't work?!

nichtsfrei: could be a compromise, yes. 
I still think that for if else chain it is not as critical as suggested within the article though.

y0urself: Honestly I have never ever seen that structure for `if else`
```
else
{
  if (...)
  {

  }
```

timopollmeier: The preferred way to use curly braces for me would be
```c
if (create_report_data->result_threat == NULL)
  {
    ...
  }
else if (...)
  {
    ...
  }
...
```
In a case like this where there are only simple assignments, I'd also be fine with not using curly braces.

bjoernricks: You are free to choose whatever you think works best for you. I wont dictate anything ;-) Just get an agreement.

nichtsfrei: I've seen that construct, mostly in projects with overly strict auto-format-rules though.

Since I don't mind either I've pushed an updated version with braces.

@timopollmeier would be ok for you if we run clang-format with the rules across the project? Since I'm used to just run that, but in gmp.c it would have changed even more :).

src/gmp.c
```diff
@@ -18527,6 +18577,157 @@ handle_modify_config (gmp_parser_t *gmp_parser, GError **error)
...
+          // but may have unwanted sideeffects.
+          // Doubling the  data set.
+          detail = g_malloc (sizeof (host_detail_t));
+          detail->ip = g_strdup (result->host);
+          detail->name = g_strdup ("detected_by");
+          detail->source_desc = g_strdup ("create_report_import");
+          detail->source_name = g_strdup (detection->source_oid);
+          detail->source_type = g_strdup ("create_report_import");
+          detail->value = g_strdup (detection->source_oid);
+          array_add (create_report_data->details, detail);
```
timopollmeier, line 18707: Besides the basic `detected_by` details, there are also `detected_by@[...]` details that should be created here.
These also contain the location (as the `[...]` in the name) in case different versions of a vulnerable product were found at different locations on the same host (like multiple ports or installation paths).

src/gmp.c
```diff
@@ -21104,49 +21305,7 @@ gmp_xml_handle_end_element (/* unused */ GMarkupParseContext* context,
...
       case CLIENT_CREATE_REPORT_RR_ERRORS_ERROR:
         {
-          create_report_result_t *result;
-
-          assert (create_report_data->results);
-
-          if (create_report_data->result_scan_nvt_version == NULL)
-            create_report_data->result_scan_nvt_version = strdup ("");
-
-          if (create_report_data->result_severity == NULL)
```
timopollmeier, line 21114: I'm not sure if this really matters in practice, but I think it would be safer if you still set `result_severity` and `result_threat` to "-3.0" and "Error" like before.

src/gmp.c
```diff
@@ -26478,7 +26589,48 @@ gmp_xml_handle_text (/* unused */ GMarkupParseContext* context,
       APPEND (CLIENT_CREATE_REPORT_RR_RESULTS_RESULT_THREAT,
               &create_report_data->result_threat);
 
-
+	  case CLIENT_CREATE_REPORT_RR_RESULTS_RESULT_DETECTION_RESULT_DETAILS_DETAIL_NAME:
```
timopollmeier, line 26592: This and the following added lines should use spaces instead of tabs.
Also, it would be nice if you could use the format
```
if (...)
  {
    ...
  }
```
instead of
```
if (...){
  ...
}
```

src/manage_sql.c
```diff
@@ -19367,9 +19367,11 @@ result_detection_reference (result_t result, report_t report,
                               "                    WHERE id = %llu);",
                               report, quoted_host, result);
     }
-
-  if (*location == NULL)
+  if (*location == NULL){
```
timopollmeier, line 19370: Another `if` block that should be formatted differently.

src/manage_sql.c
```diff
@@ -19403,8 +19404,9 @@ result_detection_reference (result_t result, report_t report,
                       report, quoted_host, result);
     }
 
-  if (*oid == NULL)
-    goto detect_cleanup;
+  if (*oid == NULL){
```
timopollmeier, line 19407: ... and another one.

timopollmeier (changes requested):
One thing I'm missing are the `detected_by@...` details I've noted below.
Other than that and a few things about code formatting, the PR already looks good to me.

src/gmp.c
y0urself: I don't care at all, too.
Just wanted to mention the error-prone-ness of the one-line-conditions ... :)

nichtsfrei:
> One thing I'm missing are the `detected_by@...` details I've noted below.
> Other than that and a few things about code formatting, the PR already looks good to me.

you're right changed it to use `detected_by@...`

timopollmeier (approved):

PR #1406, opened 2021-02-04
Also create owner WITH clause for single resources

timopollmeier:
**What**:
The function init_get_iterator2_with will now also use acl_where_owned
to generate a WITH clause in case permissions_subject is used in a
subquery.
The init_result_get_iterator_severity had to be adjusted to not generate
the same WITH clause twice.

**Why**:
This change fixes an error getting details pages for types use the
WITH clauses for permission checks.

**How did you test it**:
I confirmed it fixed the error by following the link to an OS asset from a host asset in GSA.
I also quickly checked other pages for regressions, including the report details.

**Checklist**:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


bjoernricks (approved):
Seems legit. Can't say I understand the full details of the code change.

nichtsfrei (approved):

PR #1434, opened 2021-03-03
Update report run status more consistently

timopollmeier:
**What**:
To ensure the task status changes in fork_osp_scan_handler,
stop_osp_task and stop_task_internal also update the report status
before events are triggered, they will set current_scanner_task and
global_current_report temporarily / within the forked process.

**Why**:
Because the report status was only updated after the status change event,
the "Severity at least" condition on finished reports was evaluated for the
previous report, not the latest one.

**How did you test it**:
I tested this by checking current_scanner_task and global_current_report
set_task_run_status_internal and running a task with an alert that is triggered
when a task changes to "Done" and that has a minimum severity.
In the test runs I switched the target on and off to get different severity scores.

**Checklist**:

- [ ] Tests


bjoernricks (approved):

timopollmeier:
@Mergifyio refresh

timopollmeier:
@Mergifyio backport gvmd-21.04

PR #1435, opened 2021-03-05
Improve modify_override errors, fix no NVT case

timopollmeier:
**What**:
The command will now return relevant syntax errors if the threat or
severity elements are invalid or required ones are missing.
Also, the case where no NVT OID is given has been fixed.

**Why**:
The improved error messages help clarify to users what kind of input
is expected and the NVT case make the behavior consistent with the
GMP specification.

**How did you test it**:
By running the command via gvm-pyshell.

**Checklist**:

- [ ] Tests


bjoernricks (approved):
Looks good to me

timopollmeier:
@Mergifyio refresh

Issue #1188, opened 2020-07-10
Short network connection error will automatically stop running scan task  (possible redis issue)


wisukind:
**Environment:**
```
Ubuntu 18.04
Greenbone Vulnerability Manager 9.0.1~git-e250176b-gvmd-9.0
GIT revision e250176b-gvmd-9.0
Manager DB revision 221
gvm@ov-master-eqi:~$ gsad --version
Greenbone Security Assistant 9.0.1~git-9fb2e63cd-gsa-9.0
gvm@ov-master-eqi:~$ openvas --version
OpenVAS 7.0.1
gvm-libs 11.0.1
```

**Current Behaviour:**
When using a master / slave architecture; a short network connection error with the slave will Stop related running tasks automatically. This happens randomely, and is not related to the similar bug #1061 
 
**Expected bahviour:**
Short network connection error should not interrupt running tasks. Gvmd should wait a bit longer before considering the network connection done.

On gvmd the following error would be notified:
```
lib  serv:WARNING:2020-07-09 13h49.49 UTC:8819: Failed to connect to server
md manage:WARNING:2020-07-09 13h49.49 UTC:8819: Could not connect to Scanner at ov-slave-eqi:9390
lib  serv:WARNING:2020-07-09 13h49.54 UTC:8819: Failed to connect to server
md manage:WARNING:2020-07-09 13h49.54 UTC:8819: Could not connect to Scanner at ov-slave-eqi:9390
lib  serv:WARNING:2020-07-09 13h49.56 UTC:8819: Failed to connect to server
md manage:WARNING:2020-07-09 13h49.56 UTC:8819: Could not connect to Scanner at ov-slave-eqi:9390
event task:MESSAGE:2020-07-09 13h49.57 UTC:8819: Status of task Pacy, All Networks, Offices (93694043-92df-4f1e-a194-0457b7d78191) has changed to Stopped
```
On the ospd-openvas remote side; no particular errors is logged but 0 results are returned:

```
2020-07-09 13:49:02,624 OSPD - openvas: DEBUG: (ospd.ospd) Returning 0 results
2020-07-09 13:49:02,751 OSPD - openvas: DEBUG: (ospd.server) New connection from ('10.194.157.7', 39850)
2020-07-09 13:49:17,800 OSPD - openvas: DEBUG: (ospd.server) New connection from ('10.194.157.7', 39860)
[...]

PR #1452, opened 2021-03-17
Add auto retry on scanner connection lost during a runnning task

jjnicola:
**What**:
￼Add auto retry on scanner connection lost during a runnning task

**Why**:
During a running scan, gvmd set immediately the task as "stopped". This patch adds an option to retry to connect.

**How did you test it**:
- Start a scan.
- chmod the ospd-openvas socket with 400 permission.
- check the logs
- 

**Checklist**:

- [ ] Tests
- [X] [CHANGELOG](https://github.com/greenbone/gvmd/blob/master/CHANGELOG.md) Entry

### DIFF
--- a/doc/gvmd.8.xml
+++ b/doc/gvmd.8.xml
@@ -385,8 +385,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
     <option>
       <p><opt>--scanner-connection-retry=<arg>NUMBER</arg></opt></p>
       <optdesc>
-        <p>During a running task, number of auto retry on scanner connection
-           lost, default 3.</p>
+        <p>Number of auto retries if scanner connection is lost
+           in a running task.</p>
       </optdesc>
     </option>    
     <option>
--- a/doc/gvmd.8.xml
+++ b/doc/gvmd.8.xml
@@ -383,6 +383,13 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
       </optdesc>
     </option>
     <option>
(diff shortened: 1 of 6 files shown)


doc/gvmd.8.xml
```diff
@@ -382,6 +382,13 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
            as used in GMP.</p>
       </optdesc>
     </option>
+    <option>
+      <p><opt>--scanner-connection-retry=<arg>NUMBER</arg></opt></p>
+      <optdesc>
+        <p>During a running task, number of auto retry on scanner connection
```
timopollmeier, line 388: I think a better way to phrase this would be
"Number of auto retries if scanner connection is lost in a running task."

src/manage.c
```diff
@@ -4648,6 +4685,29 @@ run_osp_task (task_t task, int from, char **report_id)
   return 0;
 }
 
+/**
+ * @brief Get the number of retry on a scanner connection lost.
```
timopollmeier, line 4689: This and the other comments below should say "number of retries" instead of "number of retry".

timopollmeier (changes requested):
Overall this looks okay aside from a few possible improvements to the documentation and two other small issues.

timopollmeier (approved):

cfi-gb:
@jjnicola @timopollmeier Could this also solve https://github.com/greenbone/gvmd/issues/1188 ?

Issue #1385, opened 2020-12-22
[GVE 20.08] CVE reports scan finished just after start without any findings


TatusGepetto:
Running CVE scan finished immidiately without any new results. Only first scan finished correctly, the new one is ended as described.

### Expected behavior
New scan runs properly and report new vulnerabilities. 

### Actual behavior
CVE scan ends immidietly.:

```
event task:MESSAGE:2020-12-22 11h35.29 UTC:1039075: Task Test-scan-TCcloud (7d5ca43f-3ee6-4eb9-8dc3-a8fbd3c1064d) has been requested to start by jwasilewski
event task:MESSAGE:2020-12-22 11h35.29 UTC:1039078: Status of task Test-scan-TCcloud (7d5ca43f-3ee6-4eb9-8dc3-a8fbd3c1064d) has changed to Running
event task:MESSAGE:2020-12-22 11h35.30 UTC:1039078: Status of task Test-scan-TCcloud (7d5ca43f-3ee6-4eb9-8dc3-a8fbd3c1064d) has changed to Done
```
### Steps to reproduce

1. Run CVE scan
2. Wait till finish(it takes few minutes for 1 IP address)
3. Execute this can again
4. Check how fast it is finished. 

### GVM versions

**gsa:** (gsad --version)
gsad --version
Greenbone Security Assistant 20.08.0~git-18ce9861e-gsa-20.08

**gvm:** (gvmd --version)
root@openvas-scanner-20-08:/opt/gvm/var/log/gvm# gvmd --version
Greenbone Vulnerability Manager 20.08.0~git-e805fc0c-gvmd-20.08
GIT revision e805fc0c-gvmd-20.08
Manager DB revision 233
Copyright (C) 2010-2020 Greenbone Networks GmbH
License: AGPL-3.0-or-later
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

**openvas-scanner:** (openvas --version)
root@openvas-scanner-20-08:/opt/gvm/var/log/gvm# openvas --version
OpenVAS 20.8.0
gvm-libs 20.8.0~git-7a9a5b4-gvm-libs-20.08
Most new code since 2005: (C) 2020 Greenbone Networks GmbH
Nessus origin: (C) 2004 Renaud Deraison <deraison@nessus.org>
License GPLv2: GNU GPL version 2
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

[...]

PR #1462, opened 2021-03-24
Fix report host end time check in CVE scans

timopollmeier:
**What**:
When looking up the latest finished report_host for a CVE scan, the
query will also consider the end_time being set to 0 instead of NULL
when the scan of the host is unfinished.

**Why**:
Without this the CVE scan could return no or less results for hosts
being scanned at the moment.

**How did you test it**:
* ran an authenticated "Full and Fast" scan of a single host with vulnerable applications, 
* confirmed that the applications were detected after the scan was finished
* ran a CVE scan using the same target and confirmed that it produced results
* started the "Full and Fast" task again
* started the CVE scan again while the other task was running and at a low progress percentage (1-2%)

Without the fix the second CVE scan produced no results, while it produces the same results as the first scan with the fix.

**Checklist**:

- [ ] Tests


bjoernricks (approved):

cfi-gb:
@timopollmeier @bjoernricks Looks like this could also fix https://github.com/greenbone/gvmd/issues/1385 or am i wrong?

PR #1465, opened 2021-03-24
Improve report format sync handling

bjoernricks:
**What**:

Don't crash if GError is not initialized and reduce missing report format dir to just a message output.

**Why**:

**How did you test it**:

**Checklist**:

- [ ] Tests


bjoernricks:
After discussion the patch seems to be obsolete.

PR #1466, opened 2021-03-26
Don't require report format for XML report

bjoernricks:
**What**:

Allow to download the XML report without having to install the XML
report format plugin which actually does nothing.

**Why**:

Remove GOS specific patch.

**How did you test it**:
This can be tested by deleting the XML report format and trying to access a report.
Without the patch the get_reports command will fail with a 404 error message:
"Failed to find report format 'a994b278-1f62-11e1-96ac-406186ea4fc5'" while
the XML can be fetched if no report format is selected.

**Checklist**:

- [ ] Tests


timopollmeier (commented):
Overall the patch works and I've fixed the clang static analyzer warning as well as some small formatting issues I found.
I've also added my way to test this to the PR description.

bjoernricks:
@timopollmeier thanks so much! More then I've ever expected.

PR #1472, opened 2021-04-01
Add --feed-lock-timeout option

timopollmeier:
**What**:
This new option sets a number of seconds to retry for if the feed is
locked in contexts (like migration or rebuilds) that do not retry on
their own (like automatic syncs).

**Why**:
This is intended to help automated upgrades of GVM modules where
ospd-openvas needs to be restarted, which locks the feed while the
VTs are reloaded.

**How did you test it**:
By doing the following:
* locking the feed lockfile and running `gvmd --rebuild-scap` - this should fail immediately.
* retrying with `gvmd --rebuild-scap --feed-lock-timeout 15` with the lock still active - this should generate a gvmd log message "Feed is currently locked..." and fail about 15 seconds later.
* running `gvmd --rebuild-scap --feed-lock-timeout 15` with the lock still active and releasing the lock while gvmd is waiting - this should leave a "Feed is currently locked..." gvmd log message and start the rebuild once the lock is released. 
* running `gvmd --rebuild-scap --feed-lock-timeout 15` without locking - this should start the rebuild immediately without a "Feed is currently locked..." log message.

**Checklist**:

- [ ] Tests

### DIFF
--- a/doc/gvmd.8
+++ b/doc/gvmd.8
@@ -61,6 +61,9 @@ Disable task scheduling.
 \fB--feed-lock-path=\fIPATH\fB\f1
 Sets the path to the feed lock file.
 .TP
+\fB--feed-lock-timeout=\fITIMEOUT\fB\f1
+Sets the number of seconds to retry for if the feed is locked in contexts (like migration or rebuilds) that do not retry on their own (like automatic syncs). Defaults to 0.
+.TP
 \fB-f, --foreground\f1
 Run in foreground.
 .TP
@@ -181,9 +184,6 @@ Time out tasks that are more than TIME minutes overdue. -1 to disable, 0 for min
 .TP
 \fB--secinfo-commit-size=\fINUMBER\fB\f1
 During CERT and SCAP sync, commit updates to the database every NUMBER items, 0 for unlimited.
-.TP
-\fB--slave-commit-size=\fINUMBER\fB\f1
-During slave updates, commit after every NUMBER updated results and hosts, 0 for unlimited.
 .TP
 \fB-c, --unix-socket=\fIFILENAME\fB\f1
 Listen on UNIX socket at FILENAME.
--- a/doc/gvmd.8
+++ b/doc/gvmd.8
@@ -62,7 +62,7 @@ Disable task scheduling.
 Sets the path to the feed lock file.
 .TP
 \fB--feed-lock-timeout=\fITIMEOUT\fB\f1
-Sets the number of seconds to retry for if the feed is locked in contexts (like migration or rebuilds) that do not retry on their own (like automatic syncs). Defaults to 0.
+Sets the number of seconds to retry for if the feed is locked in contexts (like migration or rebuilds) that do not retry on their own (like automatic syncs). Defaults to 0 (no retry).
 .TP
 \fB-f, --foreground\f1
 Run in foreground.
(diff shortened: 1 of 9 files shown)


doc/gvmd.8
```diff
@@ -182,9 +185,6 @@ Time out tasks that are more than TIME minutes overdue. -1 to disable, 0 for min
 \fB--secinfo-commit-size=\fINUMBER\fB\f1
 During CERT and SCAP sync, commit updates to the database every NUMBER items, 0 for unlimited.
 .TP
-\fB--slave-commit-size=\fINUMBER\fB\f1
-During slave updates, commit after every NUMBER updated results and hosts, 0 for unlimited.
-.TP
```
bjoernricks, line 187: Has this been removed by accident or on purpose? Maybe it belongs to another PR?

src/manage.c
```diff
@@ -6345,7 +6419,7 @@ gvm_migrate_secinfo (int feed_type)
       return -1;
     }
 
-  ret = feed_lockfile_lock (&lockfile);
```
bjoernricks, line 6348: What about the old function? Is is still used or can it be removed?

src/manage.c
```diff
@@ -5995,6 +6025,50 @@ feed_lockfile_lock (lockfile_t *lockfile)
...
+                         " will retry until %s.",
+                         __func__, iso_time (&timeout_end));
+            }
+          gvm_sleep (1);
+        }
+      else if (ret)
+        {
+          return ret;
+        }
+    } while (ret);
```
bjoernricks, line 6064: What's the meaning of ret here? 1 == locked? If yes maybe you could rename it to somthing like is_locked.

src/gvmd.c
```diff
@@ -1833,6 +1834,12 @@ gvmd (int argc, char** argv)
           &feed_lock_path,
           "Sets the path to the feed lock file.",
           "<path>" },
+        { "feed-lock-timeout", '\0', 0, G_OPTION_ARG_INT,
+          &feed_lock_timeout,
+          "Sets the number of seconds to retry for if the feed is locked"
+          " in contexts (like migration or rebuilds) that do not retry"
+          " on their own (like automatic syncs). Defaults to 0.",
```
bjoernricks, line 1841: I would explain the meaning of 0 too (i guess it is nowait).

bjoernricks (changes requested):

doc/gvmd.8
timopollmeier: This option is supposed to be removed. I think it was added back in by accident in a backport or it was just forgotten before.

src/manage.c
timopollmeier: The old function is still used in the automatic sync that has essentially its own retry loop.
I've also changed `feed_lockfile_lock_timeout` to use it now so writing the timestamp doesn't have to be duplicated.

src/manage.c
timopollmeier: I've renamed the variable and added a few comments about the conditions.

src/gvmd.c
timopollmeier: I've updated the text to mention that the default 0 means there is no retry.

bjoernricks (approved):

timopollmeier:
@mergifyio refresh

timopollmeier:
@mergifyio backport gvmd-20.08

PR #1476, opened 2021-04-07
Change CERT and SCAP back to using CVSS instead of integer scores

timopollmeier:
**What**:
The 0 - 100 integer `score` element is replaced with a CVSS `severity`
one as gvmd will continue using the previous severity scoring system.
This still changes the name of the element in SCAP and CERT info
compared to 20.08 to make the element names more consistent.

**Why**:
The new integer scores were not applied consistently and it's simpler to revert to the previous scoring system than to adjust to the new one for the upcoming release.

**How did you test it**:
By checking the non-severity fields in the list and details pages in GSA and the severity field in the XML responses. 

**Checklist**:

- [ ] Tests


src/gmp.c
```diff
@@ -8998,19 +8997,19 @@ results_xml_append_nvt (iterator_t *results, GString *buffer, int cert_loaded)
             assert (0);
           if (!next (&iterator))
             abort ();
+          severity = ovaldef_info_iterator_severity (&iterator);
           buffer_xml_append_printf (buffer,
                                     "<nvt oid=\"%s\">"
                                     "<type>ovaldef</type>"
                                     "<name>%s</name>"
-                                    "<family/>"
```
bjoernricks, line 9005: Why is the familiy element removed? It seems to be empty but I don't think it belongs to this PR.

src/gmp.c
```diff
@@ -8998,19 +8997,19 @@ results_xml_append_nvt (iterator_t *results, GString *buffer, int cert_loaded)
...
+                                    "<cvss_base>%s</cvss_base>"
```
bjoernricks, line 9005: Did we agree on adding cvss_base field? ~I suppose it would not be necessary to add it here~ I guess you did add it again for backwards compatibility of report formats. Is that correct?

src/manage_pg.c
```diff
@@ -1660,7 +1662,7 @@ create_view_vulns ()
          " AS (SELECT DISTINCT nvt FROM results"
          "     WHERE (results.severity != " G_STRINGIFY (SEVERITY_ERROR) "))"
          " SELECT id, uuid, name, creation_time, modification_time,"
-         "        score, qod, 'nvt' AS type"
+         "        score / 10.0 AS severity, qod, 'nvt' AS type"
```
bjoernricks, line 1665: Will the sorce / 10.0 stay in these queries or will it be changed in another PR?

src/manage_pg.c
```diff
@@ -3263,17 +3267,17 @@ manage_db_init_indexes (const gchar *name)
...
+           " ON scap2.cves (severity);");
 
       sql ("CREATE UNIQUE INDEX cpe_idx"
            " ON scap2.cpes (name);");
       sql ("CREATE INDEX cpes_by_creation_time_idx"
            " ON scap2.cpes (creation_time);");
       sql ("CREATE INDEX cpes_by_modification_time_idx"
            " ON scap2.cpes (modification_time);");
-      sql ("CREATE INDEX cpes_by_score"
-           " ON scap2.cpes (score);");
```
bjoernricks, line 3276: What about removing the existing now obsolete indexes? Do we need to remove them in a migration?

bjoernricks (changes requested):
Just some questions to clarify. Additionally I am not sure about migrations at all. Do we need some? Are they going to be implemented in another PR?

src/gmp.c
timopollmeier: Yes, it's for backward compatibility and to make the nvt elements consistent.

src/manage_pg.c
timopollmeier: Yes, it will be replaced again in #1477 which removes the score from NVTs.

src/manage_pg.c
timopollmeier: For the SCAP and CERT data the schema is fully rebuilt if the version changes, so this should not be necessary.

src/gmp.c
timopollmeier: I removed it because it's always empty and also missing for CVE results.

bjoernricks (approved):
Thanks a lot Timo!

PR #1483, opened 2021-04-12
Try to install PostgreSQL extensions automatically

timopollmeier:
**What**:
The check_db_extensions function will now check if the required
extensions are available to be installed and try to install them
with the database superuser role.

**Why**:
To make the 

**How did you test it**:
By removing the "uuid-ossp" extension with the SQL `DROP EXTENSION "uuid-ossp";`, then running gvmd and checking that the extension is automatically installed again with `SELECT * FROM pg_extension;`.

**Checklist**:

- [ ] Tests


bjoernricks (approved):
Looks nice!

PR #1503, opened 2021-04-20
Allow multilevel DNS names in SAN

lhannigbrinck:
**What**:

Allow more detailed configuration of the subject alternative name (SAN) fields for certificates and certificate requests by adding more environment variables. Also allow multiple values in a single variable separated by `;` for more SAN entries of a type.

**Why**:

For example configuring a DNS name with two level of subdomains is currently not possible currently. This is because of the way the SAN values are split within the single env variable. Configuring a DNS name like `subdomainof.subdomain.example.com` would result in an `ip_address = "subdomainof.subdomain.example.com" ` entry in the template. This entry is then validated as ipv6 address instead of a DNS name and fails `certtool`. There may be other errors with the current approach. 

**How did you test it**:

I created multiple self signed certificates and some CSR with different SAN combinations and verified the created SAN values manually.

**Checklist**:

- [x] Tests


nichtsfrei (approved):

nichtsfrei (changes requested):
Please resolve conflicts with 21.04 branch

nichtsfrei (approved):

PR #1517, opened 2021-05-11
Fix whole-only config family selection

timopollmeier:
**What**:
When modifying the family selection of a scan config, "whole-only" must
be either growing with all VTs or static and empty.
The error message if this is not the case now also includes the name of
the (first) rejected family.

**Why**:
This selection should ensure that these special families are either selected
completely or not at all and the growing/static ensures they stay that way
if new NVTs are added. 

**How did you test it**:
By modifying a config in GSA to include some of the "whole-only" LSC families
and exclude others.
With greenbone/gsa#2905 the config should be accepted, without the changes
there should be an error message for the first unselected "whole-only" family.

**Checklist**:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Improve VT version handling for CVE & OVAL results [#1496](https://github.com/greenbone/gvmd/pull/1496)
 - Fix migration to DB version 242 from gvmd 20.08 [#1498](https://github.com/greenbone/gvmd/pull/1498)
 - Update subject alternative name in certificate generation [#1503](https://github.com/greenbone/gvmd/pull/1503)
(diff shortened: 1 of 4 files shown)


nichtsfrei (approved):

jasons1987:
Hi 

Any news on when this might be resolved?

Thanks

bjoernricks:
It is resolved. The patch is in the 21.04 branch. Are you waiting for a release?

jasons1987:
Im running version 21.04.0

Not sure on how to patch this issue?

bjoernricks:
So I guess you are using some disto packages. In that case we can't do very much for you besides doing a new release https://community.greenbone.net/t/frequently-asked-questions-faq/5558#heading--update-my-gvm-version

jasons1987:
So i this resolved on this GVM OS?

When will there be a new release?

Thanks for the info btw

bjoernricks:
Yes this should be fixed in our GSM products already.

PR #1520, opened 2021-05-14
Rename the date column of reports to creation_time

jhelmold:
**What**:
Changed all relevant entries in manage_sql.c from date
to creation_time. Changed the relevant entries for the
creation of the database in manage_pg.c accordingly.
Added a function for the necessary database changes in
manage_migrators.c. Set the Databaseversion to 245.

**Why**:
Make report timestamps more consistent

**How did you test it**:
Checked scan runs, reports and filters

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_migrators.c
+++ b/src/manage_migrators.c
@@ -2736,42 +2736,6 @@ migrate_244_to_245 ()
 
   sql ("ALTER TABLE reports RENAME COLUMN date TO creation_time;");
 
-
-  sql ("CREATE OR REPLACE FUNCTION task_last_report (integer)"
-       " RETURNS integer AS $$"
-       /* Get the report from the most recently completed invocation of task. */
-       "  SELECT id FROM reports WHERE task = $1 AND scan_run_status = %u"
-       "  ORDER BY creation_time DESC LIMIT 1;"
-       "$$ LANGUAGE SQL;",
-       TASK_STATUS_DONE);
-
-  sql ("CREATE OR REPLACE FUNCTION task_second_last_report (integer)"
-       " RETURNS integer AS $$"
-       /* Get report from second most recently completed invocation of task. */
-       "  SELECT id FROM reports WHERE task = $1 AND scan_run_status = %u"
-       "  ORDER BY creation_time DESC LIMIT 1 OFFSET 1;"
-       "$$ LANGUAGE SQL;",
-       TASK_STATUS_DONE);
-
-  sql ("CREATE OR REPLACE FUNCTION task_severity (integer,"  // task
-       "                                          integer,"  // overrides
-       "                                          integer)"  // min_qod
-       " RETURNS double precision AS $$"
-       /* Calculate the severity of a task. */
-       "  SELECT CASE"
-       "         WHEN (SELECT target = 0"
-       "               FROM tasks WHERE id = $1)"
-       "         THEN CAST (NULL AS double precision)"
-       "         ELSE"
-       "         (SELECT report_severity ((SELECT id FROM reports"
-       "                                   WHERE task = $1"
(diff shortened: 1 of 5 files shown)


src/manage_migrators.c
```diff
@@ -2714,6 +2714,73 @@ migrate_243_to_244 ()
...
+      sql_rollback ();
+      return -1;
+    }
+
+  /* Update the database. */
+
+  sql ("ALTER TABLE reports RENAME COLUMN date TO creation_time;");
+
+
+  sql ("CREATE OR REPLACE FUNCTION task_last_report (integer)"
```
timopollmeier, line 2740: Creating the functions should not be necessary here, as `manage_create_sql_functions` will be called when starting `gvmd` after the migration.

src/manage_pg.c
```diff
@@ -1114,24 +1114,32 @@ manage_create_sql_functions ()
...
-
-      sql ("CREATE OR REPLACE FUNCTION task_second_last_report (integer)"
-           " RETURNS integer AS $$"
-           /* Get report from second most recently completed invocation of task. */
-           "  SELECT id FROM reports WHERE task = $1 AND scan_run_status = %u"
-           "  ORDER BY date DESC LIMIT 1 OFFSET 1;"
-           "$$ LANGUAGE SQL;",
-           TASK_STATUS_DONE);
+      /* column date in table reports was renamed to creation_time in version 245 */
+      if (current_db_version >= 245) {
```
timopollmeier, line 1118: I'd prefer it if you either removed the braces or used the format we use elsewhere:
```c
if (current_db_version >= 245)
  {
    sql ("...");
  }
```

src/manage_sql.c
```diff
@@ -17810,9 +17810,9 @@ task_report_previous (task_t task, report_t report, report_t *previous)
                      "SELECT id FROM reports"
                      " WHERE task = %llu"
                      " AND scan_run_status = %u"
-                     " AND date < (SELECT date FROM reports"
+                     " AND creation_time < (SELECT creation_time FROM reports"
                      "             WHERE id = %llu)"
```
timopollmeier, line 17814: Could you adjust the indentation here?

src/manage_sql.c
```diff
@@ -20964,7 +20964,7 @@ report_add_result (report_t report, result_t result)
      "task_id",                                                              \
      KEYWORD_TYPE_STRING                                                     \
    },                                                                        \
-   { "date", NULL, KEYWORD_TYPE_INTEGER },                                   \
+   { "creation_time", NULL, KEYWORD_TYPE_INTEGER },                          \
```
timopollmeier, line 20967: Could you also add "date" as an alias in this line here or as a new column in `REPORT_ITERATOR_WHERE_COLUMNS` and then add it to the filter columns?

timopollmeier (changes requested):

src/manage_pg.c
```diff
@@ -1115,31 +1115,34 @@ manage_create_sql_functions ()
              "$$ LANGUAGE SQL;");
 
       /* column date in table reports was renamed to creation_time in version 245 */
-      if (current_db_version >= 245) {
+      if (current_db_version >= 245)
+        {
         sql ("CREATE OR REPLACE FUNCTION task_last_report (integer)"
```
timopollmeier, line 1120: If you use braces here, you should also indent the statements another level.

src/manage_sql.c
timopollmeier: the WHERE belongs to the subquery, so you should increase the indentation instead of removing it.

timopollmeier (changes requested):

timopollmeier (approved):

PR #1525, opened 2021-05-18
Removed OVAL definitions from gvmd

jhelmold:
**What**:
Removed all occurences of OVAL-Definitios specific code
from the gvmd sources, especially from gmp.c, manage.c,
manage.h, manage_acl.c, manage_pg.c, manage_sql.c,
manage_sql.h, manage_sql_secinfo.c, manage_sql_secinfo.h
and manage_utils.c. Set GVMD_SCAP_DATABASE_VERSION in
CMakeLists.txt from 18 to 19.

**Why**:
OVAL definitions are not supported any longer.

**How did you test it**:
Tested scans, reports and checked database.

**Checklist**:

- [ ] Tests


src/manage_pg.c
```diff
@@ -1606,41 +1605,14 @@ create_view_vulns ()
 {
   sql ("DROP VIEW IF EXISTS vulns;");
 
-  if (sql_int ("SELECT EXISTS (SELECT * FROM information_schema.tables"
```
timopollmeier, line 1609: This block also handles the version of the `vulns` view that includes CVEs, so instead of removing it entirely you should just adjust the check to use the table `cves` and remove the last `UNION` clause that uses `ovaldefs`.

src/gmp.c
```diff
@@ -13211,36 +13103,8 @@ handle_get_info (gmp_parser_t *gmp_parser, GError **error)
...
-                             ovaldef_info_iterator_file (&info));
-          description = ovaldef_info_iterator_description (&info);
-          if (get_info_data->details == 1)
-            xml_string_append (result,
-                               "<description>%s</description>",
-                               description);
-        }
-      else if (g_strcmp0 ("cert_bund_adv", get_info_data->type) == 0)
+
+      if (g_strcmp0 ("cert_bund_adv", get_info_data->type) == 0)
```
timopollmeier, line 13107: Is there a reason you changed this from `else if` to `if`?

timopollmeier (changes requested):
It looks like you overlooked the CVEs case for the `vulns` view and there's another small change I'm not sure if it's intentional.
Also, you should also remove the file tools/ovaldef_getbyname.xsl as it's no longer used.

timopollmeier (approved):

PR #1538, opened 2021-05-28
Change default for LOCALSTATEDIR to /var

nichtsfrei:
Rather than defaulting to /usr/local/var LOCALSTATEDIR should default to
/var/ since the usage of LOCALSTATEDIR seems to be computer rather
than user specific.

To set it back to the previous default build with

-DLOCALSTATEDIR=/usr/local/var

**What**:

**Why**:

**How did you test it**:

**Checklist**:

- [ ] Tests


timopollmeier (changes requested):
This would also have to be changed in the scanner and gsad and I think you should add some documentation about testing setups and backward compatibility with older installations to the INSTALL.md file.

PR #1579, opened 2021-06-23
Solved a performance problem when filtering results by tags.

jhelmold:
**What**:
In manage_pg.c in function "manage_create_sql_functions ()":
  Set the costs of the function "user_has_access_uuid(...)"
  to IMMUTABLE COST 1000, to cause the PostgreSQL-DB to
  choose a much better execution plan.

**Why**:
The performance was really bad.

**How did you test it**:
Ran filtering results by tags in GSA

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_pg.c
+++ b/src/manage_pg.c
@@ -868,7 +868,8 @@ manage_create_sql_functions ()
        "            AND (is_get OR name = arg_permission)) > 0;"
        "  RETURN ret;"
        " END;"
-       "$$ LANGUAGE plpgsql;");
+       "$$ LANGUAGE plpgsql"
+       " IMMUTABLE COST 1000;");
 
   /* Functions in SQL. */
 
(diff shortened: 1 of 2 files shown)


src/manage_pg.c
```diff
@@ -868,7 +868,8 @@ manage_create_sql_functions ()
        "            AND (is_get OR name = arg_permission)) > 0;"
        "  RETURN ret;"
        " END;"
-       "$$ LANGUAGE plpgsql;");
+       "$$ LANGUAGE plpgsql"
+       " IMMUTABLE COST 1000;");
```
timopollmeier, line 872:
```suggestion
       " STABLE COST 1000;");
```

timopollmeier (changes requested):
Changing the function so it's not evaluated for every row is a good idea but I think we should use STABLE instead of IMMUTABLE.

timopollmeier (approved):

Issue #1608, opened 2021-07-01
[21.4.2] gvmd --rebuild fails to rebuild NVT cache


tgurr:
### Expected behavior

See `su -c "gvmd --rebuild" -s /bin/bash gvm` succeed.

### Actual behavior

After some time after issuing `su -c "gvmd --rebuild" -s /bin/bash gvm` it fails with:

`Failed to rebuild NVT cache.`

In the gvmd.log I have:

```
md   main:MESSAGE:2021-07-01 08h09.09 utc:563653:    Greenbone Vulnerability Manager version 21.4.2 (DB revision 242)
md manage:   INFO:2021-07-01 08h09.09 utc:563653:    Rebuilding NVTs.
md manage:   INFO:2021-07-01 08h12.53 utc:563653: Updating VTs in database ... 74525 new VTs, 0 changed VTs
md manage:WARNING:2021-07-01 08h12.54 utc:563653: update_nvts_from_vts: SHA-256 hash of the VTs in the database (ccf07d7f4f49fddf61efc9def8240753e3d2795dacbcde473fde32e316aaf76c) does not match the one from the scanner (c6e978aa270a0d2a3d03fafb3ef6e1de9fc75a3394ef2cd8ad410b30ca12a857).
```

Note: things appear to work despite the rebuild failure, I can issue scans from GSA and get results. no idea how serious that rebuild failure actually is for the Greenbone stack.

### Steps to reproduce

1. issue `su -c "gvmd --rebuild" -s /bin/bash gvm`.

### GVM versions

**gsa:** Greenbone Security Assistant 21.4.1~dev1
**gvm:** Greenbone Vulnerability Manager 21.4.2 / Manager DB revision 242
**openvas-scanner:** OpenVAS 21.4.1
**gvm-libs:** gvm-libs 21.4.1

### Environment

**Operating system:** `Linux exherbo 5.12.0 #1 SMP Thu Apr 29 14:22:53 CEST 2021 x86_64 GNU/Linux`
**Installation method / source:** Exherbo packages (source-based)

### Logfiles

See above.

tgurr:
I've applied the backport commit from the 21.04 branch: https://github.com/greenbone/gvmd/commit/95dfd863a7d91305824105a33554fc6e3b287e14

Stopped gvmd.service and trying to run:

`su -c "gvmd -m" -s /bin/bash gvm`

I get:

```
[...]

PR #1611, opened 2021-07-05
Fix VTs hash check and add --dump-vt-verification

timopollmeier:
**What**:
The collation of the VTs and preferences is now set explicitly in the
SQL used to get the string the VTs verification hash is calculated from
to ensure the VTs and preferences are sorted consistently.
The unhashed string is now also generated in a function and the command
line option --dump-vt-verification has been added.

**Why**:
With the recent NVTs feed this caused the VTs verification to fail. See #1608.

**How did you test it**:
By running `gvmd --dump-vt-verification`, redirecting the output to a file and comparing it to the output of a modified version of ospd-openvas that also dumps the non-hashed verification string to a file.
I also ran gvmd --rebuild to check that warning about the mismatched hash disappeared.

**Checklist**:

- [ ] Tests


src/gvmd.c
```diff
@@ -2568,6 +2573,25 @@ gvmd (int argc, char** argv)
...
+      proctitle_set ("gvmd: --dump-vt-verification");
+  
+      if (option_lock (&lockfile_checking))
+        return EXIT_FAILURE;
+
+      ret = manage_dump_vt_verification (log_config, &database);
+      log_config_free ();
+      if (ret)
+        {
+          printf ("Failed to rebuild SCAP data.\n");
```
cfi-gb, line 2590: This message seems to not match what the code is actually doing.

cfi-gb (changes requested):

src/gvmd.c
timopollmeier: I've fixed this in ca30e385696bb33255883bad8d6c319d5b98d8dc

bjoernricks (approved):

bjoernricks (approved):

timopollmeier:
@mergifyio backport 21.04

timopollmeier:
@mergifyio backport gvmd-21.04

PR #1633, opened 2021-07-15
Added retry of sensor connection for performance reports on failure.

jhelmold:
Added retry of sensor connection for performance reports on failure.

**What**
In manage.c in function get_osp_performance_string (..):
  The installation of the connection to the sensor via
  osp_connect_with_data (..) and the access to the
  performance data via osp_get_performance_ext (..) are
  now retried several times on failure.

**Why**:
This fixes a bug.

**How did you test it**:
Ran the report that uses the connection several times.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -4009,7 +4009,7 @@ get_osp_performance_string (scanner_t scanner, int start, int end,
   osp_connection_t *connection = NULL;
   osp_get_performance_opts_t opts;
   gchar *error;
-  int connection_retry;
+  int connection_retry, return_value;
 
   host = scanner_host (scanner);
   port = scanner_port (scanner);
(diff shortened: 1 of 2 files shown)


src/manage.c
```diff
@@ -4031,7 +4039,17 @@ get_osp_performance_string (scanner_t scanner, int start, int end,
   opts.titles = g_strdup (titles);
   error = NULL;
 
-  if (osp_get_performance_ext (connection, opts, performance_str, &error))
+  connection_retry = SCANNER_CONNECTION_RETRY_DEFAULT;
+  int return_value = 1;
```
timopollmeier, line 4043: Could you move the declaration to the top of the function and only do the initialization here?

timopollmeier (changes requested):

timopollmeier (approved):

PR #1636, opened 2021-07-22
Add trash columns for target "elevate" credential 

timopollmeier:
**What**:
This adds the columns for the SSH elevate credential and its location
to TARGET_ITERATOR_TRASH_COLUMNS

**Why**:
The columns for the SSH elevate credential were missing for targets in
the trashcan, so getting targets in the trashcan did not work correctly.
Without the fix the shown credentials may be wrong or getting the targets
in the trash may not work at all.
(Fixes AP-1554)

**How did you test it**:
By creating targets with different SSH elevate credentials, then moving
 them and some of the credentials to the trashcan.
When checking the the trashcan page, they should appear with the correct
credentials in particular.

**Checklist**:

- [ ] Tests


bjoernricks (approved):

jhelmold (approved):

jhelmold:
The changes look good to me.

PR #1641, opened 2021-07-26
Report row locking

timopollmeier:
**What**:
Instead of locking the whole table when deleting reports, only the
affected rows in the reports table are locked using
`SELECT ... FOR UPDATE` statements.

**Why**:
The goal if this change is to reduce the amount of deadlocks waiting
 for the reports table (AP-1506).

**How did you test it**:

**Checklist**:

- [ ] Tests


timopollmeier:
The locking has already been changed in a different way in #1710 to address the issues, so I'm closing this.

PR #1666, opened 2021-08-09
Adjusted the lack of the severities in the display of applications.

jhelmold:
Adjusted the lack of the severities in the display of applications.

**What**:
In gmp.c:
  Corrected the assignment to "detail->source_name" in
  function "gmp_xml_handle_result ()".

In manage_sql.c in function "result_detection_reference" (..):
  Added an alternative result uuid for "*ref".

**Why**:
This resolves a bug.

**How did you test it**:
Imported several Reports, ran some tasks and checked the display of the
severity.

**Checklist**:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -112,6 +112,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Deprecated
 ### Removed
 ### Fixed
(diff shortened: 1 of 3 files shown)


timopollmeier (changes requested):
The fix in `gmp_xml_handle_result ()` looks good to me but I'm not sure if we should change `result_detection_reference ()` to use the current result as a fallback for its own detection reference.

timopollmeier (approved):

PR #1669, opened 2021-08-17
 Send the script timeout to the scanner as script preferences instead as server preference

jjnicola:
Close SC-355

Depends on greenbone/gvm-libs#578 and greenbone/openvas-scanner#841

**What**:
 Send the script timeout to the scanner as script preferences instead as server preference

**Why**:
Currently is sent to openvas as `timeout.<OID>`
Now is sent as as a script preference as param with ID 0  subelemet of the vt_single in the vt_selection.

**How did you test it**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -2523,7 +2523,7 @@ launch_osp_openvas_task (task_t task, target_t target, const char *scan_id,
   hosts_str = target_hosts (target);
   ports_str = target_port_range (target);
   exclude_hosts_str = target_exclude_hosts (target);
-  
+
   clean_hosts = clean_hosts_string (hosts_str);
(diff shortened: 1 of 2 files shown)


src/manage.c
```diff
@@ -2578,6 +2578,38 @@ launch_osp_openvas_task (task_t task, target_t target, const char *scan_id,
   if (snmp_credential)
     osp_target_add_credential (osp_target, snmp_credential);
 
+  /* Initialize vts table for vulenrability tests and their preferences */
```
ArnoStiefvater, line 2581:
```suggestion
  /* Initialize vts table for vulnerability tests and their preferences */
```

ArnoStiefvater (approved):

PR #1679, opened 2021-09-03
Fix resume task.

jjnicola:
**What**:
Fix resume task.

Jira: AP-1472

**Why**:
An already finished scan, can not be stopped, and must be deleted directly.
Also, there is a connection issue after asking for the scan status. As workaround
the connection is reset.

**How did you test it**:
- Start a scan with many host.
- Wait until some hosts finish and interrupt the scan changing the permissions in the openvas socket. The task will be set as interrupted.
- Wait until ospd finished the scan successfully
- Change the socket permissions again.
- Resume the scan. 
- Check the logs (ospd, openvas and gvmd).

**Checklist**:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Deprecated
 ### Removed
 ### Fixed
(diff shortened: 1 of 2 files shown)


timopollmeier (commented):
The changes look okay but there's a debug message that should be adjusted.

timopollmeier (approved):

PR #1682, opened 2021-09-08
Send an error message instead of the fallback report text to gsad.

jhelmold:
**What**:
In gmp.c in function handle_get_system_reports (..):
  Adjusted an erroneous if-statement.

In manage.c in function manage_system_report (..):
  Replaced the fallback report text by the appropriate
  error message.

**Why**:
This change is part of a bug-fix.

**How did you test it**:
Tested by a special http-request.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -4091,6 +4091,11 @@ get_osp_performance_string (scanner_t scanner, int start, int end,
 "produce more powerful reports.  Please contact your system administrator\n" \
 "for more information.\n\n"
 
+/**
+ * @brief Get the fallback report as a string.
+ *
+ * @param[in]  fallback_report  The string for the fallback report.
+ */
 static void
 get_fallback_report_string(GString *fallback_report)
 { 
--- a/src/manage.c
+++ b/src/manage.c
@@ -3997,18 +3997,18 @@ credential_full_type (const char* abbreviation)
  * @param[in]  titles           The end titles for the performance report.
  * @param[in]  performance_str  The performance string.
  *
- * @return 0 if successful, 4 could not connect to scanner,
- *         6 failed to get performance report, -1 error
+ * @return 0 if successful, 6 could not connect to scanner or failed to get
+ *         performance report
  */
 static int
 get_osp_performance_string (scanner_t scanner, int start, int end,
-                            const char *titles, gchar **performance_str)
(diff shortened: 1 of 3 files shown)


src/manage.c
```diff
@@ -4465,110 +4516,104 @@ manage_system_report (const char *name, const char *duration,
...
+              return_code = 3;
 
+              g_warning ("%s: Failed to create performance graph -- %s",
+                         __func__, astderr);
+              g_debug ("%s: gvmcg failed with %d", __func__, exit_status);
+              g_debug ("%s: stdout: %s", __func__, astdout);
+              g_debug ("%s: stderr: %s", __func__, astderr);
+            }
+          g_free (command);
+	}
```
timopollmeier, line 4584: The indentation should use only spaces:
```suggestion
        }
```

src/manage.c
```diff
@@ -4465,110 +4516,104 @@ manage_system_report (const char *name, const char *duration,
...
-          g_string_append_printf (buffer,
-                                  "Load average for past minute:     %.1f\n",
-                                  load[0]);
-          g_string_append_printf (buffer,
-                                  "Load average for past 5 minutes:  %.1f\n",
-                                  load[1]);
-          g_string_append_printf (buffer,
-                                  "Load average for past 15 minutes: %.1f\n",
-                                  load[2]);
+          return_code = 7;
```
timopollmeier, line 4547: I think it would be simpler to move creating the fallback report here as `get_osp_performance_string` can only return 0 for success or 6 if the connection fails.

src/manage.c
```diff
@@ -4465,110 +4516,104 @@ manage_system_report (const char *name, const char *duration,
...
+          g_free (command);
+	}
+    }
+
+  if (return_code == 7)
+    {
+      buffer = g_string_new ("");
+      get_fallback_report_string(buffer);
+      *report = g_string_free (buffer, FALSE);
+    }
```
timopollmeier: As mentioned above, I think creating the fallback report can be moved instead of using a special value for `return_code` here.

timopollmeier (changes requested):

timopollmeier (approved):

jhelmold:
The backport to main is done manually by PR #1745.

PR #1691, opened 2021-09-21
Remove hr_name from gvmd

jhelmold:
**What**:
OSP scanners are no longer supported and the corresponding code is
being removed from gvmd. Because the field "hr_name" is only used by
OSP scanners, it is also removed from gvmd.

**Why**:
OSP scanners are no longer supported.

**How did you test it**:
Tested scan configs, scanner preferences, the scanner dialog and scans in GSA.

**Checklist**:

- [ ] Tests


jhelmold:
The CHANGELOG entry is done with PR #1689.

timopollmeier (approved):

PR #1694, opened 2021-10-04
Remove scanner, type from configs, re-add hr_name for nvt preferences

jhelmold:
**What**:
OSP scanners are no longer supported. The field config_type and the fields scanner
and type from the database table configs where only used in connection with OSP
scanners. Accordingly the code corresponding to these fields has been removed from
the gvmd-sources. The corresponding data and the database fields scanner and type
of the table configs have been removed from the database. Besides the hr_name has
been rebuilt for nvt preferences.

**Why**:
OSP scanners are no longer supported.
Addresses AP-1562.

**How did you test it**:
Tested the Scan Configs and the Scanners dialog. Tested the according Trashcan
functionality. Tested the Tasks dialog and ran some scans.

**Checklist**:

- [ ] Tests


timopollmeier (changes requested):
Overall this looks okay but there are a few mostly cosmetic things I'd like to have changed.

PR #1719, opened 2021-11-12
[stable] Start Community Container

y0urself:
**What**:

**Why**:

**How did you test it**:

**Checklist**:

- [ ] Tests


.github/workflows/container.yml
```diff
@@ -0,0 +1,64 @@
+name: Container Image Builds
+
+on:
+  push:
+    branches: [ main ]
```
bjoernricks, line 5:
```suggestion
    branches: [ stable ]
```

bjoernricks: Or `[main, stable, oldstable]`

bjoernricks (changes requested):

.docker/build.Dockerfile
```diff
@@ -0,0 +1,52 @@
+# Define ARG we use through the build
+ARG VERSION=unstable
```
bjoernricks, line 2: Either you must pass stable as argument in the workflows or change it here to stable
```suggestion
ARG VERSION=stable
```

.github/workflows/container.yml
```diff
@@ -0,0 +1,64 @@
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build and push Container image
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          push: true
+          tags: ${{ github.repository }}:${{ env.VERSION }}
+          file: .docker/stable.Dockerfile
```
bjoernricks, line 30: There is no `stable.Dockerfile`. Either we use a common build file for all branches and or you have to rename the `main.Dockerfile`

bjoernricks (changes requested):

.docker/build.Dockerfile
bjoernricks: Still unstable and should be stable.

bjoernricks (approved):

PR #1724, opened 2021-11-15
CHANGE: Use full path GVMD_PID_PATH for PID files, use GVMD_RUN_DIR instead of GVM_RUN_DIR 

timopollmeier:
**What**:
The calls of pidfile_create and pidfile_remove are adjusted to use a
full path defined by GVMD_PID_PATH.

The new path GVMD_RUN_DIR is specific to gvmd and only needs to be shared with other
GVM modules that need to access the gvmd socket.

**Why**:
To make gvmd work correctly with the changes in greenbone/gvm-libs/pull/626 (AP-1789)

**How did you test it**:

**Checklist**:

- [ ] Tests


bjoernricks (approved):

bjoernricks (changes requested):
The dockerfile(s) need changes for the new runtime path

bjoernricks (approved):

PR #1743, opened 2021-11-20
Use deprecation as warning for g_memdup

y0urself:
**What**:

 Fix: g_memdup is depcreated; it is recommended to use g_memdup2 instead

Fix warning:
```
warning: 'g_memdup' is deprecated: Use 'g_memdup2' instead [-Wdeprecated-declarations]
```

**Why**:

**How did you test it**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -396,8 +396,10 @@ get_certificate_info (const gchar* certificate, gssize certificate_len,
                          __func__);
               return -1;
             }
-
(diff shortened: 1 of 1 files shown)


timopollmeier (changes requested):
I think we should rather try addressing the issue and use `g_memdup2` instead.

PR #1768, opened 2021-12-17
Improves the init_preference_iterator SQL query performance

falegk:
**What**:
Changes the old SQL query from the init_preferences_iterator function to improve the query performance.
I've changed a "WHERE" clause with similar JOIN functionality.

**Why**:

I noticed that after the NVT sync (update NVT's cache etc.) this query was taking a lot of time (in some cases 6+ minutes in my VM) to complete and for this reason the whole process was slow.

EXPLAIN ANALYZE Results
Selection_023

**How did you test it**:
- I verified the query with direct queries in the database for different config records and it returns the same results as the old one.
- I rebuilt the code and I run the gvm & I synced the nvts
- _**Maybe it needs more tests**_

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_sql_configs.c
+++ b/src/manage_sql_configs.c
@@ -3421,15 +3421,14 @@ init_preference_iterator (iterator_t* iterator,
                  " UNION"
                  " SELECT nvt_preferences.name, nvt_preferences.value"
                  " FROM nvt_preferences"
-                 " WHERE nvt_preferences.name %s"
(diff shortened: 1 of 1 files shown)


timopollmeier (approved):
Thank you for your contribution, @falegk!

While I couldn't reproduce any cases where the original query took several minutes to finish, I can confirm that the updated one is noticeably faster when run on its own.

PR #1812, opened 2022-04-28
switch to setproctitle

nichtsfrei:
Refactor: move from gvm-lib proctitle-set to libbsd setproctitle

To be less dependenct on gvm-libs and have more robust way to set
proctitle this commit switches from gvm-libs proctitle-set to
setproctitle of libbsd.


bjoernricks (changes requested):
The docker image needs an update for the new dependency too

bjoernricks:
> The docker image needs an update for the new dependency too

Ah sorry it has been added in a previous commit already.

bjoernricks (approved):

bjoernricks:
Btw. an update for the container images was actually necessary https://github.com/greenbone/gvmd/commit/72ced60547f2e58b9031b5c7f6c9838e6744dbe8

cfi-gb:
It seems this change has introduced a duplicated `gvmd: gvmd:` in the process tile like e.g.:

```
gvmd         926  0.0  3.3 429024 273760 ?       SLs  Jun15  18:45 gvmd: gvmd: Wa  --foreground *snip*
```

which has been solved via #2065 now.

PR #1825, opened 2022-05-12
Use g_memdup2 instead of the deprecated g_memdup

timopollmeier:
**What**:

**Why**:
To avoid warnings about the use of a deprecated function.

**How did you test it**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -411,7 +411,11 @@ get_certificate_info (const gchar* certificate, gssize certificate_len,
               return -1;
             }
 
+#if GLIB_CHECK_VERSION(2, 68, 0)
+          cert_truncated = g_memdup2 (certificate, certificate_len);
+#else
           cert_truncated = g_memdup (certificate, certificate_len);
+#endif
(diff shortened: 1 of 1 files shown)


y0urself:
g_memdup2 is available since glib 2.68, but the Checks run with 2.66 ... 
https://github.com/greenbone/gvmd/runs/6402681362?check_suite_focus=true#step:4:36

y0urself:
I wonder how this works in scanner etc.
https://packages.debian.org/source/bullseye/glib2.0 is only available with 2.66
but according to https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1927 this should be backported to 2.66?

timopollmeier:
> g_memdup2 is available since glib 2.68, but the Checks run with 2.66 ... https://github.com/greenbone/gvmd/runs/6402681362?check_suite_focus=true#step:4:36

I've added a version check so it should work with older versions.

y0urself (approved):

jhelmold (approved):

PR #1831, opened 2022-05-23
Add some startup delay for gvmd in container

bjoernricks:
**What**:

Add a 10 seconds startup delay for gvmd in container

**Why**:

For the stable image postgres seems to need some additional time to startup
while already listening for new connections. Therefore our do loops seems to not suffice here.

**How did you test it**:

**Checklist**:

- [ ] Tests


bjoernricks:
Closed in favor of  #1832

PR #1845, opened 2022-07-07
Add command line option --vt-sync-extra-days

timopollmeier:
**What**:
This option makes gvmd fetch VTs up to a set number of days older than
the last feed timestamp when updating the VTs.

**Why**:
This is a workaround to ensure VTs are reloaded if they are added to the feed a few days after their "last modification" time.
The new option should be set to the expected delay. (T3-115)

**How did you test it**:

**Checklist**:

- Tests N/A
- [x] PR merge commit message adjusted

### DIFF
--- a/src/manage_sql_nvts.c
+++ b/src/manage_sql_nvts.c
@@ -59,6 +59,11 @@
 static gchar *osp_vt_update_socket = NULL;
 
 /**
+ * @brief Number of extra days of VTs to fetch during feed sync.
+ */
+static int vt_sync_extra_days;
+
+/**
  * @brief Get the current file socket for OSP NVT update.
  *
  * @return The path of the file socket for OSP NVT update.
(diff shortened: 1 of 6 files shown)


src/manage_sql_nvts.c
```diff
@@ -84,6 +89,31 @@ set_osp_vt_update_socket (const char *new_socket)
...
+ *
+ * @param new_extra_days The new number of extra days of VTs to fetch.
+ */
+void
+set_vt_sync_extra_days (int new_extra_days)
+{
+  if (new_extra_days > 0)
+    vt_sync_extra_days = new_extra_days;
+  else
+    vt_sync_extra_days = 0;
```
cfi-gb, line 114: Could it be possible to set a different default like `7` here? Because otherwise at least community members will have the problem that if they don't know / notice the new `vt-sync-extra-days` parameter (which will definitely happen in most cases) will run into troubles sooner then later.

Another possibility would be to just use a default hardcoded "extra days" of something between 7 and 14 days (there should be no new/changed VT entering the feed with a last_modification date later then that) which would reduce the complexity of the code.

cfi-gb (changes requested):

bjoernricks:
I've serious doubts about this PR. Yes we misused last_modification as a version for the feed. But for gvmd the version and therefore the last_modification date should be completely transparent. It should not need to know if it is a date or not. It could also be a number or hashsum or even something else.

bjoernricks (changes requested):
From a technical perspective I am heavily against this PR. It doesn't fix the issue. It just circumvents it.

bjoernricks:
Oh and it will only improve the issue for users that have updated to a new release. We just can't get rid of updating the last_modification date during the feed deployment.

PR #1885, opened 2022-12-09
Changed the compressing algorithm from xz to gz (gzip).

jhelmold:
**What**:
Changed the compressing algorithm from xz to gz (gzip) because xz does not work on every system.

**Why**:
This is a bug-fix.
Addresses T4-122

**How did you test it**:
Tested manually on nightly GOS appliance and on dev system.

**Checklist**:

- [ ] Tests

### DIFF
--- a/tools/gvm-lsc-deb-creator
+++ b/tools/gvm-lsc-deb-creator
@@ -161,7 +161,7 @@ COPYRIGHT_FILE="${DOC_DATA_DIR}/copyright"
 
 # Create data archive
 cd "${DATA_DIR}"
-tar -C "${DATA_DIR}" -acf "../data.tar.gz" "${HOME_SUBDIR}" "${DOC_SUBDIR}"
+tar -C "${DATA_DIR}" -acfz "../data.tar.gz" "${HOME_SUBDIR}" "${DOC_SUBDIR}"
 
 
 #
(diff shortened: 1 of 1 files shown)


tools/gvm-lsc-deb-creator
```diff
@@ -161,7 +161,7 @@ COPYRIGHT_FILE="${DOC_DATA_DIR}/copyright"
 
 # Create data archive
 cd "${DATA_DIR}"
-tar -C "${DATA_DIR}" -acf "../data.tar.xz" "${HOME_SUBDIR}" "${DOC_SUBDIR}"
+tar -C "${DATA_DIR}" -acf "../data.tar.gz" "${HOME_SUBDIR}" "${DOC_SUBDIR}"
```
y0urself, line 164: Are you even sure that this is actually compressed as an `xz`?
Afaik you need to set the `-J` option for that compression format or pipe the `tar` archive to `xy` or `gzip`?

bjoernricks, line 164: Would be better to set the compression algorithm explicitly to avoid such issues
```suggestion
tar -C "${DATA_DIR}" -acfz "../data.tar.gz" "${HOME_SUBDIR}" "${DOC_SUBDIR}"
```

bjoernricks (approved):

PR #1895, opened 2022-12-30
Fixed the number of results in triggered alerts.

jhelmold:
Fixed the number of results in triggered alerts.

**What**:
Now the number of results for automatically triggered alerts and manually triggered alerts are the same.

In addition now the selected report (not always the latest report) is used when manually triggering an alert.

**Why**:
This is a bug-fix. Addresses T4-420 and T4-411.

**How did you test it**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -12580,7 +12580,7 @@ escalate_2 (alert_t alert, task_t task, report_t report, event_t event,
                    * anyway, to make it easier for the compiler to see. */
                   filter = 0;
                   ret = report_content_for_alert
-                          (alert, 0, task, get,
+                          (alert, report, task, get,
                            "notice_report_format",
                            NULL,
                            /* TXT fallback */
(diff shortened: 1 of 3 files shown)


jhelmold:
https://jira.greenbone.net/browse/T4-420

src/manage_sql.c
```diff
@@ -28602,19 +28602,28 @@ manage_send_report (report_t report, report_t delta_report,
...
 
+      get_copy = *get;
       condition = alert_condition (alert);
       method = alert_method (alert);
+      if (method == ALERT_METHOD_EMAIL)
+        {
+          get_copy.filter = g_strdup_printf ("notes=1 overrides=1"
+                                             " sort-reverse=severity rows=1000");
+	  get_copy.ignore_pagination = 0;
+	}
```
timopollmeier, line 28621: I think this should apply to all cases where an alert is used and the filter should be set in the function calling `manage_send_report` if an alert_id is set. In that case a there may also be a filter defined by the alert that should be used.

timopollmeier (changes requested):

y0urself:
@jhelmold you should enable Auto-merge (squash) when you are done with your PR and use a proper conventional commit message for this. Thank you. :)

timopollmeier (approved):

PR #1899, opened 2023-01-17
Encrypting and decrypting of the RADIUS server secret key.

jhelmold:
Now the RADIUS server secret key is encrypted and no longer stored in the database as plain text. The key is also no longer send to GSA in the response, instead a placeholder is send.

**What**:

**Why**:

This is a bug-fix.
Addresses T4-123.

**How did you test it**:
Tested manually on the local development system.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -53295,6 +53295,8 @@ manage_set_ldap_info (int enabled, gchar *host, gchar *authdn,
 void
 manage_get_radius_info (int *enabled, char **host, char **key)
 {
+  char *secret;
+
   if (enabled)
     *enabled = radius_auth_enabled ();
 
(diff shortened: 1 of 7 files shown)


src/manage_sql.c
```diff
@@ -53337,12 +53345,18 @@ manage_set_radius_info (int enabled, gchar *host, gchar *key)
       g_free (quoted);
     }
 
-  if (key)
+  if (key && strcmp (key, "ThisIsAPlaceholderSecretKey!"))
```
timopollmeier, line 53348: I don't like checking for a placeholder string here. I think this should be handled by GSA and gsad only sending the key if it was changed, so this can be left as it was or maybe changed to check for an empty string.

timopollmeier (changes requested):

timopollmeier (approved):

PR #1938, opened 2023-04-05
Filter duplicate results from the scanner during report generation.

jhelmold:
## What
Implemented a functionality that filters duplicate results that come from the scanner during report generation, so that no results are listed in the report several times.

## Why
This is a bug-fix.

## References
Addresses GEA-51

## Checklist
Manually tested on my local DEV-environment,

- [ ] Tests


timopollmeier (approved):

cfi-gb:
Two follow-ups on this are:
- #1990
- #1991

PR #1939, opened 2023-04-11
Add error result and exit scan if VT list empty

mattmundell:
## What

When starting an osp-openvas scan, if the VT list is empty then exit before contacting the scanner.  Also add an error to the report to give the user a hint.

## Why

If a task was started when the NVT tables were empty, then the task was going straight to "Interrupted", with no reason given.

Note that since #1926 it should be much more rare that the NVT tables are empty, because during rebuild these tables are now truncated inside the transaction that rebuilds them.  It is still possible to truncate them by hand, or to start gvmd and GSA before the NVTs are ready.

## References

GEA-49

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -2465,7 +2465,7 @@ launch_osp_openvas_task (task_t task, target_t target, const char *scan_id,
 
   if (empty) {
     if (error)
-      *error = g_strdup ("Exiting because VT list is empty");
+      *error = g_strdup ("Exiting because VT list is empty (e.g. feed not synced yet)");
     g_slist_free_full (osp_targets, (GDestroyNotify) osp_target_free);
     // Credentials are freed with target
     g_slist_free_full (vts, (GDestroyNotify) osp_vt_single_free);


src/manage.c
```diff
@@ -2461,6 +2463,15 @@ launch_osp_openvas_task (task_t task, target_t target, const char *scan_id,
     }
   cleanup_iterator (&families);
 
+  if (empty) {
+    if (error)
+      *error = g_strdup ("Exiting because VT list is empty");
```
cfi-gb, line 2468: Maybe add a hint why this is happening (e.g. DB rebuild in progress, ...) so that one seeing this message actually knows what's going on?

mattmundell: I'm not sure what to add. Since #1926 this should not happen during rebuild. It could happen on a fresh install, before the rebuild/update has happened, then the user must run greenbone-feed-sync, but the situation is different on GOS (I guess the devices ship with an initial feed and the feed sync is automatic).  Any suggestions?

mattmundell: Sounds good.

timopollmeier (approved):

PR #1975, opened 2023-05-08
Update gvm-lsc-deb-creator

rippledj:
Fixed issue in .deb credential script that throws an error when running script twice for a username to replace a credential

## What

This new bash shell script adds a check in the deb `preinst` script to delete and GVM Local Security Checks user with the same username and GVM comment identifier. Because `postrm` is also executed each time the same package is installed where it is already present, I added a check to not delete the GVM LSC user when doing an "upgrade" (installing same package twice) to prevent the package from deleting the user when run a second time to update credentials.

## Why

The package throws errors when being installed a second time with the same username.  This creates a broken package and blocks the subsequent functionality of the system's package manager.  

- [ ] Tests

- This was tested on Ubuntu 22.04


tools/gvm-lsc-deb-creator
```diff
@@ -1,35 +1,12 @@
...
-# You should have received a copy of the GNU Affero General Public License
-# along with this program.  If not, see <http://www.gnu.org/licenses/>.
-
-# This script generates a Debian package that creates a user for GVM
-# local security checks.
-
-#
-# Variables
-#
-
```
bjoernricks, line 25: Could you keep the copyright header?

tools/gvm-lsc-deb-creator
```diff
@@ -110,14 +88,6 @@ then
...
 fi
 
-#
-# Set up error handling
-#
-handle_error() {
-  echo "DEB package generation failed" >&2
-  exit 1
-}
-trap handle_error ERR
```
bjoernricks, line 120: What's the reason for removing the signal handler?

rippledj: Sorry, I didn't mean to remove those parts, just in a rush.  I will fix those into a new PR also, I have been reading Debian documentation and policy and I feel there is a slightly better solution to the issue of an error on re-installation based on Debian policy workflow, so I will prepare that and prepare a new PR.

bjoernricks (approved):

PR #1988, opened 2023-05-22
Batch all INSERTS into vt_refs

mattmundell:
## What

There are about 500k vt_refs and 80k NVTs.

Previously the INSERTs into vt_refs were batched for each NVT, so there was one INSERT per NVT.

Now the INSERTs are batched across NVTs, so there's one INSERT per 50k vt_refs.

Configurable via definition VT_REFS_BATCH_SIZE:
- 5: 66s
- 50k: 54s
- 100k: 50s

## Why

Saves about 10s of rebuild.

## References

Waits for /gvmd/pull/1980.


timopollmeier (changes requested):
I think for larger batch actions like this it would be good to to have a command line option do set a different batch size.

mattmundell:
OK, I'll add one.

timopollmeier (approved):

PR #1993, opened 2023-05-23
Enabled certificates generation before gvmd start

rizlas:
## What

Added `gnutls-bin` package in order to use `gvm-manage-certs`

## Why

Without certificate generation gvmd will not start with `-a` and `-p` params

## References

This post  is somehow related but it's not using docker. https://forum.greenbone.net/t/configuring-gvmd-for-external-access-tls/13912

I only edited Dockerfile in the main branch. Let me know if it should be modified in other branches too.

### DIFF
--- a/.docker/prod.Dockerfile
+++ b/.docker/prod.Dockerfile
@@ -68,6 +68,9 @@ ARG DEBIAN_FRONTEND=noninteractive
 # Loading scap and cert data
 # xml-twig-tools
 
+# Required for set up certificates for GVM
+# gnutls-bin
+
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
     dpkg \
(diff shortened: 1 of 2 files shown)


.docker/prod.Dockerfile
```diff
@@ -110,6 +114,7 @@ RUN addgroup --gid 1001 --system gvmd && \
 RUN mkdir -p /run/gvmd && \
     mkdir -p /var/lib/gvm && \
     mkdir -p /var/log/gvm && \
+    gvm-manage-certs -a && \
```
bjoernricks, line 117: This would create the same certificates for **all** users of the container image. Therefore I am not going to add this change. If you really need the generation of the certificates in this container you should try adding it to the `.docker/start-gvmd.sh` file instead. Please be aware that adding certificates should stay optional.

```suggestion
```

bjoernricks (changes requested):

.docker/prod.Dockerfile
rizlas: Hi, thank you. I have modified it following your advice.

bjoernricks (approved):

bjoernricks:
It will require a new tag/release

Issue #2006, opened 2023-06-08
sql_exec_internal failed in check_osp_result_exists()


dvppvd:
### Expected behavior
No error.

### Actual behavior

The function [check_osp_result_exists()](https://github.com/greenbone/gvmd/blob/b846590e7d45af6765ed015743b185f6bb6cd1ec/src/manage_sql.c#L29005) generate a SQL statement without check for missing value for the `severity` and `qod` attributes (maybe because of NVT timed out), which may result in sql_exec_internal failed.

### Steps to reproduce
1. Run a scan.
2. Check the logs.

### GVM versions
**gvm:** 22.4.3~dev1

### Logfiles

```
md manage:WARNING:2023-06-05 17h58.16 UTC:31409: sql_exec_internal: PQexec failed: ERROR:  syntax error at or near "and"
LINE 1: ....'    and port = 'general/tcp' and severity =     and qod = ...
                                                             ^
 (7)
md manage:WARNING:2023-06-05 17h58.16 UTC:31409: sql_exec_internal: SQL: SELECT EXISTS (SELECT * FROM results   WHERE report = 137 and hash_value = '405b3db5584b5e7eb7e40fc8b0b99732'    and host = '127.0.0.1' and hostname = ''    and type = 'Error Message' and description = 'NVT timed out after 320 seconds.'    and port = 'general/tcp' and severity =     and qod =  and path = '' );
md manage:WARNING:2023-06-05 17h58.16 UTC:31409: sql_x: sql_exec_internal failed
md manage:WARNING:2023-06-05 17h58.16 UTC:31409: manage_cleanup_process_error: Error exit, setting running task to Interrupted
md manage:WARNING:2023-06-05 17h58.16 UTC:31409: sql_exec_internal: PQexec failed: ERROR:  current transaction is aborted, commands ignored until end of transaction block
 (7)
md manage:WARNING:2023-06-05 17h58.16 UTC:31409: sql_exec_internal: SQL: UPDATE reports SET scan_run_status = 13 WHERE id = 137;
md manage:WARNING:2023-06-05 17h58.16 UTC:31409: sqlv: sql_exec_internal failed
```

sed-ats:
I get the same error

when running several tasks in which there are several hosts
in version gvmd 22.4.2 everything worked fine
**GVM versions**
Greenbone Vulnerability Manager 22.5.1
Manager DB revision 254

```
[...]

PR #2018, opened 2023-06-26
Fix: Improve SQL quoting in results and host details

timopollmeier:
## What
When creating new results in make_osp_result and checking for duplicate results and host details, all strings are SQL quoted.

## Why
This fixes various SQL errors like ones when recieving an empty result severity or a host detail containing a single quote / apostrophe.

## References
GEA-223
GEA-226
GEA-228


jhelmold (approved):

cfi-gb:
@timopollmeier @jhelmold @greenbone/gvmd-maintainers I guess this should also solve / fix #2006 ?

PR #2048, opened 2023-07-18
Change: Lock table for each auto-deleted report individually

timopollmeier:
## What
The auto_delete_reports function no longer keeps the access exclusive lock on the reports table for the entire loop over the reports, but instead acquires and releases the lock for each report to delete.

## Why
This gives other processes, like ones deleting reports manually, a better chance to acquire the lock on the reports table.

## References
GEA-2

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -18917,15 +18917,7 @@ auto_delete_reports ()
 
   g_debug ("%s", __func__);
 
-  sql_begin_immediate ();
-
-  /* As in delete_report, this prevents other processes from getting the
-   * report ID. */
-  if (sql_int ("SELECT try_exclusive_lock('reports');") == 0)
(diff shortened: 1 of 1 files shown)


src/manage_sql.c
```diff
@@ -18970,31 +18962,64 @@ auto_delete_reports ()
...
+          continue;
+        }
+
+      g_debug ("%s: deleting report %llu", __func__, report);
+      ret = delete_report_internal (report);
+      if (ret == 2)
+        {
+          /* Report is in use. */
+          g_debug ("%s: %llu is in use", __func__, report);
+          continue;
```
jhelmold, line 19011:
```suggestion
          sql_rollback ();
          continue;
```

jhelmold (approved):

PR #2051, opened 2023-07-27
Update gvmd.logrotate.in

tonoitp:
## What
added "su gvm gvm" in logrotate config file

## Why
logrotate aborted on insecure folder rights.


mattmundell:
Solved by greenbone/gvmd/pull/2269.

PR #2054, opened 2023-08-04
Change: Install old sync scripts, but deprecate them

timopollmeier:
## What
The old sync scripts are installed by default for backward compatibility but also marked as deprecated.

The installation instructions are updated to mention the new sync tool and CMake option to disable to old scripts.

## Why
This makes the behavior and documentation of gvmd more consistent with that in openvas-scanner.

## References
GEA-169
Corresponding change in openvas-scanner: greenbone/openvas-scanner/pull/1393

### DIFF
--- a/.docker/prod.Dockerfile
+++ b/.docker/prod.Dockerfile
@@ -10,7 +10,7 @@ WORKDIR /source
 RUN mkdir /build && \
     mkdir /install && \
     cd /build && \
-    cmake -DCMAKE_BUILD_TYPE=Release /source && \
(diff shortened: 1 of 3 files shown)


bjoernricks (changes requested):
Hey I understand the purpose of this PR but I would be in favor of keeping the already released behavior and NOT switching to the openvas-scanner one. The old script(s) should go away and shouldn't be used anymore right now. Implementing the behavior in openvas-scanner was a compromise with the scanner team. It wasn't my favorite choice.

I like your additions to the docs but please keep `INSTALL_OLD_SYNC_SCRIPTS` at `OFF` by default.

timopollmeier:
I've created a new PR to only deprecate the old scripts and rewrite the installation instructions differently in #2055

PR #2065, opened 2023-08-22
Fix: Remove extra "gvmd: " from process titles

timopollmeier:
## What
The "gvmd: " prefix is removed from the text passed to calls to setproctitle.

## Why
The function will add a prefix of the executable name on its own, so it now no longer adds a superfluous one.

## References
GEA-240


jhelmold (approved):

cfi-gb:
Reference to the change this got introduced: #1812

PR #2069, opened 2023-08-28
increase max email attachment size

alexmateescu:
## What

With the docker container now allowing sendmail if the attachment is bigger than 1 MB the email being sent will notify that the attachment is too big to be attached.

My change uses the GVMD arguments to increase the attachment size
Added the flag and rerun container with the flag. email sent ok.

## Why

in order to be able to send bigger PDF reports

### DIFF
--- a/.docker/start-gvmd.sh
+++ b/.docker/start-gvmd.sh
@@ -20,7 +20,7 @@
 
 [ -z "$USER" ] && USER="admin"
 [ -z "$PASSWORD" ] && PASSWORD="admin"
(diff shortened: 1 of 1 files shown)


bjoernricks:
Hi,

I am not going to merge this PR because you can easily override the defaults by setting the `GVMD_ARGS` environment variable in the compose file.

If you don't want to adjust the greenbone docker compose file you can extend it easily. See https://docs.docker.com/compose/multiple-compose-files/ for more details.

PR #2095, opened 2023-10-12
Change: Use new report generation for delta report

a-h-abdelsalam:
## What
Add new report generation for delta reports. The new report generation is enabled by changing the setting of "Delta Reports Version" to 2.

## Why
The old version of delta report generation was based on the sort order which was not reliable.

## References
GEA-320


bjoernricks:
Hi, could you describe how the new diff report behaves in contrast to the old one?

jhelmold (approved):

PR #2105, opened 2023-11-02
Change: Improve table locking with timeout + retry

timopollmeier:
## What
When trying to acquire the report table lock with retries, do not use NOWAIT and sleep (), but instead use the lock_timeout setting in the current transaction.
The timeout is adjusted to be below the deadlock_timeout and the number of retries is raised accordingly.

## Why
This allows waiting processes to acquire the lock as soon as it becomes available, reducing wait times for transactions blocked by fast other transactions.

## References
GEA-327

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -22001,15 +22001,6 @@ gmp_xml_handle_end_element (/* unused */ GMarkupParseContext* context,
                       }
                     log_event_fail ("task", "Task", NULL, "created");
                     break;
-                  case 3:
-                    SENDF_TO_CLIENT_OR_FAIL(
-                      "<create_task_response"
-                      " status=\"%s\""
-                      " status_text=\"%s\"/>",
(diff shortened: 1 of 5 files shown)


src/gmp.c
```diff
@@ -21928,6 +22001,15 @@ gmp_xml_handle_end_element (/* unused */ GMarkupParseContext* context,
                       }
                     log_event_fail ("task", "Task", NULL, "created");
                     break;
+                  case 3:
```
a-h-abdelsalam, line 22004: Does `copy_task` have a return value of 3 or should the return of `request_delete_task` be checked instead ?

timopollmeier: I've removed the case now because it does not apply to the return value that is actually checked as you said.
I do not want to change the behavior for now but we can discuss this separately later.

a-h-abdelsalam (approved):

PR #2108, opened 2023-11-14
Add dependencies for "Build Documentation" workflow and run on PR

timopollmeier:
## What
Some missing dependencies have been added to the "Build XML documentation and upload coverage" job and the workflow is now run for any pull request on main, stable or oldstable to catch errors earlier.

## Checklist

- [ ] Tests

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -3,6 +3,8 @@ name: 'Build Documentation'
 on:
   push:
     branches: [ main, stable, oldstable ]
+  pull_request:
+    branches: [ main, stable, oldstable ]
 
 jobs:
   generate-doc-and-upload-coverage:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-docs.yml
```diff
@@ -3,13 +3,20 @@ name: 'Build Documentation'
...
   generate-doc-and-upload-coverage:
     name: Build XML documentation and upload coverage
     runs-on: ubuntu-latest
     container: greenbone/doxygen
     steps:
+      - name: Install gvmd doc dependencies
+        run: |
+          apt update
+          apt install --no-install-recommends -y xmltoman xsltproc
+          rm -rf /var/lib/apt/lists/*
```
bjoernricks, line 19: This isn't required. It's just for reducing the size of a image layer when publishing a container image.
```suggestion
```

PR #2109, opened 2023-11-15
Add: Use a 3 arg iso_time for GET iterators

mattmundell:
## What

Use a three argument version of pl/pgsql function `iso_time` to speed up GET iterators.

## Why

Trying to speed up SecInfo pages, when paging to the end.

This PR takes the CPEs from 9s to 7s, for example, so probably need something better.

## References

https://www.postgresql.org/docs/current/queries-limit.html 7.6 LIMIT and OFFSET:
`The rows skipped by an OFFSET clause still have to be computed inside the server; therefore a large OFFSET might be inefficient.`


mattmundell:
Closing because greenbone/gvmd/pull/2112 removes the use of iso_time entirely for these cases.

PR #2111, opened 2023-11-20
Change: move GET iterator time conversion out of SQL

mattmundell:
## What

Move the ISO time calculations in the "GET" iterators from SQL to C.

## Why

Having the iterator do the time conversion is slower when there are many rows with a large offset, because Postgres still does the conversion for every row.

`time o m m '<get_info type="nvt" filter="sort-reverse=created rows=10 first=80000"/>' > /tmp/info`
patch 0.2, 0.2, 0.16
main  1.8, 2.2, 1.7

Leads to faster SecInfo pages when paging to the end, CPEs in particular.

## References

https://www.postgresql.org/docs/current/queries-limit.html 7.6 LIMIT and OFFSET:
`The rows skipped by an OFFSET clause still have to be computed inside the server; therefore a large OFFSET might be inefficient.`

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -8045,7 +8045,7 @@ buffer_notes_xml (GString *buffer, iterator_t *notes, int include_notes_details,
 
           text = note_iterator_text (notes);
           excerpt = utf8_substring (text, 0, setting_excerpt_size_int ());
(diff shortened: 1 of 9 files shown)


mattmundell:
Replaced by greenbone/gvmd/pull/2112

PR #2126, opened 2024-01-15
Docs: Use a simpler structure for the HTML GMP doc

mattmundell:
## What

In the GMP HTML doc, use an HTML `<details>` for every item. As a result remove the contents and the summary sections.

Initially most of the top level items (eg "1 Data Types") are open.  The second level items are all closed (eg "1.1 Data Type alive_test").  This makes the initial view concise while still supporting links that jump to the second level items.

## Why

The doc is big and hard to take in. The idea is to collapse everything so it's easier to see and navigate.

One downside is that when an item is closed it is excluded from text search.

## References

Waits for greenbone/gvmd/pull/2124.


mattmundell:
Going with less intrusive greenbone/gvmd/pull/2128 instead.

PR #2127, opened 2024-01-18
Now the CPE dictionary is no longer continuously parsed again if the dictionary is corrupt.

jhelmold:
## What
Now the CPE dictionary is no longer continuously parsed again if the dictionary is corrupt.

## Why
This is a bug-fix

## References
GEA-419

### DIFF
--- a/src/manage_sql_secinfo.c
+++ b/src/manage_sql_secinfo.c
@@ -3062,7 +3062,7 @@ check_cert_db_version ()
  *
  * @return 0 success, -1 error.
  */
-static int
+static void
 update_cert_timestamp ()
 {
   GError *error;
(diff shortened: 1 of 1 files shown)


src/manage_sql_secinfo.c
```diff
@@ -3093,22 +3093,20 @@ update_cert_timestamp ()
...
+          stamp = time(NULL);
         }
 
       timestamp[8] = '\0';
       g_debug ("%s: parsing: %s", __func__, timestamp);
       stamp = parse_feed_timestamp (timestamp);
       g_free (timestamp);
       if (stamp == 0)
-        return -1;
+        stamp = time(NULL);
```
a-h-abdelsalam, line 3104: Does it make sense to add an else around those lines ? If the above if-condition is true, we already have freed `timestamp`.

src/manage_sql_secinfo.c
```diff
@@ -3364,22 +3362,20 @@ update_scap_timestamp ()
...
```
a-h-abdelsalam, line 3373: Same as above. I think we need an else here.

a-h-abdelsalam (changes requested):

PR #2128, opened 2024-01-22
Add: hiding of sections in HTML GMP doc

mattmundell:
## What

In the HTML GMP doc, use HTML's `DETAILS` to allow the reader to open and close sections. Start with all the RNC sections closed.

Also improve the RNC boxes, and add boxes around the XML code in the examples.

## Why

The doc is very long which makes it hard to navigate. The RNC is likely used less often.

## References

Waits for greenbone/gvmd/pull/2124.


mattmundell:
Moved to smaller PRs.

PR #2130, opened 2024-01-23
Fix: The report now has the full size (no longer zero bytes).

jhelmold:
## What
The Vulnerability Report PDF had in special cases only a length of zero bytes (it was empty) when sent by an alert. This is no longer the case.

## Why
This is a bug-fix.

## References

GEA-408
## Checklist
Tested manually on my development system

- [ ] Tests

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -8439,7 +8439,7 @@ buffer_overrides_xml (GString *buffer, iterator_t *overrides,
             "<in_use>0</in_use>"
             "<active>%i</active>"
             "<end_time>%s</end_time>"
-            "<text excerpt=\"%s\">%s</text>"
+            "<text excerpt=\"0\">%s</text>"
             "<hosts>%s</hosts>"
             "<port>%s</port>"
             "<threat>%s</threat>"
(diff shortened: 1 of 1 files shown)


src/gmp.c
```diff
@@ -8439,7 +8439,7 @@ buffer_overrides_xml (GString *buffer, iterator_t *overrides,
             "<in_use>0</in_use>"
             "<active>%i</active>"
             "<end_time>%s</end_time>"
-            "<text>%s</text>"
+            "<text excerpt=\"%s\">%s</text>"
```
mattmundell, line 8442:
```suggestion
            "<text excerpt=\"0\">%s</text>"
```

src/gmp.c
```diff
@@ -8458,6 +8458,7 @@ buffer_overrides_xml (GString *buffer, iterator_t *overrides,
             get_iterator_modification_time (overrides),
             override_iterator_active (overrides),
             end_time > 1 ? iso_time (&end_time) : "",
+            "0",
```
mattmundell, line 8461:
```suggestion
```

mattmundell (changes requested):
Seems like something the PDF generator should be able to handle, but nice to have the excerpt attribute consistently there.

Would be good if you could do the same for notes.

PR #2132, opened 2024-02-09
Fix: Handle permission clause for result permissions.

jhelmold:
## What
Results of a task could not bee seen any more by a user that has read permission
to the task. This problem is fixed now.

## Why
This is a bug-fix

## References
GEA-477

## Checklist
Tested manually on my development system.

- [ ] Tests

### DIFF
--- a/src/manage_acl.c
+++ b/src/manage_acl.c
@@ -1143,9 +1143,10 @@ acl_where_owned_user (const char *user_id, const char *user_sql,
           else if (strcmp (type, "result") == 0)
             permission_clause
              = g_strdup_printf ("%s"
-                                " OR results%s.task IN"
+                                " OR EXISTS"
                                 " (SELECT id FROM %spermissions_subject"
-                                "  WHERE resource_type = 'task'"
+                                "  WHERE resource = results%s.task"
+                                "  AND resource_type = 'task'"
                                 "  AND (%s))",
                                 clause,
                                 get->trash ? "_trash" : "",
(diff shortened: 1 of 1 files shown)


src/manage_acl.c
```diff
@@ -1143,9 +1143,10 @@ acl_where_owned_user (const char *user_id, const char *user_sql,
...
              = g_strdup_printf ("%s"
-                                " OR results%s.task IN"
+                                " OR EXISTS"
                                 " (SELECT id FROM %spermissions_subject"
-                                "  WHERE resource_type = 'task'"
+                                "  WHERE resource = results%s.task"
+                                "  AND resource_type = 'task'"
                                 "  AND (%s))",
                                 clause,
                                 get->trash ? "_trash" : "",
```
a-h-abdelsalam, line 1152: I think `get->trash ? "_trash" : ""` and `with_prefix ? with_prefix : ""` also need to be switched. Do you agree ?

a-h-abdelsalam (changes requested):

src/manage_acl.c
jhelmold: Yes sorry, you are right. I will change that.

a-h-abdelsalam (approved):

src/manage_acl.c
```diff
@@ -1143,13 +1143,14 @@ acl_where_owned_user (const char *user_id, const char *user_sql,
           else if (strcmp (type, "result") == 0)
             permission_clause
              = g_strdup_printf ("%s"
-                                " OR results%s.task IN"
+                                " OR EXISTS"
                                 " (SELECT id FROM %spermissions_subject"
```
mattmundell, line 1147: Feels like this should be IN to match the other cases. In 28054e8dff5328572e649f2d0562e746acc86451 in PR  #2118 @timopollmeier changed the EXISTS to IN, but he forgot to change `id` to `resource`. Maybe messed up his performance measurements, but maybe he was measuring the other cases.

PR #2137, opened 2024-02-22
Fix: Free quoted strings in SCAP update

timopollmeier:
## What
Some of the quoted strings that were not freed in the CPE and CVE updates now are.

## Why
This a bug fix that will reduce peak memory consumption of SCAP updates significantly.

## References
GEA-490


a-h-abdelsalam (approved):

bjoernricks:
Nice catch 🚀

PR #2140, opened 2024-02-27
Fix start gvmd script

bjoernricks:
## What

Update and fix the container startup script

## Why

It wasn't possible to override the postgres data dir and contained an issue when users similar to the initial user exist (like if the user is `admin` and the users `foo_admin` and `admin_bar` exist).


mattmundell (approved):
Looks fine.

PR #2145, opened 2024-03-05
Fix: Handle value of last_update and SCAP schema when feed update fails.

jhelmold:
## What
The storage of the value of last update and the handling of the scap and scap2 schemata are adjusted in case of a
failing feed update. So the changed logic for last update takes effect.

## Why
This is part of a bug-fix

## References
GEA-419

## Checklist

- [ ] Tests


src/manage_sql_secinfo.c
```diff
@@ -3438,6 +3415,39 @@ update_scap_end ()
...
     }
+}
+
+/**
+ * @brief Finish scap update.
+ *
+ * @return 0 success, -1 error.
+ */
+static int
+update_scap_end ()
```
a-h-abdelsalam, line 3426: I think this function is no longer returning -1. It returns 0 in all cases now. Do you think it make sense to change it to return void ? If so, the callers would need to be updated accordingly to return 0 on success.

```suggestion
 */
static void
update_scap_end ()
```

src/manage_sql_secinfo.c
```diff
@@ -3451,6 +3461,46 @@ update_scap_end ()
   return 0;
```
a-h-abdelsalam, line 3461: If the return type is updated.
```suggestion
```

src/manage_sql_secinfo.c
```diff
@@ -3451,6 +3461,46 @@ update_scap_end ()
...
+      /* View 'vulns' contains references into the SCAP schema, so it is
+       * removed by the CASCADE. */
+      create_view_vulns ();
+    }
+  else
+    {
+      /* reset scap2 schema */
+      manage_db_init ("scap");
+      manage_db_init_indexes ("scap");
+      manage_db_add_constraints ("scap");
```
a-h-abdelsalam, line 3490: Is it a good idea to error check these calls ? In case, one of them fails, we may not proceed with the next ones. 
What do you think ?

a-h-abdelsalam (approved):

PR #2147, opened 2024-03-06
Change: Use XML file iterator for CVE updates

timopollmeier:
## What
The update of the CVEs now also uses the XML file iterator instead of loading and parsing the whole file.
Also, a memory leak in insert_cve_products has been fixed.

## Why
This reduces the amount of memory used during the update.

## References
GEA-498

### DIFF
--- a/src/manage_sql_secinfo.c
+++ b/src/manage_sql_secinfo.c
@@ -2370,8 +2370,7 @@ insert_cve_products (element_t list, resource_t cve,
           gchar *quoted_product, *product_decoded;
           gchar *product_tilde;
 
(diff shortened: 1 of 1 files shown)


timopollmeier:
@bjoernricks This uses the same version of gvm-libs as #2142 and I've raised the version requirement there.

bjoernricks:
> @bjoernricks This uses the same version of gvm-libs as #2142 and I've raised the version requirement there.

nice! just wanted to be sure :-)

a-h-abdelsalam (approved):

PR #2165, opened 2024-04-10
Add: New multi_selection report format param type

timopollmeier:
## What
A new parameter type for report formats and report configs is added that allows selecting multiple predefined options.

## Why
This can later be used to add new customization options to report formats like a column selection for CSV.

## References
GEA-403


src/manage_sql_report_formats.c
```diff
@@ -2406,6 +2407,28 @@ report_format_param_type_min (report_format_t report_format, const char *name)
   return min;
 }
 
+static int
+report_format_param_value_in_options (report_format_param_t param,
+                                      const char *value)
```
a-h-abdelsalam, line 2412: You may want to add a description for this function.

src/manage_sql_report_formats.c
```diff
@@ -2533,6 +2545,86 @@ report_format_validate_param_value (report_format_t report_format,
...
+              if (!cJSON_IsString (array_item))
+                {
+                  if (error_message)
+                    {
+                      *error_message
+                        = g_strdup_printf ("value of param \"%s\""
+                                           " contains a non-string value",
+                                           name);
+                    }
+                  cJSON_Delete (json);
```
a-h-abdelsalam, line 2582: I think I would be cautious about deleting the array while iterating over it. I would prefer to either set a flag and break or jump out of the loop and delete the array afterwards.

src/manage_sql_report_formats.c
```diff
@@ -2533,6 +2545,86 @@ report_format_validate_param_value (report_format_t report_format,
...
+              if (! report_format_param_value_in_options (param, string))
+                {
+                  if (error_message)
+                    {
+                      *error_message
+                        = g_strdup_printf ("\"%s\" in value of param \"%s\""
+                                           " is not a valid selection option",
+                                           string, name);
+                    }
+                  cJSON_Delete (json);
```
a-h-abdelsalam, line 2595: Same for this one.

a-h-abdelsalam (changes requested):

src/manage_sql_report_formats.c
timopollmeier: The function returns immediately after this so this should be fine.

src/manage_sql_report_formats.c
timopollmeier: The function also returns here.

src/manage_sql_report_formats.c
timopollmeier: I've added one in ae41309

a-h-abdelsalam (approved):

PR #2184, opened 2024-04-29
Change: Raised log level for output of backtrace for signal handler.

jhelmold:
## What
We raised the log level for the output of the back trace in the signal handler so that
the back trace is always printed, when a signal occurs.

## Why
When a signal occurs the back trace is sometimes not printed. We assume that the
log level is not evaluated correctly in that case. We now set the log level of the back trace
messages fixed to the maximum, so that the back trace is hopefully always printed.

## References
GEA-526

## Checklist

- [ ] Tests

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -17133,12 +17133,12 @@ cleanup_manage_process (gboolean cleanup)
 void
 manage_cleanup_process_error (int signal)
 {
-  g_debug ("Received %s signal", strsignal (signal));
+  g_message ("Received %s signal", strsignal (signal));
   if (sql_is_open ())
     {
       if (current_scanner_task)
         {
-          g_warning ("%s: Error exit, setting running task to Interrupted",
+          g_message ("%s: Error exit, setting running task to Interrupted",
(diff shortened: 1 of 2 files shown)


src/manage_sql.c
```diff
@@ -17133,12 +17133,12 @@ cleanup_manage_process (gboolean cleanup)
...
 manage_cleanup_process_error (int signal)
 {
-  g_debug ("Received %s signal", strsignal (signal));
+  g_message ("Received %s signal", strsignal (signal));
   if (sql_is_open ())
     {
       if (current_scanner_task)
         {
-          g_warning ("%s: Error exit, setting running task to Interrupted",
+          g_message ("%s: Error exit, setting running task to Interrupted",
```
timopollmeier, line 17141: I think we should only change the debug messages, not the warning.
```suggestion
          g_warning ("%s: Error exit, setting running task to Interrupted",
```

timopollmeier (changes requested):

timopollmeier (approved):

PR #2187, opened 2024-05-01
Update setup-mta.sh to include tls_certcheck

Nicolaj-H:
## What

Added a simple option to ignore certificate errors.

## Why

could be if your running self-signed, or if the root certificate isn't accepted, maybe you just need a fast way to enable sending reports while waiting for your mail provider to update their expired certificate.


bjoernricks:
Hi and thanks for your contribution! Could you also update the docs at https://github.com/greenbone/docs/blob/main/src/22.4/container/workflows.md?plain=1#L336 for this change?

Nicolaj-H:
> Hi and thanks for your contribution! Could you also update the docs at https://github.com/greenbone/docs/blob/main/src/22.4/container/workflows.md?plain=1#L336 for this change?

i have made a PR at the docs, documenting the option.
https://github.com/greenbone/docs/pull/467

PR #2219, opened 2024-06-11
Add: Container push into greenbone registry

pascalholthaus:
## What
Add: Container push into greenbone registry

## Why
Get the currently at dockerhub uploaded images into the greenbone registry.

## References
DEVOPS-1060

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,4 +1,4 @@
-name: Build and Push
+name: Build and Push to Greenbone Registry
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/push.yml
```diff
@@ -0,0 +1,55 @@
+name: Build and Push
```
y0urself, line 1:
```suggestion
name: Build and Push to Harbor
```

pascalholthaus: Should we really use the registry application name?

y0urself: I don't care. Ignore my suggestion.

y0urself (approved):

PR #2248, opened 2024-07-03
Fix: Escape control chars in UTF-8 TLS cert DNs

timopollmeier:
## What
ASCII control characters (0x01 - 0x1F and 0x7F) are now escaped in the DNs of TLS certficates even they are valid UTF-8.
Characters in the range 0x80 - 0xFF are escaped if the DN is not valid UTF-8.
This character escaping for TLS certificates is applied more consistently.

## Why
This addresses issues with XML parsers, e.g. ones used by report formats, not being able to process control characters.

## References
GEA-607

## Checklist

- [x] Tests

### DIFF
--- a/src/sql.c
+++ b/src/sql.c
@@ -161,7 +161,7 @@ sql_quote (const char* string)
  * @return Freshly allocated, quoted string. Free with g_free.
  */
 gchar*
-sql_ascii_escape_and_quote (const char* string)
+sql_ascii_escape_and_quote (const char* string, const char* exceptions)
 {
   gchar *escaped_string;
   gchar *quoted_string;
@@ -171,9 +171,9 @@ sql_ascii_escape_and_quote (const char* string)
   if (string == NULL)
     return NULL;
 
-  escaped_string = strescape_check_utf8 (string, NULL);
+  escaped_string = strescape_check_utf8 (string, exceptions);
   quoted_string = sql_quote (escaped_string);
-  g_free (quoted_string);
+  g_free (escaped_string);
 
   return quoted_string;
 }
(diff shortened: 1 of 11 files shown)


src/sql.c
```diff
@@ -150,38 +150,30 @@ sql_quote (const char* string)
...
- * @brief Quotes a string for use in SQL statements, also ASCII escaping it
- *         if it is not valid UTF-8.
+ * @brief Quotes a string for use in SQL statements, also ASCII escaping it.
  *
- * @param[in]  string  String to quote, has to be \\0 terminated.
+ * The ASCII escaping excludes characters 0x80 - 0xFF for valid UTF-8 strings
+ * and includes them otherwise.
+ *
+ * @param[in]  string       String to quote, has to be \\0 terminated.
+ * @param[in]  exceptions   Optional exceptions for the escaping.
```
mattmundell, line 159: Param is missing.

src/sql.c
```diff
@@ -150,38 +150,30 @@ sql_quote (const char* string)
...
-      gchar *escaped_string;
-      escaped_string = g_strescape (string, "");
-      quoted_string = sql_quote (escaped_string);
-      g_free (escaped_string);
-    }
+    return NULL;
+
+  escaped_string = strescape_check_utf8 (string, NULL);
+  quoted_string = sql_quote (escaped_string);
+  g_free (quoted_string);
```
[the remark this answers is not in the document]
timopollmeier: Yes, I've fixed this now.

src/sql.c
timopollmeier: Okay, I've added the missing param now.

a-h-abdelsalam (approved):

PR #2263, opened 2024-07-18
Logrotate fix

syspect-sec:
## What

This will add proper permissions to the logrotate configuration file, which is placed into the `/etc/logrotate.d/gvmd` during the make build process. 

## Why

It was pointed out from [this forum post]((https://forum.greenbone.net/t/logrotate-debian/18706/7)) that the logs do not have the correct permission configuration for logrotate.

## References

[Forum post with discussion](https://forum.greenbone.net/t/logrotate-debian/18706/7)

## Checklist

- [ *] Tests - Tested and works on Ubuntu 22.04


cfi-gb:
Seems there is a similar PR already open since quite some time: #2051

mattmundell:
Moved to a single commit in greenbone/gvmd/pull/2269.

PR #2266, opened 2024-07-22
Add: New --min-mem-feed-update option

timopollmeier:
## What
A new option is added which will make the automatic feed update wait until a minimum amount of physical memory is available.

## Why
Additionally, the --mem-wait-retries can be used to set the number of retries waiting for memory to be available in each process.

## References
GEA-179

## Checklist

- [ ] Tests

### DIFF
--- a/doc/gvmd.8
+++ b/doc/gvmd.8
@@ -121,8 +121,14 @@ Maximum size of user-defined message text in alert emails, in bytes.
 \fB--max-ips-per-target=\fINUMBER\fB\f1
 Maximum number of IPs per target.
 .TP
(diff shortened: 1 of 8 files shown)


mattmundell (approved):
Looks good.

Maybe overkill passing `check_min_mem_feed_update` to `wait_for_mem` because it's always the same. I guess you're thinking of the future.

Also the `get*_mem_*` functions could be removed.

timopollmeier:
@mattmundell Yes, I added a parameter for the memory check function in case we add more later and I left the getter functions for consistency with other startup options.

PR #2271, opened 2024-07-25
Add: New option --max-concurrent-scan-updates

timopollmeier:
## What
The new startup option --max-concurrent-scan-updates is added which allows limiting the number of scan updates which can run at the same time.

## Why
This can be used to limit the peak memory and CPU usage when running multiple scans at the same time.

## References
GEA-179

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -48,6 +48,7 @@
 
 #include "debug_utils.h"
 #include "gmp_base.h"
+#include "ipc.h"
 #include "manage.h"
 #include "manage_acl.h"
 #include "manage_configs.h"
@@ -75,7 +76,6 @@
 #include <string.h>
 #include <strings.h>
 #include <sys/file.h>
-#include <sys/sem.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <sys/wait.h>
@@ -220,17 +220,6 @@ static int schedule_timeout = SCHEDULE_TIMEOUT_DEFAULT;
  *        lost in a running task.
  */
 static int scanner_connection_retry = SCANNER_CONNECTION_RETRY_DEFAULT;
-
-/**
- * @brief System V semaphore set key for gvmd actions.
- */
-static key_t semaphore_set_key = -1;
-
-/**
- * @brief System V semaphore set id for gvmd actions.
- */
-static int semaphore_set = -1;
-
 
 
 /* Certificate and key management. */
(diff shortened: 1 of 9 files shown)


src/manage.c
```diff
@@ -7758,3 +7908,115 @@ delete_resource (const char *type, const char *resource_id, int ultimate)
...
+      g_free (key_file_name);
+      return -1;
+    }
+
+  semaphore_set = semget (semaphore_set_key, 1, 0660 | IPC_CREAT);
+  if (semaphore_set < 0)
+    {
+      g_warning ("%s: error getting semaphore set: %s",
+                 __func__, strerror (errno));
+      return -1;
```
a-h-abdelsalam, line 7960:
```suggestion
      g_free (key_file_name);
      return -1;
```

src/manage.c
```diff
@@ -7758,3 +7908,115 @@ delete_resource (const char *type, const char *resource_id, int ultimate)
...
+ *
+ * A negative op_value will try to decrease the semaphore value
+ *  and wait if needed.
+ * A positive op_value will increase the semaphore value.
+ * Zero as op_value will wait for the semaphore value to become zero.
+ *
+ * (See semop from sys/sem.h)
+ *
+ * @param[in]  semaphore_index  The index of the semaphore in the gvmd set.
+ * @param[in]  op_value   The operation value
```
a-h-abdelsalam, line 7989: `timeout` parameter and return value could be added to the description.

src/manage.c
```diff
@@ -7758,3 +7908,115 @@ delete_resource (const char *type, const char *resource_id, int ultimate)
...
+    unsigned short  *array;  /* Array for GETALL, SETALL */
+    struct seminfo  *__buf;  /* Buffer for IPC_INFO
+                                (Linux-specific) */
+};
+
+/**
+ * @brief Initializes the semaphore set for gvmd actions.
+ *
+ * Needs max_concurrent_scan_updates to be set.
+ */
```
a-h-abdelsalam, line 7930: return value could be added to the description.

src/manage.c
```diff
@@ -7758,3 +7908,115 @@ delete_resource (const char *type, const char *resource_id, int ultimate)
   assert (0);
   return -1;
 }
+
+
+/* Inter-Process Communication */
```
mattmundell, line 7913: Could we move this section to a new file? It looks self contained and manage.c is already large.

mattmundell (changes requested):

src/manage.c
```diff
@@ -7758,3 +7908,115 @@ delete_resource (const char *type, const char *resource_id, int ultimate)
...
+
+/**
+ * @brief Union type for values of semctl actions
+ */
+union semun {
+    int              val;    /* Value for SETVAL */
+    struct semid_ds *buf;    /* Buffer for IPC_STAT, IPC_SET */
+    unsigned short  *array;  /* Array for GETALL, SETALL */
+    struct seminfo  *__buf;  /* Buffer for IPC_INFO
+                                (Linux-specific) */
```
mattmundell, line 7923:
```suggestion
  int val;                ///< Value for SETVAL
  struct semid_ds *buf;   ///< Buffer for IPC_STAT, IPC_SET
  unsigned short *array;  ///< Array for GETALL, SETALL
  struct seminfo *__buf;  ///< Buffer for IPC_INFO (Linux-specific)
```

src/manage.c
```diff
@@ -7916,17 +7916,18 @@ delete_resource (const char *type, const char *resource_id, int ultimate)
  * @brief Union type for values of semctl actions
  */
 union semun {
-    int              val;    /* Value for SETVAL */
-    struct semid_ds *buf;    /* Buffer for IPC_STAT, IPC_SET */
-    unsigned short  *array;  /* Array for GETALL, SETALL */
-    struct seminfo  *__buf;  /* Buffer for IPC_INFO
-                                (Linux-specific) */
+    int              val;    ///< Value for SETVAL
```
mattmundell: This is still indenting by 4 whereas all our `typedef struct`s and functions etc indent by 2.

I can understand breaking the convention to line up the field names though, if you like those.

mattmundell (dismissed):

PR #2272, opened 2024-07-26
Change: Retry on deadlocks in sql_x

timopollmeier:
## What
The sql_x function (and thus functions using it like sql_string) now also retries executing the statement in case of a deadlock like the sql function.

## Why
This addresses an issue with queries of the auth_cache aborting if there are too many concurrent ones.

## References
GEA-655

### DIFF
--- a/src/sql.c
+++ b/src/sql.c
@@ -208,7 +208,7 @@ sql_insert (const char *string)
  *
  * @return 0 success, 1 gave up (even when retry given),
  *         2 reserved (lock unavailable), 3 unique constraint violation,
- *         -1 error.
+ *         4 deadlock, -1 error.
  */
 int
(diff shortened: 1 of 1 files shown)


src/sql.c
```diff
@@ -400,6 +402,16 @@ sql_x (char* sql, va_list args, sql_stmt_t** stmt_return)
           sql_finalize (*stmt_return);
           continue;
         }
+      else if (ret == -5)
```
mattmundell, line 405:
```suggestion
      if (ret == -5)
```
I know it's minor but the `else` bugs me because it makes it look like an `else` is necessary. Breaks the pattern with the `if` statements above.

mattmundell (approved):

a-h-abdelsalam (approved):

PR #2277, opened 2024-08-07
Add: Added the loading of CPE match rules from JSON files.

jhelmold:
## What
Added the functionality to load the CPE match rules for CVEs from JSON files and to store them in the database.

## Why
This functionality is needed for the matching of CPEs to rules as used in CVEs.

## References
GEA-624

## Checklist

- [ ] Tests

### DIFF
--- a/src/manage_pg.c
+++ b/src/manage_pg.c
@@ -3479,6 +3479,22 @@ manage_db_init (const gchar *name)
            "  cve_refs INTEGER DEFAULT 0,"
            "  nvd_id text);");
 
+      sql ("CREATE TABLE scap2.cpe_match_nodes"
+           " (id SERIAL PRIMARY KEY,"
+           "  parent_id INTEGER DEFAULT 0,"
+           "  cve_id INTEGER DEFAULT 0,"
+           "  operator text);");
+
+      sql ("CREATE TABLE scap2.cpe_match_range"
+           " (id SERIAL PRIMARY KEY,"
+           "  node_id INTEGER DEFAULT 0,"
+           "  vulnerable INTEGER DEFAULT 0,"
+           "  cpe text,"
+           "  version_start_incl text,"
+           "  version_start_excl text,"
+           "  version_end_incl text,"
+           "  version_end_excl text);");
+
       sql ("CREATE TABLE scap2.cpe_details"
            " (id SERIAL PRIMARY KEY,"
            "  cpe_id text,"
--- a/src/manage_sql_secinfo.c
+++ b/src/manage_sql_secinfo.c
@@ -2709,8 +2709,8 @@ json_object_item_double (cJSON *object, char *key, double fallback)
  *
  * @return The (database) id of the node.
  */
-static int
-save_node (long int parent_id, long int cve_id, char * operator)
+static resource_t
+save_node (resource_t parent_id, resource_t cve_id, char *operator)
 {
   return sql_int64_0
            ("INSERT INTO scap2.cpe_match_nodes"
@@ -2803,7 +2803,7 @@ add_cpe_match_rules (long int id, cJSON *match_rules)
  *                        specific tree level.
  */
 static void
-load_nodes (long int parent_id, long int cveid, cJSON *nodes)
+load_nodes (resource_t parent_id, resource_t cveid, cJSON *nodes)
 {
   cJSON *node;
   long int id;
--- a/src/manage_sql_secinfo.c
+++ b/src/manage_sql_secinfo.c
@@ -2712,19 +2712,15 @@ json_object_item_double (cJSON *object, char *key, double fallback)
 static int
 save_node (long int parent_id, long int cve_id, char * operator)
 {
-  long int id;
-
-  id = sql_int64_0
-         ("INSERT INTO scap2.cpe_match_nodes"
-          " (parent_id, cve_id, operator)"
-          " VALUES"
-          " (%i, %i, '%s')"
-          " RETURNING scap2.cpe_match_nodes.id;",
-          parent_id,
-          cve_id,
-          operator);
-
-  return id;
+  return sql_int64_0
+           ("INSERT INTO scap2.cpe_match_nodes"
+            " (parent_id, cve_id, operator)"
+            " VALUES"
+            " (%i, %i, '%s')"
+            " RETURNING scap2.cpe_match_nodes.id;",
+            parent_id,
+            cve_id,
+            operator);
 }
 
 /**
@@ -2781,8 +2777,8 @@ add_cpe_match_rules (long int id, cJSON *match_rules)
       sql
        ("INSERT INTO scap2.cpe_match_range"
         " (node_id, vulnerable, cpe,"
-	"  version_start_incl, version_start_excl,"
-	"  version_end_incl, version_end_excl)"
+        "  version_start_incl, version_start_excl,"
+        "  version_end_incl, version_end_excl)"
         " VALUES"
         " (%ld, %d, '%s', '%s', '%s', '%s', '%s')",
         id,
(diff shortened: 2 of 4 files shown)


jhelmold:
For this PR to work, Timos additions for the json "pull" parser have to be merged first.

src/manage_sql_secinfo.c
```diff
@@ -2656,6 +2657,485 @@ insert_cve_from_entry (element_t entry, element_t last_modified,
...
+         ("INSERT INTO scap2.cpe_match_nodes"
+          " (parent_id, cve_id, operator)"
+          " VALUES"
+          " (%i, %i, '%s')"
+          " RETURNING scap2.cpe_match_nodes.id;",
+          parent_id,
+          cve_id,
+          operator);
+
+  return id;
```
a-h-abdelsalam, line 2727: I think `sql_int64_0` returns `long long int` so may be using `resource_t` makes sense as the return and parameters type for `parent_id` and `cve_id` ?

I also suggest skipping the local variable
```suggestion
  return sql_int64_0
         ("INSERT INTO scap2.cpe_match_nodes"
          " (parent_id, cve_id, operator)"
          " VALUES"
          " (%i, %i, '%s')"
          " RETURNING scap2.cpe_match_nodes.id;",
          parent_id,
          cve_id,
          operator);
```

src/manage_sql_secinfo.c
```diff
@@ -2793,18 +3273,31 @@ update_scap_cves ()
...
+          count++;
+        }
+      if (fnmatch ("nvdcve-2.0-*.xml", cve_path, 0) == 0)
+        {
+          if (update_cve_xml (cve_path, hashed_cpes))
+            {
+              g_dir_close (dir);
+              g_hash_table_destroy (hashed_cpes);
+              cleanup_iterator (&cpes);
+              return -1;
```
a-h-abdelsalam, line 3296: Maybe this needs to be an if-else to check for JSON first and if not fallback to XML ?

src/manage_sql_secinfo.c
```diff
@@ -2656,6 +2657,485 @@ insert_cve_from_entry (element_t entry, element_t last_modified,
...
+  cJSON_ArrayForEach (description_item_json, description_data_json)
+    {
+      char *lang = json_object_item_string (description_item_json, "lang");
+      if (lang != NULL && strcmp(lang, "en") == 0)
+        description_value = json_object_item_string (description_item_json, "value");
+    }
+
+  char *quoted_description = sql_quote (description_value);
+
+  cve_db_id = sql_int64_0
```
a-h-abdelsalam, line 2965: I think this should be long long or better resource_t

src/manage_sql_secinfo.c
```diff
@@ -2656,6 +2657,485 @@ insert_cve_from_entry (element_t entry, element_t last_modified,
...
+            }
+          if (handle_json_cve_item (entry))
+            {
+              gvm_json_pull_event_cleanup (&event);
+              gvm_json_pull_parser_cleanup (&parser);
+              fclose (cve_file);
+              return -1;
+            }
+          increment_transaction_size (&transaction_size);
+          gvm_json_pull_parser_next (&parser, &event);
```
a-h-abdelsalam, line 3113: I think you may want to free entry at the end of each iteration and in case of an error.
```suggestion
          gvm_json_pull_parser_next (&parser, &event);
          cJSON_Delete (entry);
```

src/manage_sql_secinfo.c
```diff
@@ -2656,6 +2657,485 @@ insert_cve_from_entry (element_t entry, element_t last_modified,
...
+
+  if (event.type == GVM_JSON_PULL_EVENT_OBJECT_START)
+    {
+      gboolean cve_items_found = FALSE;
+      while (!cve_items_found)
+        {
+          gvm_json_pull_parser_next (&parser, &event);
+	  gvm_json_path_elem_t *path_tail = g_queue_peek_tail (event.path);
+	  if (event.type == GVM_JSON_PULL_EVENT_ARRAY_START &&
+              strcmp (path_tail->key, "CVE_Items") == 0)
```
a-h-abdelsalam, line 3072: Does it make sense to check `path_tail` for NULL here before accessing the key ?

src/manage_sql_secinfo.c
```diff
@@ -2656,6 +2657,485 @@ insert_cve_from_entry (element_t entry, element_t last_modified,
...
+              return -1;
+            }
+          if (handle_json_cve_item (entry))
+            {
+              gvm_json_pull_event_cleanup (&event);
+              gvm_json_pull_parser_cleanup (&parser);
+              fclose (cve_file);
+              return -1;
+            }
+          increment_transaction_size (&transaction_size);
```
a-h-abdelsalam, line 3112: I think `sql_begin_immediate ()` and a `sql_commit () `are also needed ?

src/manage_sql_secinfo.c
```diff
@@ -2656,6 +2657,485 @@ insert_cve_from_entry (element_t entry, element_t last_modified,
...
+      while (event.type == GVM_JSON_PULL_EVENT_OBJECT_START)
+        {
+          entry = gvm_json_pull_expand_container (&parser, &error_message);
+          if (error_message)
+            {
+              g_warning ("%s: Error expanding CVE item: %s", __func__, error_message);
+              gvm_json_pull_event_cleanup (&event);
+              gvm_json_pull_parser_cleanup (&parser);
+              fclose (cve_file);
+              return -1;
```
a-h-abdelsalam, line 3103:
```suggestion
              cJSON_Delete (entry);
              return -1;
```

src/manage_sql_secinfo.c
```diff
@@ -2656,6 +2657,485 @@ insert_cve_from_entry (element_t entry, element_t last_modified,
...
+            }
+          if (handle_json_cve_item (entry))
+            {
+              gvm_json_pull_event_cleanup (&event);
+              gvm_json_pull_parser_cleanup (&parser);
+              fclose (cve_file);
+              return -1;
```
a-h-abdelsalam, line 3110:
```suggestion
              cJSON_Delete (entry);
              return -1;
```

src/manage_sql_secinfo.c
```diff
@@ -2656,6 +2657,485 @@ insert_cve_from_entry (element_t entry, element_t last_modified,
...
+  cJSON_ArrayForEach(node, nodes)
+    {
+      operator = cJSON_GetObjectItemCaseSensitive(node, "operator");
+      id = save_node (parent_id, cveid, operator->valuestring);
+      cpe_match_rules = cJSON_GetObjectItemCaseSensitive(node, "cpe_match");
+      add_cpe_match_rules (id, cpe_match_rules);
+      child_nodes = cJSON_GetObjectItemCaseSensitive(node, "children");
+      load_nodes (id, cveid, child_nodes);
+    }
+}
```
a-h-abdelsalam, line 2833: I know the macro handles that but I would suggest checking `nodes` for NULL and returning if so.

Does the return of `cJSON_GetObjectItemCaseSensitive` need to be checked for NULL ?

src/manage_sql_secinfo.c
a-h-abdelsalam: I think the function parameters (`parent_id` and `cve_id`) and return type could also be `resource_t` . Same for `load_nodes` since they are all connected.

src/manage_sql_secinfo.c
a-h-abdelsalam: I think if the directory has both JSON and XML files, then it will match them both. First will match the JSON then the XML. Right ? Do we need to handle this case ?

src/manage_pg.c
```diff
@@ -3530,6 +3530,22 @@ manage_db_init (const gchar *name)
...
+      sql ("CREATE TABLE scap2.cpe_match_range"
+           " (id SERIAL PRIMARY KEY,"
+           "  node_id INTEGER DEFAULT 0,"
+           "  vulnerable INTEGER DEFAULT 0,"
+           "  cpe text,"
+           "  version_start_incl text,"
+           "  version_start_excl text,"
+           "  version_end_incl text,"
+           "  version_end_excl text);");
+
```
a-h-abdelsalam, line 3548: I am not sure. Is updating GVMD_SCAP_DATABASE_VERSION required ?

src/manage_sql_secinfo.c
jhelmold: I changed the last if to an else if, so it should be okay now, shouldn't it?

src/manage_pg.c
jhelmold: I am also not sure. Perhaps we can omit it for the time being and possibly catch up on it when Timo is back.

a-h-abdelsalam (approved):

PR #2283, opened 2024-08-21
Add: Add settings for date and time format.

a-h-abdelsalam:
## What

Added independent date and time format setting.

## Why
Currently the date and time formats are set with the language setting. This allows setting the time format to 12h/24h and the date format to wdmy/wmdy independently.

## References
GEA-430

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -16041,6 +16041,28 @@ check_db_settings ()
          "  || ' in SecInfo updates before the end of the file'"
          "  || ' being processed.',"
          "  '100' );");
+
+  if (sql_int ("SELECT count(*) FROM settings"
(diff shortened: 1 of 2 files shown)


mattmundell:
This looks OK. I take it this is for the use of clients like GSA, and all times in GMP will stay as they are?

Asking because there's already a bit of a mess in GMP timezones, eg GET_RESULTS can return multiple timezones in the same response:
```xml
    <modification_time>2023-03-27T12:39:24Z</modification_time>
      <severities score="0.0">
          <date>2021-02-25T13:11:24+02:00</date>
```

a-h-abdelsalam:
@mattmundell yes this is only intended for use in GSA

mattmundell (approved):

PR #2299, opened 2024-09-30
Change: Adjust loading of CPEs to new JSON API

timopollmeier:
## What
CPEs can now be loaded from JSON files based on the NVD API.

As some fields in the old XML differ the JSON API, they are replaced by similar fields:
"nvd_id" is replaced by "cpe_name_id" and "status" is replaced by "deprecated".

The "raw_data" will no longer be available after switching to JSON, so the references are and "deprecated_by" element are handled explictly.

## Why
These changes are made because the XML-based data feeds have been deprecated by NVD.

## References
GEA-623


src/manage_sql_secinfo.c
```diff
@@ -2081,14 +2242,435 @@ insert_scap_cpe_details (inserts_t *inserts, element_t cpe_item)
...
+      first = inserts_check_size (inserts);
+
+      g_string_append_printf (inserts->statement,
+                              "%s ('%s', '%s', '%s')",
+                              first ? "" : ",",
+                              quoted_name,
+                              quoted_ref,
+                              quoted_type);
+
+      inserts->current_chunk_size++;
```
a-h-abdelsalam, line 2518: You may want to free `quoted_ref` and `quoted_type`
```suggestion
      inserts->current_chunk_size++;
      g_free (quoted_ref);
      g_free (quoted_type);      
```

src/manage_sql_secinfo.c
```diff
@@ -2081,14 +2242,435 @@ insert_scap_cpe_details (inserts_t *inserts, element_t cpe_item)
...
+                "  modification_time, deprecated,"
+                "  cpe_name_id)"
+                " VALUES",
+                " ON CONFLICT (uuid) DO UPDATE"
+                " SET name = EXCLUDED.name,"
+                "     title = EXCLUDED.title,"
+                "     creation_time = EXCLUDED.creation_time,"
+                "     modification_time = EXCLUDED.modification_time,"
+                "     deprecated = EXCLUDED.deprecated,"
+                "     deprecated_by_id = EXCLUDED.deprecated_by_id,"
```
a-h-abdelsalam, line 2571: I think this one can be removed now.
```suggestion
                "     deprecated_by_id = EXCLUDED.deprecated_by_id,"
```

a-h-abdelsalam (changes requested):

a-h-abdelsalam (approved):

PR #2303, opened 2024-10-08
Change: Extend get_feeds GMP command.

a-h-abdelsalam:
## What

Added information on whether the feed owner and feed import roles are set and whether the user has access to feed resources.

## Why
To provide a more descriptive feedback to the user in GSA if the scanning is not possible. 

## References
GEA-704

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -12996,15 +12996,11 @@ handle_get_feeds (gmp_parser_t *gmp_parser, GError **error)
 
   if (feed_owner_uuid != NULL && strlen (feed_owner_uuid) > 0)
     feed_owner_set = TRUE;
-  else
-    g_warning ("%s: No feed owner set.", __func__);
 
   setting_value (SETTING_UUID_FEED_IMPORT_ROLES, &feed_roles);
 
   if (feed_roles != NULL && strlen (feed_roles) > 0)
     feed_import_roles_set = TRUE;
-  else
-    g_warning ("%s: No feed import roles set.", __func__);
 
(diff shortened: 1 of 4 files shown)


src/gmp.c
```diff
@@ -12986,10 +12987,59 @@ handle_get_feeds (gmp_parser_t *gmp_parser, GError **error)
...
+  gboolean feed_owner_set, feed_import_roles_set, feed_resources_access;
+
+  feed_owner_set = feed_import_roles_set = feed_resources_access = FALSE;
+
+  setting_value (SETTING_UUID_FEED_IMPORT_OWNER, &feed_owner_uuid);
+
+  if (feed_owner_uuid != NULL && strlen (feed_owner_uuid) > 0)
+    feed_owner_set = TRUE;
+  else
+    g_warning ("%s: No feed owner set.", __func__);
```
timopollmeier, line 13000: I think we should not log a warning here and just remove the message.

src/gmp.c
```diff
@@ -12986,10 +12987,59 @@ handle_get_feeds (gmp_parser_t *gmp_parser, GError **error)
...
+
+  setting_value (SETTING_UUID_FEED_IMPORT_ROLES, &feed_roles);
+
+  if (feed_roles != NULL && strlen (feed_roles) > 0)
+    feed_import_roles_set = TRUE;
+  else
+    g_warning ("%s: No feed import roles set.", __func__);
```
timopollmeier, line 13007: I would also remove this.

timopollmeier (changes requested):

timopollmeier (approved):

PR #2313, opened 2024-10-21
Change: Log to stderr, new log file build option.

jhelmold:
## What
Now the log messages are logged to stderr by default. The log file for the generated gvmd_log.conf can now be set by a new build option GVMD_LOG_FILE.

## Why
Logging to stderr makes the messages also accessible in docker containers.

## References
GEA-420

## Checklist

- [ ] Tests


mattmundell (changes requested):
same requests as gvm-libs

jhelmold:
> same requests as gvm-libs

Did you mean: "same request as gsad"?

PR #2314, opened 2024-10-24
Added reading of gzip files for CPEs.

jhelmold:
## What
Added the functionality to also read CPE data from gzip files.

## Why
We want to be able to use a new feed that uses less disc space.

## References
GEA-635

Requires https://github.com/greenbone/gvm-libs/pull/847.

## Checklist

- [ ] Tests

### DIFF
--- a/src/manage_sql_secinfo.c
+++ b/src/manage_sql_secinfo.c
@@ -2522,21 +2522,35 @@ update_scap_cpes_from_json_file (const gchar *path)
   inserts_t inserts, deprecated_by_inserts;
   gvm_json_pull_parser_t parser;
   gvm_json_pull_event_t event;
-  FILE *json_stream = fopen (path, "r");
-  if (json_stream == NULL)
+  FILE *cpe_file;
+
+  int fd = open (path, O_RDONLY);
+
+  if (fd < 0)
     {
-      g_warning ("%s: Could not open file '%s': %s",
+      g_warning ("%s: Failed to open CPE file '%s': %s",
                  __func__, path, strerror(errno));
       return -1;
     }
 
-  gvm_json_pull_parser_init (&parser, json_stream);
+  g_info ("Updating %s", path);
+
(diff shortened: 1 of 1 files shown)


src/manage_sql_secinfo.c
```diff
@@ -2854,8 +2886,15 @@ update_scap_cpes ()
...
 
   full_path = g_build_filename (GVM_SCAP_DATA_DIR,
-                                "nvd-cpes.json",
+                                "nvd-cpes.json.gz",
                                 NULL);
+  if (g_stat (full_path, &state))
+    {
+      g_free (full_path);
+      full_path = g_build_filename (GVM_SCAP_DATA_DIR,
+                                    "nvd-cpes.json",
```
mattmundell, line 2895: `update_scap_cpes_from_json_file` expects a gz file so this will always fail, right? Perhaps just return -1 here.

src/manage_sql_secinfo.c
```diff
@@ -2888,8 +2927,12 @@ update_scap_cpes ()
 
   ret = update_scap_cpes_from_json_file (full_path);
   if (ret)
-    return -1;
+    {
+      g_free (full_path);
```
mattmundell, line 2931: Nitpick: could free before if.

mattmundell (changes requested):

src/manage_sql_secinfo.c
jhelmold: I tried it explicitly once again. The function "gvm_gzip_open_file_reader_fd (..)" in "update_scap_cpes_from_json_file (..)" and the function "fopencookie (..)" within that function can also handle uncompressed files. (Timo knew that.) So it also works for nvd-cpes.json.

mattmundell: Ah OK, gzread handles it.

mattmundell (approved):

PR #2317, opened 2024-11-07
Remove delta reports feature toggle and non-default-code

a-h-abdelsalam:
## What
Remove delta reports feature toggle and non-default-code (v1)

## Why
No need to maintain v1 code any more. Delta reports v2 was enabled by default in GEA-271.

## References
GEA-701


mattmundell (approved):
Compiles fine and the reports in GSA look OK. Glad to see this code gone.

PR #2320, opened 2024-11-11
Change: Insert CVEs products from CPE matches table

a-h-abdelsalam:
## What
Insert CVEs products from CPE matches table when updating the SCAP from JSON files.

## Why
Products column was not populated during CVE update from JSON

## References
GEA-625

### DIFF
--- a/src/manage_sql_secinfo.c
+++ b/src/manage_sql_secinfo.c
@@ -3406,6 +3406,7 @@ handle_cve_configurations (resource_t cve_db_id, char * cve_id,
                            cJSON* configurations_json)
 {
   cJSON *configuration_item;
+  GString *software = g_string_new ("");
 
   cJSON_ArrayForEach (configuration_item, configurations_json)
     {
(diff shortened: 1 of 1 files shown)


jhelmold (approved):

src/manage_sql_secinfo.c
```diff
@@ -3495,11 +3496,30 @@ handle_cve_configurations (resource_t cve_db_id, char * cve_id,
...
+    if (software->len > 0)
+      {
+        gchar *quoted_software = sql_quote (software->str);
+        sql ("UPDATE scap2.cves"
+             " SET products = '%s'"
+             " WHERE id = %llu;",
+             quoted_software, cve_db_id);
+        g_free (quoted_software);
+      }
+    g_string_free (software, TRUE);
```
mattmundell, line 3521: This is indented 2 to far (compare with the return below). Made me think that the free was happening inside the iteration above.

I find this cJSON_ArrayForEach style awkward, especially when the body is so large. I think this style makes it tricky for the auto indentation in editors.

PR #2321, opened 2024-11-13
Doc: Drop mention of obsolete dependency on xml-twig-tools in Dockerfile

cfi-gb:
## What
In #2142 the dependency itself has been dropped but it was missed to remove the relevant comment as well.

## Why
Don't mention an unnecessary dependency

## References
- #2142 
- greenbone/docs#524


y0urself (approved):

bjoernricks (approved):

y0urself:
You can not use forks @cfi-gb, this is since over 2 years. Please recreate your PR with a feature branch.

PR #2322, opened 2024-11-13
Flush trailing whitespace

mattmundell:
## What

Remove all spaces from the ends of lines.

## Why

I was reviewing #2317 earlier and the trailing whitespace changes got in the way.

## How

```sh
# show
PKG=gvmd; SRC=$HOME/fresh/main; cd $SRC/$PKG && find . -type f -name \*.[ch] -exec egrep -l " +$" {} \;
# flush
PKG=gvmd; SRC=$HOME/fresh/main; cd $SRC/$PKG && find . -type f -name \*.[ch] -print0 | xargs -0 sed --in-place='' -E "s/[ ]+$//"
```

## References

See #2323 for an idea on how to keep the space out.

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -5513,7 +5513,7 @@ gmp_xml_handle_start_element (/* unused */ GMarkupParseContext* context,
 
             append_attribute (attribute_names, attribute_values, "config_id",
                               &get_reports_data->config_id);
-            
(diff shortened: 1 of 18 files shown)


a-h-abdelsalam (approved):
Nice!

PR #2327, opened 2024-11-22
Fix: Rollback on create_credential privkey error

timopollmeier:
## What
If the private key checks in create_credential fail, the SQL transaction is now rolled back.

## Why
This fixes SQL transaction errors and possible incomplete credentials being created if there are multiple GMP commands per connection and one of them is create_credential failing because of the private key.

## References
GEA-736

## Checklist

- [ ] Tests


mattmundell (approved):

mattmundell:
There's another one on line 35763 if you feel like adding it.

mattmundell (approved):

PR #2328, opened 2024-11-26
Fix: Avoid DB check inserts in cleanup-sequences

timopollmeier:
## What
When running gvmd with the --optimize=cleanup-sequences option, the database checks during initialization will avoid most inserts. Setting database_version and max_hosts in the meta table now uses INSERT ... ON CONFLICT instead of deleting the previous entry.

## Why
Skipping the inserts prevents the cleanup failing if one of the sequence has already run out of ids.
The change to the meta table update reduces the risk of the meta table id sequence running out.

## References
GEA-576

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -930,13 +930,15 @@ cert_check_time ()
  *
  * @param[in]  log_config  Log configuration.
  * @param[in]  database    Database.
+ * @param[in]  avoid_db_check_inserts  Whether to avoid inserts in DB check.
  *
  * @return 0 success, -1 error, -2 database is too old,
  *         -3 database needs to be initialised from server,
  *         -5 database is too new.
  */
 int
-manage_option_setup (GSList *log_config, const db_conn_info_t *database)
+manage_option_setup (GSList *log_config, const db_conn_info_t *database,
+                     int avoid_db_check_inserts)
 {
   int ret;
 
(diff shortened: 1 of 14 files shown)


src/manage_sql.c
```diff
@@ -16695,19 +16725,19 @@ check_db (int check_encryption_key)
...
     goto fail;
-  if (check_db_report_formats ())
+  if (check_db_report_formats (avoid_db_check_inserts))
     goto fail;
   if (check_db_report_formats_trash ())
     goto fail;
-  check_db_permissions ();
-  check_db_settings ();
+  check_db_permissions (avoid_db_check_inserts);
+  check_db_settings (avoid_db_check_inserts);
```
a-h-abdelsalam, line 16740: Some of the above functions (`check_db_roles`, `check_db_nvt_selectors`, `check_db_scanners`, `check_db_permissions`, `check_db_settings`) are not called elsewhere and they only return when `avoid_db_check_inserts = 1`. Isn't it sufficient to wrap an if around them ?

src/manage_sql_port_lists.c
```diff
@@ -2652,12 +2652,17 @@ update_port_list (port_list_t port_list, const gchar *name,
 
 /**
  * @brief Check port lists, for startup.
+ *
+ * @param[in]  avoid_db_check_inserts  Whether to avoid inserts.
  */
 void
-check_db_port_lists ()
+check_db_port_lists (int avoid_db_check_inserts)
```
a-h-abdelsalam, line 2659: The function prototype in `manage_sql_port_lists.h` could be also updated

a-h-abdelsalam (changes requested):

PR #2340, opened 2025-01-10
Fix: Use regular expressions to count & filter compliance results

a-h-abdelsalam:
## What
 Use regular expressions to count and filter compliance results instead of the `LIKE` operator.

## Why
Using `LIKE` is less restrictive with `%` matching any sequence of characters between `Compliant:` and the compliance state. One example was `INCOMPLETE` results being matched and counted as `NO` due to 'NO' appearing elsewhere in the description. 

This was leading to incorrect total compliance counts which appears in GSA as if there are hidden results.

## References
GEA-842


mattmundell (dismissed):
I'm guessing you could drop the brackets around the `\\s`s, but it looks OK.

mattmundell (approved):

PR #2346, opened 2025-01-23
Add: Setting to enable new CVE scan CPE matching

timopollmeier:
## What
The setting "CVE-CPE Matching Version" has been added that allows switching between the old "affected products" based matching for CVE scans and the new one based on the extended matching rules. For now the old version will be used by default.

## Why
The new matching does not work correctly yet, so it is disabled by default and can be enabled for future fixes.

## References
GEA-863

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -16310,6 +16310,17 @@ check_db_settings ()
           "  'User Interface Date Format',"
           "  'Preferred date format to be used in client user interfaces.',"
           "  'system_default' );");
+
+  if (sql_int ("SELECT count(*) FROM settings"
+              " WHERE uuid = '" SETTING_UUID_CVE_CPE_MATCHING_VERSION "'"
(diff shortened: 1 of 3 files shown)


src/manage_sql.c
```diff
@@ -16310,6 +16310,17 @@ check_db_settings ()
           "  'User Interface Date Format',"
           "  'Preferred date format to be used in client user interfaces.',"
           "  'system_default' );");
+
+  if (sql_int ("SELECT count(*) FROM settings"
+              " WHERE uuid = '" SETTING_UUID_USER_INTERFACE_DATE_FORMAT "'"
```
a-h-abdelsalam, line 16315:
```suggestion
              " WHERE uuid = '" SETTING_UUID_CVE_CPE_MATCHING_VERSION "'"
```

a-h-abdelsalam (changes requested):

a-h-abdelsalam (approved):

Issue #2342, opened 2025-01-15
Error while loading Report


serdam97:
Rejection: An internal error occurred while getting a report. The report could not be delivered. 

Diagnostics: Failure to receive response from manager daemon.

```
 Error     at new _Rejection ([https://](https://XXXXX:9392/assets/index-Dh8yFcKu.js:69:407308)     at https://XXXXX:9392/assets/index-Dh8yFcKu.js:69:411037
```

```
ii  greenbone-feed-sync                    24.9.0-2                             all          New script for syncing the Greenbone Community Feed (Python 3)
ii  greenbone-security-assistant           23.3.0~precompiled-0kali1            all          remote network security auditor - web interface
ii  greenbone-security-assistant-common    22.4.0-0kali1                        all          transitional package
ii  gvm                                    24.11.1                              all          remote network security auditor - metapackage and useful scripts
ii  gvm-tools                              24.8.0-1                             all          Remote control the Greenbone Vulnerability Manager
ii  gvmd                                   24.0.0-1                             amd64        Manager Module of Greenbone Vulnerability Manager
ii  gvmd-common                            24.0.0-1                             all          architecture independent files for Greenbone Vulnerability Manager
ii  libgvm22t64:amd64                      22.14.0-2                            amd64        remote network security auditor - shared libraries
ii  openvas-scanner                        23.13.1-1                            amd64        remote network security auditor - scanner
ii  ospd-openvas                           22.7.1-2                             all          OSP server implementation to allow GVM to remotely control an OpenVAS Scanner
ii  postgresql-16-pg-gvm                   22.6.5-1                             amd64        PostgreSQL extension for ical object manipulation
ii  postgresql-17-pg-gvm                   22.6.5-2+b1                          amd64        PostgreSQL extension for ical object manipulation
ii  python3-gvm                            24.8.0-1                             all          Greenbone Vulnerability Management Python Library (Python 3)
```

After the last update I have this problem when I try to load the reports and exporting them (it shows unknown error)

TotalGriffLock:
You are not alone: https://forum.greenbone.net/t/error-while-loading-report/19868/5

TotalGriffLock:
Happy to confirm that seems to have fixed it!

FlagHunter01:
Heyyyyyyyyy, so this is happening to me:

```
[...]

PR #2348, opened 2025-01-23
Fix: Prevent sporadic signals when loading reports.

jhelmold:
## What
Now the returned error code of the functions gnutls_x509_crt_get_dn (..) and gnutls_x509_crt_get_issuer_dn (..) is considered to prevent the sporadic occurrence of signals when loading reports.

## Why
This is a bug-fix.

## References
GEA-866

## Checklist

Because the bug could not be reproduced on our systems the (positive) effect of the change could not be tested.

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -527,9 +527,12 @@ get_certificate_info (const gchar* certificate, gssize certificate_len,
           gchar *buffer;
           gnutls_x509_crt_get_dn (gnutls_cert, NULL, &buffer_size);
           buffer = g_malloc (buffer_size);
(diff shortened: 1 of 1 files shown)


timopollmeier (approved):

cfi-gb:
> Because the bug could not be reproduced on our systems the (positive) effect of the change could not be tested.

Maybe the users on #2342 can give some feedback as segfaults and failed to load reports are mentioned there?

TotalGriffLock:
I will be able to check this for you tomorrow evening, I will update once I have an answer

TotalGriffLock:
Happy to confirm that seems to have fixed it!

PR #2350, opened 2025-01-24
Drop the old feed sync scripts

bjoernricks:
## What

Drop the old feed sync scripts

## Why

The old shell based feed sync scripts got replaced by the Python tool in a separate repository.


CMakeLists.txt
```diff
@@ -38,10 +38,6 @@ endif (NOT CMAKE_BUILD_TYPE)
 OPTION (ENABLE_COVERAGE "Enable support for coverage analysis" OFF)
 OPTION (DEBUG_FUNCTION_NAMES "Print function names on entry and exit" OFF)
 # the shell based scripts got replaced by https://github.com/greenbone/greenbone-feed-sync/
```
mattmundell, line 40: How about removing this line too?

bjoernricks: Missed that. Thanks for the hint!

bjoernricks:
I am not sure if merging this PR will require to increment the major version in the next release, because we remove functionality.

mattmundell (approved):

mattmundell:
> I am not sure if merging this PR will require to increment the major version in the next release, because we remove functionality.

Sounds reasonable.

Issue #2353, opened 2025-01-27
The size of the database explodes when PostgreSQL uses the archive log.


mdc-webdb:
Whenever the feeds are updated,
the entire database is rewritten instead of just the changes. This leads to an extreme use of space on the database server if the wal archive mode is used. However, this is a prerequisite for incremental database backups.

4D617274696E:
We are currently investigating a solution for this problem, no ETA yet.

timopollmeier:
I've created #2355 to limit the size of certain INSERT queries that were getting very large and there will be another patch release of gvmd once this is merged.

mdc-webdb:
Hi @timopollmeier,
I have tested it, but it is making it much worse.
Now with feed update of today, 27G of WAL archive logs are generated for an database with an size of 11GB.
```
SELECT pg_size_pretty( pg_database_size('gvmd') );
 pg_size_pretty 
----------------
 11 GB
```
After the feed update
```
du -h --max-depth=1 /var/lib/pgsql/wal_archive
27G     /var/lib/pgsql/wal_archive
```
Before:
```
du -h --max-depth=1 /var/lib/pgsql/wal_archive
30M     /var/lib/pgsql/wal_archive
```

bjoernricks:
Hi @mdc-webdb, as we don't use the wal archive mode (at least not at the moment) we don't have a solution for your issue. If you have ideas on how to improve the situation, let us know in this issue. Maybe we can take another look into it again. Until then I am sorry but we wont be able to fix it.

mdc-webdb:
Hi @bjoernricks,
I think that before a record is written to the database during the update, it must first be checked whether it already exists with the same content. This is because an SQL update statement is always executed ‘stupidly’, even if what is written is already in the database. This means that a transaction is created each time, which must be logged.

PR #2355, opened 2025-01-28
Fix: Limit query size for affected products update

timopollmeier:
## What
The update of the affected products is now limited to 1000 CVEs per SQL query. This number can be changed with the --affected-products-query-size option.
Also, an index has been added to make the queries a little faster.

## Why
This is meant to prevent the WAL from becoming very large during the affected products update.

## References
GEA-872
#2353

### DIFF
--- a/src/manage_sql_secinfo.c
+++ b/src/manage_sql_secinfo.c
@@ -72,6 +72,11 @@
 #define CPE_MAX_CHUNK_SIZE 10000
 
 /**
+ * @brief Query size for affected products updates.
+ */
+static int affected_products_query_size = AFFECTED_PRODUCTS_QUERY_SIZE_DEFAULT;
+
+/**
  * @brief Commit size for updates.
  */
 static int secinfo_commit_size = SECINFO_COMMIT_SIZE_DEFAULT;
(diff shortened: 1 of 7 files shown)


src/manage_sql_secinfo.c
```diff
@@ -4063,17 +4086,40 @@ update_scap_cves ()
...
-  sql ("INSERT INTO scap2.affected_products"
-       "  SELECT DISTINCT scap2.cpe_match_nodes.cve_id, scap2.cpes.id"
-       "    FROM scap2.cpe_match_nodes, scap2.cpe_nodes_match_criteria,"
-       "         scap2.cpe_matches, scap2.cpes"
-       "    WHERE scap2.cpe_match_nodes.id = scap2.cpe_nodes_match_criteria.node_id"
-       "      AND scap2.cpe_nodes_match_criteria.vulnerable = 1"
-       "      AND scap2.cpe_nodes_match_criteria.match_criteria_id ="
-       "            scap2.cpe_matches.match_criteria_id"
-       "      AND scap2.cpe_matches.cpe_name_id = scap2.cpes.cpe_name_id;");
+  init_iterator(&cves_iter,
```
mattmundell, line 4093: There should be a `cleanup_iterator` for this.

src/manage_sql_secinfo.c
```diff
@@ -4063,17 +4086,40 @@ update_scap_cves ()
...
+                "SELECT DISTINCT cve_id FROM scap2.cpe_match_nodes");
+
+  int count = 0;
+
+  cve_ids_buffer = g_string_new("");
```
mattmundell, line 4098: This string should be freed.

src/manage_sql_secinfo.c
```diff
@@ -4063,17 +4086,40 @@ update_scap_cves ()
...
+          g_string_truncate (cve_ids_buffer, 0);
+          g_message("%s: Products of %d CVEs processed", __func__, count);
+        }
+    }
+
+  if (cve_ids_buffer->len)
+    {
+      exec_affected_products_sql (cve_ids_buffer->str);
+      g_string_truncate (cve_ids_buffer, 0);
+      g_message("%s: Products of %d CVEs processed", __func__, count);
```
mattmundell, line 4120: Missing space before `(`. A few above too.

src/manage_sql_secinfo.c
```diff
@@ -4063,17 +4086,40 @@ update_scap_cves ()
...
-       "         scap2.cpe_matches, scap2.cpes"
-       "    WHERE scap2.cpe_match_nodes.id = scap2.cpe_nodes_match_criteria.node_id"
-       "      AND scap2.cpe_nodes_match_criteria.vulnerable = 1"
-       "      AND scap2.cpe_nodes_match_criteria.match_criteria_id ="
-       "            scap2.cpe_matches.match_criteria_id"
-       "      AND scap2.cpe_matches.cpe_name_id = scap2.cpes.cpe_name_id;");
+  init_iterator(&cves_iter,
+                "SELECT DISTINCT cve_id FROM scap2.cpe_match_nodes");
+
+  int count = 0;
```
mattmundell, line 4096: I know it's pedantic but this breaks the pattern of putting the variable declarations all together at the top of the block, and then putting the variable definitions here below. Especially because you've mixed the two styles in the function.

PR #2360, opened 2025-01-29
Fix: Made Verinice reports include attached reports.

jhelmold:
## What
It was not possible to export a "Verinice ISM" or "Verinice ISM all results" report correctly using a report config with attached report formats. That is working now.

## Why
This is a bug-fix.

## References
GEA-840

## Checklist

- [ ] Tests

### DIFF
--- a/src/manage_sql_report_formats.c
+++ b/src/manage_sql_report_formats.c
@@ -3997,18 +3997,18 @@ apply_report_format (gchar *report_format_id,
 
   rf_dependencies_string
     = sql_string ("SELECT value"
-                  "  FROM report_format_params"
-                  " WHERE report_format = %llu"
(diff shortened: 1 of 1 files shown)


a-h-abdelsalam (approved):

src/manage_sql_report_formats.c
```diff
@@ -3997,11 +3997,18 @@ apply_report_format (gchar *report_format_id,
 
   rf_dependencies_string
```
mattmundell, line 3998: A bit late but this could have an `if (report_config)` around it, because the report_config is optional. It would save the SQL query when the report_config is 0 (and maybe prevent some weird situations if someone somehow has a report_config with id 0).

PR #2362, opened 2025-01-30
Fix: Reject creating tags of the wrong subtype

a-h-abdelsalam:
## What
Reject creating tags for resources of the wrong subtype.
Creation of tags of type `audit_report`, `audit` or `policy` is now rejected for reports, tasks or scan configs respectively. Similarly creating tags of type `report`, `task` or `config` is rejected for audit reports, audits or policies.

The tag resources iterator and tag resources count are updated to account for subtypes.

## Why
Validation of type / subtype consistency when creating tags was missing.
The tags count and resources list should be consistent with the tag subtype.

## References
GEA-871


jhelmold (dismissed):

src/manage_sql.c
```diff
@@ -57061,24 +57102,29 @@ tag_add_resources_list (tag_t tag, const char *type, array_t *uuids,
...
   if (type_is_info_subtype (type))
     resource_permission = g_strdup ("get_info");
   else if (type_is_asset_subtype (type))
     resource_permission = g_strdup ("get_assets");
   else if (type_is_report_subtype (type))
     {
       resource_permission = g_strdup ("get_reports");
-      type = g_strdup("report");
+      g_free (resource_type);
+      resource_type = g_strdup("report");
```
mattmundell, line 57115:
```suggestion
      resource_type = g_strdup ("report");
```
Would be great if the CI would catch these.

timopollmeier (approved):

PR #2373, opened 2025-02-24
Change: Use SQL COPY for loading CVEs

timopollmeier:
## What
When loading the CVEs with secinfo_fast_init enabled, COPY statements will be used to load the CVEs, including references, configurations and affected products.

## Why
This is much faster than using INSERT and UPDATE statements.

## References
GEA-928

### DIFF
--- a/src/manage_sql_secinfo.c
+++ b/src/manage_sql_secinfo.c
@@ -3762,7 +3762,7 @@ handle_cve_references (resource_t cve_db_id, char *cve_id,
   int ret = 0;
   cJSON *reference_item;
   static GHashTable *urls = NULL;
-  
+
   if (secinfo_fast_init && urls == NULL)
     urls = g_hash_table_new (g_str_hash, g_str_equal);
 
@@ -3816,7 +3816,7 @@ get_cve_configuration_fields (cJSON* configuration_item,
                               int *negate)
 {
   *nodes_array = cJSON_GetObjectItemCaseSensitive (configuration_item,
-                                                  "nodes");
+                                                   "nodes");
   if (!cJSON_IsArray (*nodes_array))
     {
       g_warning ("%s: 'nodes' field missing or not an array for %s.",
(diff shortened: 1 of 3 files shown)


mattmundell:
Hey @timopollmeier there's a lot to go through here. Any chance you could break it into multiple commits so it's easier to digest? Otherwise I'm fine if someone else wants to review.

src/manage_sql_secinfo.c
```diff
@@ -63,7 +63,7 @@
  */
 #define G_LOG_DOMAIN "md manage"
 
-
+
```
mattmundell, line 66: Note that these form feeds help me navigate the sections of the file in Emacs. I'm happy for them to be removed because they're quite obscure, but I think it would be better to do it all at once for the whole repo in a separate PR.

src/manage_sql_secinfo.c
```diff
@@ -3601,285 +3607,843 @@ set_root_id (long int id, long int root_id)
...
+      (*cpe_db_id_ptr)++;
+      affected_product_cpe_db_id = (*cpe_db_id_ptr);
+      gchar *escaped_cpe_name, *escaped_cpe_name_id;
+
+      escaped_cpe_name = sql_copy_escape (match_cpe_name);
+      escaped_cpe_name_id = sql_copy_escape (match_cpe_name_id);
+
+      g_hash_table_insert (hashed_cpes,
+                            g_strdup (match_cpe_name),
+                            GINT_TO_POINTER (*cpe_db_id_ptr));
```
a-h-abdelsalam, line 4072: The indentation might need a fix here.
```suggestion
                           g_strdup (match_cpe_name),
                           GINT_TO_POINTER (*cpe_db_id_ptr));
```

src/manage_sql_secinfo.c
mattmundell: Thanks.

Here's what I use, by the way.
```sh
# show
PKG=gvmd; SRC=$HOME/fresh/main; cd $SRC/$PKG && find . -type f -name \*.[ch] -exec egrep -l " +$" {} \;
# flush
PKG=gvmd; SRC=$HOME/fresh/main; cd $SRC/$PKG && find . -type f -name \*.[ch] -print0 | xargs -0 sed --in-place='' -E "s/[ ]+$//"
```
Would be nice to check it in CI.

PR #2377, opened 2025-02-27
Change: Handle timeout more like other VT prefs

timopollmeier:
## What
The timeout preferences of VTs in configs/policies are now handled more like normal VT preferences instead of a scanner ones, which is also more consistent with default preferences of the VTs themselves.

## Why
This simplifies the special handling needed and fixes timeouts in the config being overwritten by their default values.

## References
GEA-799

### DIFF
--- a/src/manage_sql_configs.c
+++ b/src/manage_sql_configs.c
@@ -1747,9 +1747,11 @@ manage_nvt_preference_add (const char *name, const char *value,
  *
  * @param[in]  iterator  Iterator.
  * @param[in]  oid       OID of NVT, NULL for all preferences.
+ * @param[in]  include_timeout  Whether to include the timeout preference.
  */
 void
-init_nvt_preference_iterator (iterator_t* iterator, const char *oid)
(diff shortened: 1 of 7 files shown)


src/manage_sql_configs.c
```diff
@@ -1762,15 +1764,17 @@ init_nvt_preference_iterator (iterator_t* iterator, const char *oid)
...
                      /* Task preferences. */
                      " AND name != 'max_checks'"
                      " AND name != 'max_hosts'"
                      " ORDER BY name ASC",
                      sql_ilike_op (),
                      quoted_oid,
-                     quoted_oid,
+                     include_timeout ? "" : "AND name != '",
+                     include_timeout ? "" : quoted_oid,
+                     include_timeout ? "" : "'",
```
a-h-abdelsalam, line 1777: Does this need to have `:0:entry:timeout` appended ? Not sure.

a-h-abdelsalam (approved):

PR #2380, opened 2025-03-06
Openvasd integration

jjnicola:
## What
Add: functions to handle an Openvasd scan task
SC-1056
SC-1053
SC-1058
SC-1057
SC-1060

## Checklist

- [ ] Tests


src/CMakeLists.txt
```diff
@@ -31,7 +31,11 @@ pkg_check_modules (LIBGVM_BASE REQUIRED libgvm_base>=22.17)
 pkg_check_modules (LIBGVM_UTIL REQUIRED libgvm_util>=22.17)
 pkg_check_modules (LIBGVM_OSP REQUIRED libgvm_osp>=22.17)
 pkg_check_modules (LIBGVM_GMP REQUIRED libgvm_gmp>=22.17)
-pkg_check_modules (LIBGVM_OPENVASD REQUIRED libgvm_openvasd>=22.17)
+if (OPENVASD)
+    pkg_check_modules (LIBGVM_OPENVASD REQUIRED libgvm_openvasd>=22.17)
+else (OPENVASD)
+    message (STATUS "OPENVASD flag is not enabled")
```
bjoernricks, line 37: Shouldn't be this a failure? if somebody enables OPENVASD he wants to build with it. and if that isn't possible because some dependency isn't installed the configure step should fail in my opinion.

src/gvmd.c
```diff
@@ -1106,11 +1106,15 @@ handle_sigabrt_simple (int signal)
 static int
 update_nvt_cache_osp (const gchar *update_socket)
 {
+#ifdef OPENVASD
+  setproctitle ("Openvasd: Updating NVT cache");
```
bjoernricks, line 1110: Can we have a consistent naming? just openvasd. not OpenVASD and not Openvasd.
```suggestion
  setproctitle ("openvasd: Updating NVT cache");
```

src/gvmd.c
```diff
@@ -2865,6 +2869,8 @@ gvmd (int argc, char** argv, char *env[])
         type = SCANNER_TYPE_OPENVAS;
       else if (!strcasecmp (scanner_type, "OSP-Sensor"))
         type = SCANNER_TYPE_OSP_SENSOR;
+      else if (!strcasecmp (scanner_type, "Openvasd"))
```
bjoernricks, line 2872:
```suggestion
      else if (!strcasecmp (scanner_type, "openvasd"))
```

src/gvmd.c
```diff
@@ -2907,6 +2913,8 @@ gvmd (int argc, char** argv, char *env[])
             type = SCANNER_TYPE_OPENVAS;
           else if (!strcasecmp (scanner_type, "OSP-Sensor"))
             type = SCANNER_TYPE_OSP_SENSOR;
+          else if (!strcasecmp (scanner_type, "Openvasd"))
```
bjoernricks, line 2916:
```suggestion
          else if (!strcasecmp (scanner_type, "openvasd"))
```

src/CMakeLists.txt
timopollmeier: It think it does this already: If you enable the OPENVASD flag the configuration will fail if you don't have the library installed because the pkg_check_modules has the REQUIRED option.

ozgen (approved):

PR #2384, opened 2025-03-10
Fix: Allow missing cpeMatch in CVE configurations

timopollmeier:
## What
Allow missing cpeMatch in CVE configurations.

## Why
The CVEs feed can contain configurations without a cpeMatch field if a CVE only affects products for which no CPE has been assigned.

## References
GEA-981


bjoernricks (approved):

cfi-gb:
Only for tracking purposes (e.g. some one searching for the relevant error here on Github), this seems to be relevant to:

https://forum.greenbone.net/t/no-scap-database-gets-current-result-but-null-values/20628

which had the following message attached:

```
handle_cve_configurations: cpeMatch missing or not an array for CVE-2024-1719
```

Issue #2316, opened 2024-10-30
No certificate validation possible when sending mail via gvmd in docker


perlfisch:
### Expected behavior

The official gvmd Docker container should be able to send mail and validate the submission servers certificate.

### Actual behavior

No CAs are available in the container gvmd. `ca-certificates` is not installed and validation of all certificates fails. The only way to get mail working persistently is by modifying the containers compose to mount the hosts ca-certificates.crt or disabling certificate validation entirely.

Disabling certificate validation, which is dangerous and should not be done, is the only solution that comes up when [googling the error](https://forum.greenbone.net/t/alert-email-problems/15385/15) and will likely be implemented by many people trying to get mail working in the container.

I've confirmed that manually installing `ca-certificates` into a running gvmd container fixes the problem.

### Steps to reproduce

1. Setup gvmd via docker based on https://greenbone.github.io/docs/latest/22.4/container/index.html
2. Configure mail as described in https://greenbone.github.io/docs/latest/22.4/container/workflows.html#setting-up-a-mail-transport-agent-inside-docker-container to any public provider like mailgun, google, m365 etc.
3. Try sending any mail, either through gvm or manually via sendmail.

### GVM versions

Stable official docker containers as of 2024-10-30

### Environment

**Operating system:**

Debian 12 stable

**Installation method / source:** (packages, source installation)

Official docker compose

### Logfiles

```

email: system failed with ret 17664, 69, read FROM TO < /tmp/gvmd-args-Ln7YTZ; /usr/sbin/sendmail -f "$FROM" "$TO" < /tmp/gvmd-content-8HCXWY > /dev/null 2>&1

```

```
```

perlfisch:
Installing anything into a Docker container is not persistent and will be gone once it's restarted. Instead of documenting manual changes that have to be reapplied after restarting the container, `ca-certificates` should be installed via apt as part of the [Dockerfile](https://github.com/greenbone/gvmd/blob/main/.docker/prod.Dockerfile) to persistently fix this issue.

cfi-gb:
You could try to create a PR to the relevant file and see if the repo maintainers are accepting such a change (see previous notes about usually requiring community contributions for such topics).

PR #2390, opened 2025-03-14
CA certificates

EarMaster:
## What

Install CA certificates when MTA is setup for successful TLS verification.

## Why

A secure SMTP configuration requires a successful TLS verification. This silently fails without `ca-certificates` installed.

## References

#2316

### DIFF
--- a/.docker/setup-mta.sh
+++ b/.docker/setup-mta.sh
@@ -16,7 +16,4 @@ if [ -n "$MTA_HOST" ]; then
     [ -n "$MTA_LOGFILE" ] && echo "logfile $MTA_LOGFILE" >> $CONFIG
     chown gvmd:mail $CONFIG
     chmod 750 $CONFIG
-    echo "installing ca-certificates for tls verification"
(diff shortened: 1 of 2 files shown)


bjoernricks (changes requested):
Hi, thanks for the hint. Could you just add ca-certificates to https://github.com/greenbone/gvmd/blob/main/.github/runtime-dependencies.list instead? IMHO it should be always installed in the container image.

EarMaster:
I also considered this, but as this seems only be needed when you use a MTA I decided to put it there. I added the change, it is up to you what you want to use.

bjoernricks (approved):

bjoernricks:
ca-certs are also required if you want to use curl with https. so it's always a good idea to install them.

bjoernricks:
Our CI/CD doesn't work with PRs from forked repos sadly. To shortcut the situation I created https://github.com/greenbone/gvmd/pull/2404 that integrates your change.

PR #2393, opened 2025-03-17
Fix: openvasd scanner bool server preferences

jjnicola:
## What
Fix: openvasd scanner bool server preferences
Jira: SC-1255

## Why
Send the bool values as integers, as OSPD expects. Then, ospd-only, openvad+ospd and openvasd-openvas scenarios receives the type

## Checklist

- [ ] Tests


timopollmeier (approved):

jjnicola:
greenbone/openvas-scanner#167

PR #2404, opened 2025-03-31
Install ca-certificates in container images

bjoernricks:
## What

Install ca-certificates in container images
## Why
Add the ca-certificates to the container images to allow SSL/TLS verification for example when using curl with http or TLS for the MTA setup.

## References

Replaces #2390
Closes https://github.com/greenbone/gvmd/issues/2316 https://github.com/greenbone/docs/issues/483

### DIFF
--- a/.github/runtime-dependencies.list
+++ b/.github/runtime-dependencies.list
@@ -49,6 +49,10 @@
 # Required for set up certificates for GVM
 # gnutls-bin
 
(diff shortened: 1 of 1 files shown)


cfi-gb:
> Replaces #2390

Seems to also close:
- #2316
- greenbone/docs#483

timopollmeier (approved):

PR #2411, opened 2025-04-08
Add: Added loading of VT metadata from JSON feed file.

a-h-abdelsalam:
## What

Added the loading of VT metadata from a JSON feed file. This is currently behind a feature switch `FEED_VT_METADATA` which is disabled by default.

`GVM_NVT_DATA_DIR` should be set to the path where the vt metadata feed file along with a timestamp can be located.

VT metadata can be loaded from the feed regardless of whether OPENVASD is enabled.

Scanner preferences still need to be read from the scanner.

## Why
This should reduce scanner ramp-up time.

Depends on: https://github.com/greenbone/gvm-libs/pull/915

## References
GEA-936

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -5292,11 +5292,27 @@ feed_sync_required ()
         break;
     }
 
-  if (nvts_feed_version_status () == 1)
+#if FEED_VT_METADATA == 0
+  if (nvts_feed_version_status_from_scanner () == 1)
     {
       g_debug ("%s: NVTs need to be updated", __func__);
       return TRUE;
     }
+#else
+  feed_status_ret = nvts_feed_version_status_from_timestamp ();
+  switch (feed_status_ret)
+    {
+      case 1:
+      case 2:
+      case 3:
+      case 4:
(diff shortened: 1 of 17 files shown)


src/manage.c
```diff
@@ -5362,11 +5362,27 @@ feed_sync_required ()
...
     {
       g_debug ("%s: NVTs need to be updated", __func__);
       return TRUE;
     }
+#else
+  feed_status_ret = nvts_feed_version_status_from_timestamp ();
+  switch (feed_status_ret)
+    {
+      case 1:
+      case 2:
```
timopollmeier, line 5376: Will the tables be created later if they are missing at this point? If so, it is not clear how.

timopollmeier (changes requested):

src/manage.c
a-h-abdelsalam: The function `prepare_nvts_insert` called from `update_nvts_from_json_file` should create all the `_rebuild `tables followed by `finalize_nvts_insert` to rename them. Is there something I am missing here ?

PR #2425, opened 2025-04-23
Change: set default vector to empty string

janlucasschwoerer-greenbone:
## What
Set default value of vector to empty string

## Why
Prevent issues with what is shown when the cvss vector is empty or not included.

## References
https://jira.greenbone.net/browse/GEA-888

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -13813,7 +13813,9 @@ handle_get_info (gmp_parser_t *gmp_parser, GError **error)
                              cve_info_iterator_severity (&info)
                               ? cve_info_iterator_severity (&info)
                               : "",
-                             cve_info_iterator_vector (&info),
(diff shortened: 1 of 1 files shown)


src/gmp.c
```diff
@@ -13813,7 +13813,9 @@ handle_get_info (gmp_parser_t *gmp_parser, GError **error)
                              cve_info_iterator_severity (&info)
                               ? cve_info_iterator_severity (&info)
                               : "",
-                             cve_info_iterator_vector (&info),
+                              cve_info_iterator_vector (&info)
```
mattmundell, line 13816:
```suggestion
                             cve_info_iterator_vector (&info)
```

timopollmeier (approved):

PR #2437, opened 2025-05-08
Fix: Convert integer port to string using g_strdup_printf

ozgen:
## What

Replaced a call to g_strdup(res->port) with g_strdup_printf("%d", res->port) in add_openvasd_result_to_report() to fix a type mismatch.

## Why

Using g_strdup() caused a compiler error due to implicit int-to-pointer conversion.

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -57186,7 +57186,7 @@ add_openvasd_result_to_report (openvasd_result_t res, gpointer *results_aux)
   test_id = res->oid;
   host = res->ip_address;
   hostname = res->hostname;
-  port = g_strdup (res->port);
+  port = g_strdup_printf ("%d", res->port);
 
   /* Add report host if it doesn't exist. */
   manage_report_host_add (rep_aux->report, host, 0, 0);


jhelmold (approved):

src/manage_sql.c
```diff
@@ -57186,7 +57186,7 @@ add_openvasd_result_to_report (openvasd_result_t res, gpointer *results_aux)
   test_id = res->oid;
   host = res->ip_address;
   hostname = res->hostname;
-  port = g_strdup (res->port);
+  port = g_strdup_printf ("%d", res->port);
```
mattmundell, line 57189: @ozgen @jhelmold I'm wondering how this is correct. `port` in `openvasd_result_t` was changed to a `gchar *` in https://github.com/greenbone/gvm-libs/pull/920.

ozgen: @mattmundell I’m aware of my mistakes and will correct them in the next PR.

ozgen: @mattmundell  
here is my next PR : https://github.com/greenbone/gvmd/pull/2439

PR #2439, opened 2025-05-08
Fix: Revert incorrect port formatting from int to gchar*

ozgen:
## What

Revert incorrect port formatting from int to gchar*

## Why

The original changes incorrectly treated res->port, a gchar *, as an int, which led to heap corruption at runtime during scan execution. This caused instability and failure in openvasd processes. The fix restores the correct type handling.

## References

https://github.com/greenbone/gvmd/pull/2437

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -57175,7 +57175,7 @@ add_openvasd_result_to_report (openvasd_result_t res, gpointer *results_aux)
   test_id = res->oid;
   host = res->ip_address;
   hostname = res->hostname;
-  port = g_strdup (res->port);
+  port = res->port;
 
   /* Add report host if it doesn't exist. */
   manage_report_host_add (rep_aux->report, host, 0, 0);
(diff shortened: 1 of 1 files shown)


src/manage_sql.c
```diff
@@ -57038,7 +57038,7 @@ check_openvasd_result_exists (report_t report, task_t task,
   else if (res->port > 0)
     {
       char buf[6];
-      snprintf (buf, sizeof(buf) , "%d", res->port);
+      snprintf (buf, sizeof(buf) , "%s", res->port);
```
mattmundell, line 57041: This needs a better solution, because the code around it is still expecting port to be an int. For example, `buf` is only 6 chars long.

I expect the whole if-elseif-else block can be replaced with `port_str = res->port ?: ""` but @jjnicola will have a better idea.

Note that currently `port_str` is leaking, and `port_str = ""` should have been `port_str = g_strdup ("")`.

mattmundell (changes requested):

src/manage_sql.c
```diff
@@ -57212,7 +57201,6 @@ add_openvasd_result_to_report (openvasd_result_t res, gpointer *results_aux)
         }
       desc = res->message;
       g_free (hash_value);
-      g_free (port);
```
mattmundell, line 57215: It's freeing `port` at the end too.

mattmundell (changes requested):

PR #2442, opened 2025-05-13
Throw an error when providing an invalid LDAP certificate.

jhelmold:
## What
Now an error is thrown, when an invalid certificate is provided in the Administration/LDAP dialog.

## Why
This fixes a bug.

## References
GEA-1072

## Checklist

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -24815,9 +24815,12 @@ gmp_xml_handle_end_element (/* unused */ GMarkupParseContext* context,
                       setting = g_slist_next (setting);
                     }
 
-                  manage_set_ldap_info (ldap_enabled, ldap_host, ldap_authdn,
-                                        ldap_plaintext, ldap_cacert,
-                                        ldap_ldaps_only);
+                  if (manage_set_ldap_info (ldap_enabled, ldap_host, ldap_authdn,
+                                            ldap_plaintext, ldap_cacert,
+                                            ldap_ldaps_only))
+                    SEND_TO_CLIENT_OR_FAIL (XML_ERROR_SYNTAX
+                                             ("modify_auth",
+                                              "Invalid certificate"));
                 }
               if (strcmp (group, "method:radius_connect") == 0)
                 {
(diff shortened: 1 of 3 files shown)


src/gmp.c
```diff
@@ -24815,9 +24815,12 @@ gmp_xml_handle_end_element (/* unused */ GMarkupParseContext* context,
                       setting = g_slist_next (setting);
                     }
 
-                  manage_set_ldap_info (ldap_enabled, ldap_host, ldap_authdn,
-                                        ldap_plaintext, ldap_cacert,
-                                        ldap_ldaps_only);
+                  if (manage_set_ldap_info (ldap_enabled, ldap_host, ldap_authdn,
+                                            ldap_plaintext, ldap_cacert,
+                                            ldap_ldaps_only))
+                    SEND_TO_CLIENT_OR_FAIL (XML_ERROR_SYNTAX
```
mattmundell, line 24821: This will continue around the loop and end at L24854 where it will send a second response. It should have a break like the `SEND_TO_CLIENT_OR_FAIL` above.

mattmundell (changes requested):

src/gmp.c
mattmundell: Hmm, actually the break above it also an error because the OK will also be sent, from what I can see.

The break is supposed to break out of the whole command, but it is just breaking out of the while loop.

Might have to use a flag variable. This case has got a bit long though, would be nice to move the loop body to a function.

src/manage_sql.c
```diff
@@ -53864,11 +53865,31 @@ manage_set_ldap_info (int enabled, gchar *host, gchar *authdn,
...
   if (cacert)
     {
-      sql ("DELETE FROM meta WHERE name LIKE 'ldap_cacert';");
-      quoted = sql_quote (cacert);
-      sql ("INSERT INTO meta (name, value) VALUES ('ldap_cacert', '%s');",
-           quoted);
-      g_free (quoted);
+      /* check if the certificate is valid */
+      time_t activation_time, expiration_time;
+      gchar *md5_fingerprint, *issuer;
```
mattmundell, line 53870: I think you could leave these out, to save having to free them. `get_certificate_info` only sets them if they're given, and the return code is the same either way.

mattmundell: Similar for the times.

PR #2448, opened 2025-05-19
Change: Move processing of imported reports to a scheduled process.

a-h-abdelsalam:
## What
Move processing of imported reports to a scheduled process.
Use semaphores to limit the number of open database connections and concurrent reports processed.

- Added setting `max-database-connections` for limiting open database connections and `db-semaphore-timeout` to set the timeout when no connections are available

- Added `max-concurrent-report-processing` to limit the number of imported reports processed concurrently

## Why

To avoid exhausting PostgreSQL connection slots when importing a large number of reports. 

## References

GEA-1025


src/manage.c
```diff
@@ -5413,6 +5428,130 @@ manage_sync (sigset_t *sigmask_current,
...
+{
+  lockfile_t lockfile;
+  iterator_t reports;
+  report_t report;
+  int pid, ret;
+
+  reinit_manage_process ();
+  manage_session_init (current_credentials.uuid);
+
+  init_iterator (&reports, "SELECT id FROM reports"
```
mattmundell, line 5447: Too much for me to review all of this, but SQL should go in `manage_sql.c` (or some other `*_sql.c`). One option would be to add add a new iterator, see for example `init_report_iterator_task`.

timopollmeier (approved):

PR #2463, opened 2025-06-03
Change: use defines for report format UUIDs

mattmundell:
## What

Use `#defines` for UUIDs of the XML and Anonymous XML report formats.

## Why

Neater than spreading the UUIDs around.


bjoernricks (approved):
I always like such changes :-)

PR #2470, opened 2025-06-10
Fix: Avoid clearing of name and comment if omitted when modifying alerts

a-h-abdelsalam:
## What

Avoid clearing of name and comment if omitted when modifying alerts. Now they are unchanged if omitted and cleared only when explicitly specified without a value.

Example: 
`<modify_alert alert_id="8044940c-abe7-4cdf-b1ec-1579e24f447c"><name>My Test Alert</name><comment></comment><method>Email<data><name>message</name> Scan has not finished. </data></method></modify_alert>`

should clear the comment.

## Why
Previously they were cleared if omitted from the command.

## References
GEA-1082

### DIFF
--- a/src/manage_sql_alerts.c
+++ b/src/manage_sql_alerts.c
@@ -1134,8 +1134,7 @@ modify_alert (const char *alert_id, const char *name, const char *comment,
               GPtrArray *method_data)
 {
   int index, ret;
-  gchar *quoted_name, *quoted_comment, *item;
-  gchar *new_name, *new_comment;
+  gchar *new_name, *new_comment, *item;
   alert_t alert;
   filter_t filter;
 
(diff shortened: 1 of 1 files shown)


src/manage_sql_alerts.c
```diff
@@ -1135,6 +1135,7 @@ modify_alert (const char *alert_id, const char *name, const char *comment,
 {
   int index, ret;
   gchar *quoted_name, *quoted_comment, *item;
```
mattmundell: Could move these down into the blocks that use them, if you like.

mattmundell (dismissed):

src/manage_sql_alerts.c
mattmundell: (I meant quoted_name and quoted_comment.)

a-h-abdelsalam: good idea, moved them in [1873423](https://github.com/greenbone/gvmd/commit/1873423cfeea17aa8ddfbc42b074185e81036821)

mattmundell (approved):

PR #2474, opened 2025-06-15
Added the agent controller sensor scanner type to gvmd.

jhelmold:
## What
Added the agent controller sensor scanner type (SCANNER_TYPE_AGENT_CONTROLLER_SENSOR) to gvmd.

## Why
This is needed for the implementation of agents.

## References
GEA-964

## Checklist

- [ ] Tests


src/gmp.c
```diff
@@ -17214,7 +17214,7 @@ handle_get_scanners (gmp_parser_t *gmp_parser, GError **error)
         }
 #endif
 #if ENABLE_AGENTS
-      if ((scanner_iterator_type (&scanners) == SCANNER_TYPE_AGENT_CONTROLLER)
+      if ((scanner_iterator_type (&scanners) == SCANNER_TYPE_AGENT_CONTROLLER_SENSOR)
```
timopollmeier: I think `SCANNER_TYPE_AGENT_CONTROLLER` should also still be handled here.

src/gmp.c
```diff
@@ -17223,7 +17223,7 @@ handle_get_scanners (gmp_parser_t *gmp_parser, GError **error)
           if (!agent_control_get_details_from_iterator (&scanners, &desc, &params))
             {
               SENDF_TO_CLIENT_OR_FAIL
-               ("<info><scanner><name>agent-controller</name><version>0.1</version>"
+               ("<info><scanner><name>agent-controller-sensor</name><version>0.1</version>"
```
timopollmeier: I think this should remain as "agent-controller" because it is the actual scanner is the same but OSP sensors are also not handled, so maybe we should clarify this first.

timopollmeier (changes requested):

PR #2475, opened 2025-06-16
Move and simplify OSP scan handling

timopollmeier:
## What
Most functions to handle OSP scans are moved to new manage_osp files
and functions shared between OSP and openvasd scans are moved to
manage_openvas.
The data used to connect to an OSP scanner is now bundled inside a
struct type, osp_connect_data_t so individual pieces like the host,
port and keys don't have to be passed individually.

## Why
This is a preparation for creating a gvmd scan queue.

Moving the OSP functions reduces the size of the more generic `manage` and `manage_sql` files.
The use of the struct makes functions using OSP connections easier to read and keep
consistent.

## References
GEA-1106


src/manage_osp.c
```diff
@@ -0,0 +1,1166 @@
...
+ *         osp_connect_data_free
+ */
+osp_connect_data_t *
+osp_connect_data_from_scanner (scanner_t scanner)
+{
+  gboolean has_relay;
+  osp_connect_data_t *conn_data = g_malloc0 (sizeof (osp_connect_data_t));
+
+  has_relay = scanner_has_relay (scanner);
+  conn_data->host = scanner_host (scanner, has_relay);
```
a-h-abdelsalam: I think this line is duplicated ?
```suggestion
```

src/manage_openvas.c
```diff
@@ -0,0 +1,346 @@
...
+ */
+
+/**
+ * @file manage_osp_credentials.c
+ * @brief Greenbone Vulnerability Manager OSP-style credentials handling.
+ */
+
+#include "manage_openvas.h"
+#include "manage_sql.h"
+
```
a-h-abdelsalam, line 13: Do you want to add a log domain ?

src/manage_osp.c
```diff
@@ -0,0 +1,1166 @@
...
+
+/**
+ * @file scan_handler.c
+ * @brief Headers for Greenbone Vulnerability Manager OSP scan handling.
+ */
+
+#include "ipc.h"
+#include "manage_osp.h"
+#include "manage_sql.h"
+
```
a-h-abdelsalam, line 14: Same here for the log domain

a-h-abdelsalam (approved):

PR #2478, opened 2025-06-20
Add scan queue handling for OSP scans

timopollmeier:
## What
A new scan queue is added which can be enabled to limit the number of
scan handlers which are active at the same time by periodically
switching which tasks / reports are being updated.

## Why
This allows more tasks to be running at the same time without consuming
resources by having the scan handler run the whole time.

## References
GEA-1106

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -56,6 +56,7 @@
 #include "manage_port_lists.h"
 #include "manage_report_configs.h"
 #include "manage_report_formats.h"
+#include "manage_scan_queue.h"
 #include "manage_sql.h"
 #include "manage_sql_secinfo.h"
 #include "manage_sql_nvts.h"
@@ -4151,6 +4152,21 @@ manage_sync (sigset_t *sigmask_current,
 }
 
 /**
+ * @brief Handle queued task actions like the scan queue or report processing.
+ */
+void
+manage_queued_task_actions ()
+{
+  reinit_manage_process ();
+  manage_session_init (current_credentials.uuid);
+  
+  setproctitle ("Manage process report imports");
+  manage_process_report_imports ();
+  setproctitle ("Manage scan queue");
+  manage_handle_scan_queue ();
+}
+
+/**
  * @brief Perform any processing of imported reports that is due.
  *
  * In gvmd, periodically called from the main daemon loop.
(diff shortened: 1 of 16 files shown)


src/manage.c
```diff
@@ -1528,13 +1529,58 @@ fork_osp_scan_handler (task_t task, target_t target, int from,
...
+      g_free (report_id);
+      return -1;
+    }
+
+  scan_queue_add (report);
+  set_task_run_status (task, TASK_STATUS_REQUESTED);
+  set_report_scan_run_status (report, TASK_STATUS_REQUESTED);
+  g_debug ("%s: report %s (%llu) added to scan queue",
+           __func__, report_id, report);
+  return 0;
```
a-h-abdelsalam, line 1582: `report_id` could be freed here.
```suggestion
  g_free (report_id);
  return 0;
```

src/manage_scan_handler.c
```diff
@@ -0,0 +1,288 @@
...
+
+  return 0;
+}
+
+/**
+ * @brief Fork a new handler process for a given scan queue entry.
+ * 
+ * @param[in]  report_id  UUID of the scan to handle.
+ * @param[in]  report     Row id of the scan to handle.
+ * @param[in]  task       Row id of the scan to handle.
```
a-h-abdelsalam:
```suggestion
 * @param[in]  report     Row id of the report.
 * @param[in]  task       Row id of the task.
```

src/manage_sql.c
```diff
@@ -40255,16 +40272,20 @@ manage_report_host_add (report_t report, const char *host, time_t start,
...
   sql ("INSERT INTO report_hosts"
        " (report, host, start_time, end_time, current_port, max_port)"
        " SELECT %llu, '%s', %lld, %lld, 0, 0"
        " WHERE NOT EXISTS (SELECT 1 FROM report_hosts WHERE report = %llu"
        "                   AND host = '%s');",
        report, quoted_host, (long long) start, (long long) end, report,
        quoted_host);
+  report_host = sql_int64_0 ("SELECT id FROM report_hosts"
+                             " WHERE report = %llu AND host = '%s';",
+                             report, quoted_host);
```
a-h-abdelsalam: Could `RETURNING` be used here to avoid the separate query ?

timopollmeier: Currently, using `RETURNING` would not cover the case where the ´report_host` rowid already exists.
If I added an `UNIQUE` constraint to the `report_hosts` table, I could do that by avoiding duplicates with a `ON CONFLICT` clause.

PR #2485, opened 2025-07-16
Add: Added new GMP commands for OCI image targets

a-h-abdelsalam:
## What

Added new GMP commands for OCI image targets: `create_oci_image_target`, `modify_oci_image_target`, `get_oci_image_targets `and `delete_oci_image_target`.

Changes are behind a feature toggle: `ENABLE_CONTAINER_SCANNING`

## Why
Required for container scanning

## References
GEA-1161


src/gmp.c
```diff
@@ -5517,6 +5538,9 @@ gmp_xml_handle_start_element (/* unused */ GMarkupParseContext* context,
 
             set_client_state (CLIENT_GET_OVERRIDES);
           }
+#if ENABLE_CONTAINER_SCANNING
```
ozgen, line 5541: this should be over from `GET_OVERRIDES`

src/gmp.c
```diff
@@ -21284,6 +21343,13 @@ gmp_xml_handle_end_element (/* unused */ GMarkupParseContext* context,
           set_client_state (CLIENT_AUTHENTIC);
         break;
 
+      case CLIENT_CREATE_OCI_IMAGE_TARGET:
+        if (create_oci_image_target_element_end (gmp_parser,
+                                                 error,
+                                                element_name))
```
ozgen: add 1 space to `element_name` for indentation

src/gmp.c
```diff
@@ -28296,6 +28368,9 @@ gmp_xml_handle_text (/* unused */ GMarkupParseContext* context,
       APPEND (CLIENT_CREATE_OVERRIDE_THREAT,
               &create_override_data->threat);
 
+      case CLIENT_CREATE_OCI_IMAGE_TARGET:
```
ozgen: can you reorder the `CLIENT_CREATE_OCI_IMAGE_TARGET`, this should be below the `CLIENT_CREATE_NOTE`

src/gmp_oci_image_targets.c
```diff
@@ -0,0 +1,804 @@
...
+          case -1:
+          default:
+            SEND_TO_CLIENT_OR_FAIL
+             (XML_INTERNAL_ERROR ("create_oci_image_target"));
+            log_event_fail ("image_target", "OCI Image Target",
+                            NULL, "created");
+            break;
+        }
+
+      create_oci_image_target_reset ();
```
ozgen, line 175: add free for `error_message`

src/manage_oci_image_targets.c
```diff
@@ -0,0 +1,263 @@
...
+  if (!rest || g_strcmp0 (rest, "") == 0)
+    {
+      g_warning ("%s: OCI URL is empty after prefix", __func__);
+      return -1;
+    }
+
+  gchar **parts = g_strsplit (rest, "/", -1);
+  int parts_len = g_strv_length (parts);
+
+  if (g_str_has_prefix (parts[0], "[")) // ipv6 in brackets
```
ozgen, line 104: Do we need to check `part[0] `is NULL or not?

src/manage_oci_image_targets.c
```diff
@@ -0,0 +1,263 @@
...
+  const gchar *rest = oci_url + 6;
+
+  if (!rest || g_strcmp0 (rest, "") == 0)
+    {
+      g_warning ("%s: OCI URL is empty after prefix", __func__);
+      return -1;
+    }
+
+  gchar **parts = g_strsplit (rest, "/", -1);
+  int parts_len = g_strv_length (parts);
```
ozgen, line 102: Do we need to check` if parse_len == 0` ?

src/manage_oci_image_targets.c
```diff
@@ -0,0 +1,263 @@
...
+      g_free (host);
+      g_free (port);
+      return -1;
+    }
+
+  g_free (host);
+
+  if (port && *port)
+    {
+      int port_num = atoi (port);
```
ozgen, line 163: `atoi()` returns 0 on both "0" and invalid inputs (like "abc"). Use `strtol()` instead

src/manage_oci_image_targets_tests.c
```diff
@@ -0,0 +1,173 @@
...
+
+Ensure (manage_oci_image_targets, rejects_invalid_ipv6)
+{
+  gchar *given;
+  given = g_strdup ("oci://[]:12345/myregistry.com");
+
+  assert_that (valid_oci_url(given), is_equal_to (-1));
+  g_free (given);
+}
+
```
ozgen, line 141: can you add also test scenatio with empty input like "oci://"?

src/manage_resources.h
```diff
@@ -38,6 +38,7 @@ typedef resource_t report_format_param_t;
 typedef resource_t role_t;
 typedef resource_t note_t;
 typedef resource_t nvt_t;
+typedef resource_t oci_image_target_t;
```
ozgen: I prefer to define `oci_image_target_t` in `manage_oci_image_targets.h`.

src/manage_sql_oci_image_targets.c
```diff
@@ -0,0 +1,918 @@
+#include "debug_utils.h"
```
ozgen, line 6: Add Licence comment in this file

src/manage_sql_oci_image_targets.c
```diff
@@ -0,0 +1,918 @@
...
+      sql_rollback ();
+      return 1;
+    }
+  
+  quoted_name = sql_quote (name ?: "");
+
+  if (!validate_oci_image_references (image_references, error_message))
+    {
+      sql_rollback ();
+      return 2;
```
ozgen, line 67: we need to free the quoted_name before return 2

src/manage_sql_oci_image_targets.c
```diff
@@ -0,0 +1,918 @@
...
+    {
+      if (strcmp (credential_id, "0"))
+        {
+          gchar *type;
+          if (find_credential_with_permission (credential_id,
+                                               &credential,
+                                               "get_credentials"))
+            {
+              sql_rollback ();
+              return -1;
```
ozgen, line 90: we need to free quoted_name and quoted_comment before return

src/manage_sql_oci_image_targets.c
```diff
@@ -0,0 +1,918 @@
...
+            }
+
+          if (credential == 0)
+            {
+              sql_rollback ();
+              return 4;
```
ozgen, line 99: same here

src/manage_sql_oci_image_targets.c
```diff
@@ -0,0 +1,918 @@
...
+            }
+
+          type = credential_type (credential);
+          if (strcmp (type, "up"))
+            {
+              sql_rollback ();
```
ozgen, line 105: we need to free type, quoted_name and quoted_comment before return

src/manage_sql_oci_image_targets.c
```diff
@@ -0,0 +1,918 @@
...
+              return 5;
+            }
+          g_free (type);
+        }
+      else
+        {
+          sql_rollback ();
+          return 3;
```
ozgen, line 120: we need to free quoted_name and quoted_comment before return

src/manage_sql_oci_image_targets.c
```diff
@@ -0,0 +1,918 @@
...
+          if (credential == 0)
+            {
+              sql_rollback ();
+              return 5;
+            }
+
+          type = credential_type (credential);
+          if (strcmp (type, "up"))
+            {
+              sql_rollback ();
```
ozgen, line 305: free the type before return

src/manage_sql_oci_image_targets.c
```diff
@@ -0,0 +1,918 @@
...
+ * @brief Modify an OCI image target.
+ *
+ * @param[in]   oci_image_target_id  UUID of target.
+ * @param[in]   name                 Name of target.
+ * @param[in]   comment              Comment on target.
+ * @param[in]   credential_id        Credential.
+ * @param[in]   image_references     List of image urls.
+ * @param[out]  error_message        Error message if any.
+ *
+ * @return 0 success, 1 failed to find target,
```
ozgen, line 194: I personally find enums more readable than integer return codes, but I’ll leave it up to you.

src/manage_sql_oci_image_targets.c
```diff
@@ -0,0 +1,918 @@
...
+  /* Remove from trash tables. */
+
+  sql ("DELETE FROM oci_image_targets_trash WHERE id = %llu;",
+       resource);
+
+  sql_commit ();
+  return 0;
+}
+
+/**
```
ozgen, line 579: can you move these macros to the `manage_sql_oci_image_targets.h` header file

src/manage_sql_oci_image_targets.h
```diff
@@ -0,0 +1,32 @@
+/* Copyright (C) 2024 Greenbone AG
```
ozgen: change the year with 2025

src/manage_oci_image_targets.c
a-h-abdelsalam: I don't think this is needed. The reason is that `rest` is already checked for NULL or empty string a few lines above and therefore `g_strsplit` should return an array of at least one element.

src/manage_oci_image_targets.c
a-h-abdelsalam: I don't think so for the same reason above, The length should be at least 1.

src/manage_oci_image_targets.c
a-h-abdelsalam: yes, this is why the if-condition on the following line checks for invalid inputs:
`if (port_num <= 0 || port_num > 65535)`

src/manage_sql_oci_image_targets.c
a-h-abdelsalam: I would prefer to keep it consistent with similar functions.

src/manage_sql_oci_image_targets.c
a-h-abdelsalam: I think you mean the filter/iterator columns. If so, I would prefer to keep the consistency with how they are defined for other resources. Check `manage_sql_agent_installers.c` for example which was recently created.

ozgen: You're right, I wasn’t aware of the recent changes and the consistency in the new modules like `manage_sql_agent_installers.c`. Thanks for pointing that out, and sorry for the confusion.

src/manage_oci_image_targets.c
ozgen: That make sense, thanks for the clarification.

src/gmp_oci_image_targets.c
a-h-abdelsalam: I removed the error message from the copy. I don't think it is needed since there is no validation of refs

src/manage_sql_oci_image_targets.h
```diff
@@ -0,0 +1,32 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @file manage_sql_report_configs.h
```
ozgen, line 20: can you change the command also its still `manage_sql_report_configs`

ozgen (dismissed):

PR #2487, opened 2025-07-17
Add: Add create, modify, get, and delete support for agent groups

ozgen:
## What

This PR implements full GMP command support for agent group management. Specifically, it adds the following commands:

- create_agent_group
- modify_agent_group
- get_agent_groups
- delete_agent_group

## Why

These changes add the backend support needed to use agent groups in scans, making it easier to manage and reuse sets of agents.

## References

GEA-976

### DIFF
--- a/src/gmp_agent_groups.c
+++ b/src/gmp_agent_groups.c
@@ -96,6 +96,9 @@ get_agent_groups_run (gmp_parser_t *gmp_parser, GError **error)
 
   while (1)
     {
+      scanner_t scanner;
+      char *agent_scanner_name, *agent_scanner_uuid;
+
       ret = get_next (&agent_groups, &get_agent_groups_data.get, &first, &count,
                       init_agent_group_iterator);
       if (ret == 1)
@@ -107,6 +110,10 @@ get_agent_groups_run (gmp_parser_t *gmp_parser, GError **error)
           return;
         }
 
+      scanner = agent_group_iterator_scanner (&agent_groups);
+      agent_scanner_uuid = scanner_uuid (scanner);
+      agent_scanner_name = scanner_name (scanner);
+
       // Start <agent_group>
       SEND_GET_COMMON (agent_group,
                        &get_agent_groups_data.get,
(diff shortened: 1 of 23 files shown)


src/gmp_agent_groups.c
```diff
@@ -0,0 +1,713 @@
...
+      // Start <agent_group>
+      SEND_GET_COMMON (agent_group,
+                       &get_agent_groups_data.get,
+                       &agent_groups);
+
+      SENDF_TO_CLIENT_OR_FAIL ("<scanner id=\"%s\">"
+                               "<name>%s</name>"
+                               "</scanner>",
+                               scanner_uuid (agent_group_iterator_scanner (&agent_groups)),
+                               scanner_name (agent_group_iterator_scanner (&agent_groups)));
```
timopollmeier: The strings from `scanner_uuid` and `scanner_name` have to be freed.
It may be a good idea to include the name and UUID in the iterator columns so they are freed by the iterator functions and you could even add them to the filter columns.

src/gmp_agent_groups.c
```diff
@@ -0,0 +1,713 @@
...
+
+      gchar *agents = agent_group_agents_string (get_iterator_resource (&agent_groups));
+      SENDF_TO_CLIENT_OR_FAIL ("<agents>%s</agents>", agents ? agents : "");
```
timopollmeier: I would make this a list of elements like this so the front-end could link to the agents by UUID:
```xml
<agent id="..."><name>...</name></agent>
```
and I would also only include it with the details.

Maybe there should also be an `<agent_count>` element which is always included, whether details are requested or not. That one may also be a filterable column.

src/manage_sql_agent_groups.c
```diff
@@ -0,0 +1,639 @@
...
+
+  if (!agent_uuids || agent_uuids->count == 0)
+    return AGENT_GROUP_RESP_NO_AGENTS_PROVIDED;
+
+  // GET scanner ID from the first agent
+  scanner_t scanner = 0;
+  int ret = get_scanner_from_agent_uuid (agent_uuids->agent_uuids[0], &scanner);
+  agent_group_resp_t map_response = map_get_scanner_result_to_agent_group_resp (ret);
+  if (map_response != AGENT_GROUP_RESP_SUCCESS)
+    return map_response;
```
timopollmeier, line 183: I think you should also add a check if the user has the permission to access the scanner.

src/manage_sql_agent_groups.c
```diff
@@ -0,0 +1,639 @@
...
+
+  scanner_t scanner = 0;
+  int ret = get_scanner_by_agent_group_id (agent_group, &scanner);
+  if (ret == -1)
+    {
+      db_copy_buffer_cleanup (&buffer);
+      sql_rollback ();
+      return AGENT_GROUP_RESP_SCANNER_NOT_FOUND;
+    }
+  group_data->scanner = scanner;
```
timopollmeier, line 324: Maybe check the scanner permissions here as well.

src/manage_sql_agent_groups.c
```diff
@@ -0,0 +1,639 @@
...
+    {
+      const gchar *uuid = agent_uuids->agent_uuids[i];
+      agent_t agent_id = agent_id_by_uuid_and_scanner (uuid, group_data->scanner);
+
+      if (agent_id == 0)
+        {
+          db_copy_buffer_cleanup (&buffer);
+          sql_rollback ();
+          return AGENT_GROUP_RESP_AGENT_SCANNER_MISMATCH;
+        }
```
timopollmeier: This does not seem to distinguish between a scanner mismatch and the agent not existing at all.

src/manage_sql_agent_groups.c
```diff
@@ -0,0 +1,639 @@
...
+  for (int i = 0; i < agent_uuids->count; ++i)
+    {
+    const gchar *uuid = agent_uuids->agent_uuids[i];
+    agent_t agent_id = agent_id_by_uuid_and_scanner (uuid, group_data->scanner);
+    if (agent_id == 0)
+    {
+      db_copy_buffer_cleanup (&buffer);
+      sql_rollback ();
+      return AGENT_GROUP_RESP_AGENT_SCANNER_MISMATCH;
+    }
```
timopollmeier: As above, this does not seem to have a check for the agent not existing at all.

timopollmeier (approved):

PR #2488, opened 2025-07-23
Add: add support to check if agents are in use

ozgen:
## What

Adds support for checking whether agents are currently in use by any agent group, including those in the trash table.

## Why

This change is necessary to prevent accidental deletion or modification of agents that are still referenced in agent groups. 

## References

GEA-968

### DIFF
--- a/src/gmp_agents.c
+++ b/src/gmp_agents.c
@@ -445,7 +445,7 @@ modify_agents_run (gmp_parser_t *gmp_parser, GError **error)
         SEND_TO_CLIENT_OR_FAIL (XML_INTERNAL_ERROR ("modify_agents"));
         log_event_fail ("agents", "Agents", NULL, "modified");
         break;
-
+      case AGENT_RESPONSE_IN_USE_ERROR:
       default:
         SEND_TO_CLIENT_OR_FAIL (XML_INTERNAL_ERROR ("modify_agents"));
         log_event_fail ("agents", "Agents", NULL, "modified");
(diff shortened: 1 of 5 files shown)


src/manage_agents.h
```diff
@@ -106,7 +109,8 @@ typedef enum {
   AGENT_RESPONSE_INVALID_ARGUMENT = -8,             ///< Failed invalid argument
   AGENT_RESPONSE_INVALID_AGENT_OWNER = -9,          ///< Failed getting owner UUID
   AGENT_RESPONSE_AGENT_NOT_FOUND = -10,             ///< Failed getting owner UUID
-  AGENT_RESPONSE_INTERNAL_ERROR = -11               ///< Internal error
+  AGENT_RESPONSE_INTERNAL_ERROR = -11,              ///< Internal error
+  AGENT_RESPONSE_IN_USE_ERROR = -12                 ///< Agent is in used by an Agent Group
```
a-h-abdelsalam:
```suggestion
  AGENT_RESPONSE_IN_USE_ERROR = -12                 ///< Agent is used by an Agent Group
```

src/manage_sql_agents.c
```diff
@@ -577,8 +577,14 @@ agent_writable (agent_t agent)
 int
 agent_in_use (agent_t agent)
```
a-h-abdelsalam, line 578: The header for this function could be updated

src/gmp_agents.c
```diff
@@ -445,7 +445,7 @@ modify_agents_run (gmp_parser_t *gmp_parser, GError **error)
         SEND_TO_CLIENT_OR_FAIL (XML_INTERNAL_ERROR ("modify_agents"));
         log_event_fail ("agents", "Agents", NULL, "modified");
         break;
-
+      case AGENT_RESPONSE_IN_USE_ERROR:
```
a-h-abdelsalam, line 448: Do you want to specify here the reason for failure like in the delete case ?

src/manage_sql_agents.c
ozgen: this func is already in the manage_agents.h ref: https://github.com/greenbone/gvmd/blob/main/src/manage_agents.h#L193

a-h-abdelsalam: Sorry for the confusion. By "header" I meant the description:
`@return Always returns 0 (not in use).`

src/gmp_agents.c
ozgen: Since `modify_agents` never returned `AGENT_RESPONSE_IN_USE_ERROR`, I added the `AGENT_RESPONSE_IN_USE_ERROR` case to `modify_agents` to ensure all enum values are covered consistently.

a-h-abdelsalam (approved):

PR #2493, opened 2025-07-28
Add: add support for creating tasks with agent group

ozgen:
## What

Added support for creating a task using an agent group without requiring a target.

## Why

To allow agent-based scans to be configured directly via agent groups, enabling more flexible task creation.

## References

GEA-977


src/gmp.c
```diff
@@ -24109,14 +24135,61 @@ gmp_xml_handle_end_element (/* unused */ GMarkupParseContext* context,
 
           /* Check for the right combination of target and config. */
 
-          if (create_task_data->target_id == NULL)
+          if (create_task_data->target_id == NULL && create_task_data->agent_group_id == NULL)
             {
               SEND_TO_CLIENT_OR_FAIL
                (XML_ERROR_SYNTAX ("create_task",
                                   "A target is required"));
```
timopollmeier, line 24142:
```suggestion
                                  "A target or agent group is required"));
```

src/gmp.c
```diff
@@ -24109,14 +24135,61 @@ gmp_xml_handle_end_element (/* unused */ GMarkupParseContext* context,
...
+                }
+
+              SENDF_TO_CLIENT_OR_FAIL (XML_OK_CREATED_ID ("create_task"),
+                                       tsk_uuid);
+              make_task_complete (create_task_data->task);
+              log_event ("task", "Task", tsk_uuid, "created");
+              g_free (tsk_uuid);
+              create_task_data_reset (create_task_data);
+              set_client_state (CLIENT_AUTHENTIC);
+              break;
```
timopollmeier: I think the case where an agent_group id is given should only handle checking and setting that agent_group_id and agent_group_location and not end the response early.
Instead you could add an `else` as a line just before the `#endif /* ENABLE_AGENTS */`.

src/manage.h
```diff
@@ -606,6 +606,21 @@ task_target_in_trash (task_t);
 void
 set_task_target (task_t, target_t);
 
+#if ENABLE_AGENTS
+void
+set_task_agent_group (task_t task, agent_group_t agent_group);
+
+void
+set_task_agent_group_location (task_t task);
```
timopollmeier: If you don't use `set_task_agent_group_location` separately from `set_task_agent_group`, maybe you should combine the two.

src/manage_sql.c
```diff
@@ -32518,6 +32658,8 @@ delete_scanner (const char *scanner_id, int ultimate)
     }
 
 #if ENABLE_AGENTS
+   // Delete agent group tasks related to the scanner.
+  delete_agent_group_tasks_by_scanner (scanner);
```
timopollmeier: I don't think tasks should be deleted if their assigned scanners are deleted.

As it is now, moving the scanner to the trashcan should just ensure tasks, agents and agent groups correctly identify the scanner as being in the trash, e.g. by updating the scanner_location field. (This would also mean change the previous behavior of deleting the agents and agent groups in this case.)
Fully deleting the scanner would be blocked if there are still any tasks using it. If there are no tasks deleting the agents and agent groups together with fully deleting the scanner would be fine.

src/manage_sql_agent_groups.c
```diff
@@ -436,9 +436,26 @@ delete_agent_group (const char *agent_group_uuid, int ultimate)
...
+         "     agent_group_location = " G_STRINGIFY (LOCATION_TRASH)
+         " WHERE agent_group = %llu"
+         " AND agent_group_location = " G_STRINGIFY (LOCATION_TABLE) ";",
+         trash_id,
+         agent_group);
+
       permissions_set_locations ("agent_group", agent_group, trash_id, LOCATION_TRASH);
       tags_set_locations ("agent_group", agent_group, trash_id, LOCATION_TRASH);
     }
+  else if (sql_int ("SELECT count(*) FROM tasks"
```
timopollmeier: I think this should use the `agent_group_in_use` function to ensure the indicator that the agent group is still in use is consistent with the delete command failing because the agent group is still in use.

timopollmeier (changes requested):

src/gmp.c
```diff
@@ -1093,6 +1094,7 @@ create_task_data_reset (create_task_data_t *data)
   free (data->schedule_periods);
   free (data->target_id);
   free (data->usage_type);
+  free (data -> agent_group_id);
```
mattmundell, line 1097:
```suggestion
  free (data->agent_group_id);
```

src/gmp.c
```diff
@@ -7941,6 +7959,14 @@ gmp_xml_handle_start_element (/* unused */ GMarkupParseContext* context,
                               &create_task_data->config_id);
             set_client_state (CLIENT_CREATE_TASK_CONFIG);
           }
+#if ENABLE_AGENTS
+        else if (strcasecmp ("AGENT_GROUP", element_name) == 0)
+          {
+            append_attribute (attribute_names, attribute_values, "id",
+                              &create_task_data -> agent_group_id);
```
mattmundell, line 7966:
```suggestion
                              &create_task_data->agent_group_id);
```

src/gmp.c
```diff
@@ -27087,6 +27131,21 @@ gmp_xml_handle_end_element (/* unused */ GMarkupParseContext* context,
                                   modify_task_data->task_id,
                                   "modified");
                   break;
+
+#if ENABLE_AGENTS
+              case 18 :
+                if (send_find_error_to_client
+                  ("modify_task", "agent_group",
+                   modify_task_data -> agent_group_id, gmp_parser))
```
mattmundell:
```suggestion
                   modify_task_data->agent_group_id, gmp_parser))
```

src/gmp.c
```diff
@@ -27087,6 +27131,21 @@ gmp_xml_handle_end_element (/* unused */ GMarkupParseContext* context,
...
+                  {
+                    error_send_to_client (error);
+                    return;
+                  }
+                log_event_fail ("agent_group", "Agent Group",
+                                modify_task_data -> task_id,
```
mattmundell:
```suggestion
                                modify_task_data->task_id,
```

src/gmp.c
```diff
@@ -27087,6 +27147,20 @@ gmp_xml_handle_end_element (/* unused */ GMarkupParseContext* context,
...
+#if ENABLE_AGENTS
+              case 18 :
+                if (send_find_error_to_client ("modify_task", "agent_group",
+                                               modify_task_data->agent_group_id,
+                                               gmp_parser))
+                  {
+                    error_send_to_client (error);
+                    return;
+                  }
+                log_event_fail ("agent_group", "Agent Group",
```
mattmundell: I think this should be `"task", "Task"`, because the log will say "Agent Group TASK_ID could not be modified".

src/manage_sql.c
```diff
@@ -8787,6 +8787,68 @@ set_task_target (task_t task, target_t target)
        task);
 }
 
+#if ENABLE_AGENTS
+/**
+ * @brief Set the agent group of a task.
```
timopollmeier, line 8792:
```suggestion
 * @brief Set the agent group of a task, also updating the agent group location.
```

src/manage_sql.c
```diff
@@ -32477,6 +32621,15 @@ delete_scanner (const char *scanner_id, int ultimate)
     {
       scanner_t trash_scanner;
 
+#if ENABLE_AGENTS
+      // check scanner is in used by agent_group
```
timopollmeier:
```suggestion
      // check scanner is in use by agent_group
```

src/manage_sql_agent_groups.c
```diff
@@ -509,6 +523,16 @@ restore_agent_group (const char *agent_group_uuid)
...
   tags_set_locations ("agent_group", trash_id, restored_id, LOCATION_TABLE);
 
+  /* Update the agent_group in any tasks. */
+  sql ("UPDATE tasks"
+        " SET agent_group = %llu,"
+        " agent_group_location = " G_STRINGIFY (LOCATION_TABLE)
+        " WHERE agent_group = %llu"
+        " AND agent_group_location = " G_STRINGIFY (LOCATION_TRASH),
+        restored_id,
+        trash_id);
```
timopollmeier:
```suggestion
  sql ("UPDATE tasks"
       " SET agent_group = %llu,"
       " agent_group_location = " G_STRINGIFY (LOCATION_TABLE)
       " WHERE agent_group = %llu"
       " AND agent_group_location = " G_STRINGIFY (LOCATION_TRASH),
       restored_id,
       trash_id);
```

src/manage_sql_agent_groups.c
```diff
@@ -722,4 +746,97 @@ agent_group_agent_iterator_name (iterator_t *iterator)
...
+ * @return 1 if in use, else 0.
+ */
+int
+trash_agent_group_in_use (agent_group_t agent_group)
+{
+  return !!sql_int ("SELECT count(*) FROM tasks"
+                  " WHERE agent_group = %llu"
+                  " AND agent_group_location = "
+                  G_STRINGIFY (LOCATION_TRASH),
+                  agent_group);
```
timopollmeier:
```suggestion
  return !!sql_int ("SELECT count(*) FROM tasks"
                    " WHERE agent_group = %llu"
                    " AND agent_group_location = "
                    G_STRINGIFY (LOCATION_TRASH),
                    agent_group);
```

src/manage_sql_agent_groups.c
```diff
@@ -436,9 +436,23 @@ delete_agent_group (const char *agent_group_uuid, int ultimate)
            " SELECT %llu, agent_id FROM agent_group_agents WHERE group_id = %llu;",
            trash_id, agent_group);
 
+    /* Update the location of the agent_group in any trashcan tasks. */
```
timopollmeier: I think the moving the agent group to the trash should fail if the agent group is in use by a task that is not hidden/in the trash.
At the same time, restoring a task should require that the associated agent group is not is in the trash.

src/manage_sql_agent_groups.c
```diff
@@ -436,9 +436,23 @@ delete_agent_group (const char *agent_group_uuid, int ultimate)
...
+         "     agent_group_location = " G_STRINGIFY (LOCATION_TRASH)
+         " WHERE agent_group = %llu"
+         " AND agent_group_location = " G_STRINGIFY (LOCATION_TABLE) ";",
+         trash_id,
+         agent_group);
+
       permissions_set_locations ("agent_group", agent_group, trash_id, LOCATION_TRASH);
       tags_set_locations ("agent_group", agent_group, trash_id, LOCATION_TRASH);
     }
+  else if (agent_group_in_use (agent_group))
```
timopollmeier: With an "ultimate" delete, this should also have a check if it's in use by hidden tasks (unlike the check I suggested above).
Also, you you should add a `trash_agent_group_in_use` when trying to ultimately delete an agent group that is in the trash (around line 404).

timopollmeier (changes requested):

timopollmeier (approved):

PR #2508, opened 2025-08-08
Fix: Set default SIGCHLD handler for scan handlers

timopollmeier:
## What
When forking the scan handler process in fork_scan_handler the signal handler for SIGCHLD is now set to SIG_DFL.

## Why
This fixes not being able to run sub-processes like the relay mapper with g_spawn_sync.

## References
GEA-1208

### DIFF
--- a/src/manage_scan_handler.c
+++ b/src/manage_scan_handler.c
@@ -215,9 +215,12 @@ fork_scan_handler (const char *report_id, report_t report, task_t task,
                 close (pipe_fds[1]);
                 reinit_manage_process ();
 
+                // Reset SIGCHLD handler to default so the process can
+                // use common functions to wait for its own child processes.
                 memset (&action, '\0', sizeof (action));
                 sigemptyset (&action.sa_mask);
                 action.sa_handler = SIG_DFL;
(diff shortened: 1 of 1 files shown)


src/manage_scan_handler.c
```diff
@@ -211,6 +214,19 @@ fork_scan_handler (const char *report_id, report_t report, task_t task,
                 // Grandchild
                 close (pipe_fds[1]);
                 reinit_manage_process ();
+
```
ozgen, line 217: We may need a commend for explanation:
`// Reset SIGCHLD so g_spawn_sync() can wait() its child (parent may have SIGCHLD=SIG_IGN).
`

src/manage_scan_handler.c
```diff
@@ -211,6 +214,19 @@ fork_scan_handler (const char *report_id, report_t report, task_t task,
...
+                memset (&action, '\0', sizeof (action));
+                sigemptyset (&action.sa_mask);
+                action.sa_handler = SIG_DFL;
```
ozgen, line 222: maybe we can also set flags explicitly for readability.

`action.sa_flags = 0;`

PR #2510, opened 2025-08-12
Removed the gvmd --optimize migrate-relay-sensors option.

jhelmold:
## What
The gvmd --optimize option "migrate-relay-sensors" has been removed. 

## Why
This functionality is not needed any more and is obsolete.

## References
GEA-388


src/manage.c
```diff
@@ -2339,28 +2334,6 @@ set_relay_mapper_path (const char *new_path)
   relay_mapper_path = new_path ? g_strdup (new_path) : NULL;
 }
 
-/**
- * @brief Gets whether to migrate sensors if relays do not match.
- *
- * @return Whether to migrate sensors if relays do not match.
- */
-int
-get_relay_migrate_sensors ()
```
mattmundell, line 2348: These two need to be removed from `manage.h` too.

PR #2511, opened 2025-08-12
Fix: Use report configs in alerts

timopollmeier:
## What
When generating reports for alerts, the method data setting a report config now actually used instead of always falling back to the default report format parameters.

## Why
The alerts are meant to use the set report configs.

## References
GEA-1155


mattmundell (approved):
Looks OK to me. Works with a small test too.

PR #2522, opened 2025-08-21
Add scan agent config support

ozgen:
## What

Extend agent_data to include the Agent Controller’s structured config and add modify_agent_control_scan_config; update GMP docs and <modify_agents> to accept <config>.

## Why

Align with the Agent Controller model (deprecating old interval fields) to make the configuration clearer

## References

GEA-1126


src/gmp_agent_control_scan_agent_config.c
```diff
@@ -0,0 +1,366 @@
...
+      e = entity_child (hb, "interval_in_seconds");
+      if (e)
+        cfg->heartbeat.interval_in_seconds =
+          atoi (entity_text (e) ? entity_text (e) : "0");
+
+      e = entity_child (hb, "miss_until_inactive");
+      if (e)
+        cfg->heartbeat.miss_until_inactive =
+          atoi (entity_text (e) ? entity_text (e) : "0");
+    }
```
timopollmeier: Couldn't this be moved to a separate function that can be reused for the modify_agent command?

src/gmp_agents.c
```diff
@@ -352,113 +447,249 @@ modify_agents_run (gmp_parser_t *gmp_parser, GError **error)
...
+              cfg->heartbeat.interval_in_seconds = atoi (t ? t : "0");
+            }
+          /* miss_until_inactive */
+          e = entity_child (hb, "miss_until_inactive");
+          if (e)
+            {
+              const char *t = entity_text (e);
+              cfg->heartbeat.miss_until_inactive = atoi (t ? t : "0");
+            }
+        }
```
timopollmeier: This is the part I that could I imagine could be replaced by the function I suggested above.

src/manage_agent_common.h
```diff
@@ -15,24 +15,49 @@
...
+gvmd_agent_connector_free (gvmd_agent_connector_t conn);
+
 agent_uuid_list_t
 agent_uuid_list_new (int count);
 
 void
 agent_uuid_list_free (agent_uuid_list_t uuid_list);
 
+gchar *
+concat_error_messages (const GPtrArray *errors, const gchar *sep);
```
timopollmeier: I think this is generic enough to move to utils.h

src/manage_agent_common.c
```diff
@@ -46,14 +49,133 @@ agent_uuid_list_new (int count)
...
+ * @brief Join an array of gchar* with a separator, skipping NULL/empty entries.
+ *
+ * @param[in]  arr  GPtrArray of gchar* (may be NULL).
+ * @param[in]  sep  Separator string, default to "; " if NULL.
+ *
+ * @return Newly allocated joined string, or NULL if no non-empty entries.
+ *         Caller must g_free().
+ */
+gchar *
+concat_error_messages (const GPtrArray *errors, const gchar *sep)
```
timopollmeier: I think this is generic enough to move to utils.c and maybe you could add a prefix parameter (see my comment about the use of this function below).

src/gmp_agent_control_scan_agent_config.c
```diff
@@ -0,0 +1,366 @@
...
+      SEND_TO_CLIENT_OR_FAIL (
+        XML_ERROR_UNAVAILABLE ("modify_agent_control_scan_config",
+          "Could not connect to Agent-Controller"));
+      log_event_fail ("agent_control_scan_config", "Agent Control Scan Config",
+                      scanner_uuid, "modified");
+      modify_agent_control_scan_config_reset ();
+      return;
+
+    case -3:
+      gchar *status_text = concat_error_messages (errs, "; ");
```
timopollmeier: I think you should add something like "Validation failed for <config>: " before the list of errors.

PR #2525, opened 2025-08-25
Add: Ability to get, modify and verify scanners of type 'container-image'

robindittmar:
## What

* Modified --get-scanners to support scanner type 'container-image'
* Modifed --modify-scanner to support scanner type 'container-image'
* Modified --verify-scanner to support scanner type 'container-image'
* Added automatic creation of default 'container-image' scanner

## Why

Required to support container image scanners

## References

GEA-1228


src/gvmd.c
```diff
@@ -3216,6 +3216,10 @@ gvmd (int argc, char** argv, char *env[])
             type = SCANNER_TYPE_OPENVASD;
           else if (!strcasecmp (scanner_type, "openvasd-sensor"))
             type = SCANNER_TYPE_OPENVASD_SENSOR;
+#ifdef ENABLE_CONTAINER_SCANNING
```
mattmundell, line 3219:
```suggestion
#if ENABLE_CONTAINER_SCANNING
```
`../CMakeList.txt` ensures that `ENABLE_CONTAINER_SCANNING` is always defined, so this will always be true with `#ifdef`.

PR #2546, opened 2025-09-11
Add: support agent group task scans via HTTP scanner API

ozgen:
## What

Adds support for creating scans for agent group tasks using the HTTP scanner API.

## Why

This is needed to allow gvmd to manage agent-based scans consistently with agent controller.

## References

GEA-1258

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -28037,15 +28037,6 @@ gmp_xml_handle_end_element (/* unused */ GMarkupParseContext* context,
                                   resume_task_data->task_id,
                                   "resumed");
                   break;
-                case -10:
-                  SEND_TO_CLIENT_OR_FAIL
-                  (XML_ERROR_SYNTAX ("resume_task",
-                    "Agent task is not support"
-                    " this process"));
-                  log_event_fail ("task", "Task",
-                                  resume_task_data->task_id,
-                                  "resumed");
-                    break;
                 default: /* Programming error. */
                   assert (0);
                   SEND_TO_CLIENT_OR_FAIL (XML_INTERNAL_ERROR ("resume_task"));
--- a/src/manage.c
+++ b/src/manage.c
@@ -2609,7 +2609,6 @@ run_agent_control_task (task_t task, char **report_id);
  *         -4 target missing hosts,
  *         -6 already a task running in this process,
  *         -9 fork failed.
- *         -10 not supported
  */
 static int
 run_task (const char *task_id, char **report_id, int from)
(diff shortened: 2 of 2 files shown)


src/gmp.c
```diff
@@ -28037,6 +28037,15 @@ gmp_xml_handle_end_element (/* unused */ GMarkupParseContext* context,
...
                   break;
+                case -10:
+                  SEND_TO_CLIENT_OR_FAIL
+                  (XML_ERROR_SYNTAX ("resume_task",
+                    "Agent task is not support"
+                    " this process"));
+                  log_event_fail ("task", "Task",
+                                  resume_task_data->task_id,
+                                  "resumed");
+                    break;
```
timopollmeier: I think you should just re-use the existing return code 4 for the case that resuming is not supported.

src/manage.c
```diff
@@ -2650,6 +2657,17 @@ run_task (const char *task_id, char **report_id, int from)
     return run_openvasd_task (task, from, report_id);
 #endif
 
+#if ENABLE_AGENTS
+  if (scanner_type (scanner) == SCANNER_TYPE_AGENT_CONTROLLER
+    || scanner_type (scanner) == SCANNER_TYPE_AGENT_CONTROLLER_SENSOR)
+    {
+      if (from == 1)
+        // Resume task is not supported by agent controller
+        return -10;
```
timopollmeier:
```suggestion
        return 4;
```

src/manage.c
```diff
@@ -2603,6 +2609,7 @@ run_openvasd_task (task_t task, int from, char **report_id);
  *         -4 target missing hosts,
  *         -6 already a task running in this process,
  *         -9 fork failed.
+ *         -10 not supported
```
timopollmeier:
```suggestion
```
There is already return code 4 for that.

src/manage.c
```diff
@@ -2839,6 +2857,7 @@ stop_task (const char *task_id)
...
```
timopollmeier:
```suggestion
```

src/manage.c
```diff
@@ -7726,7 +7745,7 @@ handle_openvasd_scan (task_t task, report_t report, const char *scan_id)
 
               result_start += g_slist_length (results);
 
-              parse_openvasd_report (task, report, results, start_time,
+              parse_http_scanner_report (task, report, results, start_time,
                                      end_time);
```
timopollmeier:
```suggestion
              parse_http_scanner_report (task, report, results, start_time,
                                         end_time);
```

PR #2553, opened 2025-09-15
Add: Extend get_aggregates to support Agents, Agent Groups, and Agent Installers.  Add --no-default-certs option to scanner creation to bypass

ozgen:
## What
- Added `--no-default-certs` option to scanner creation (`manage_create_scanner`)  
   allows bypassing certificate file validation and skipping autogenerated credential creation.  
- Extended `get_aggregates` support to include **Agents**, **Agent Groups**, and **Agent Installers**.  

## Why
The new `--no-default-certs` option provides flexibility for use cases where the scanner does not need certs.
Extending `get_aggregates` makes agent-related entities first-class citizens in the API, enabling consistent filtering and aggregation similar to other resources. 

## References

GEA-1275


src/manage_sql.c
```diff
@@ -31295,107 +31300,116 @@ manage_create_scanner (GSList *log_config, const db_conn_info_t *database,
...
   ret = create_scanner (name, NULL, host, port, type, &scanner, ca_pub,
                         used_credential_id, relay_host, relay_port);
-  g_free (ca_pub);
-  g_free (key_pub);
-  g_free (key_priv);
-  g_free (used_credential_id);
+  if (ca_pub) g_free (ca_pub);
+  if (key_pub) g_free (key_pub);
+  if (key_priv) g_free (key_priv);
+  if (used_credential_id) g_free (used_credential_id);
```
timopollmeier: It's safe to use g_free with NULL pointers, so this doesn't have to be changed.

src/gvmd.c
```diff
@@ -2521,6 +2522,10 @@ gvmd (int argc, char** argv, char *env[])
           " Either 'OpenVAS', 'OSP', 'OSP-Sensor'"
           " or a number as used in GMP.",
           "<scanner-type>" },
+          { "no-default-certs", '\0', 0, G_OPTION_ARG_NONE,
+          &scanner_no_default_certs,
```
timopollmeier:
```suggestion
        { "no-default-certs", '\0', 0, G_OPTION_ARG_NONE,
          &scanner_no_default_certs,
```

PR #2558, opened 2025-09-17
Fix: use modify_scanner permission for modify_agent_control_scan_agent_config

ozgen:
## What

Use modify_scanner permission for modify_agent_control_scan_agent_config

## Why

Simply the permissions for modify_agent_control_scan_agent_config.

## References

GEA-1278

### DIFF
--- a/src/gmp_agent_control_scan_agent_config.c
+++ b/src/gmp_agent_control_scan_agent_config.c
@@ -10,8 +10,8 @@
  */
 
 #if ENABLE_AGENTS
-
 #include "gmp_agent_control_scan_agent_config.h"
+#include "manage_acl.h"
 #include "manage_agent_control_scan_config.h"
 #include "manage_sql.h"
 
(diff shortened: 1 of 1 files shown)


timopollmeier (approved):

src/gmp_agent_control_scan_agent_config.c
```diff
@@ -114,6 +114,14 @@ modify_agent_control_scan_config_run (gmp_parser_t *gmp_parser, GError **error)
 
   scanner_uuid = entity_attribute (root, "agent_control_id");
 
+  if (!acl_user_may ("modify_scanner"))
```
mattmundell, line 117: I think the `find_scanner_with_permission` below should also use `"modify_scanner"`, instead of using `"get_scanners"`.

If you need `modify_scanner` permission on the scanner to modify an agent, then you should have permission to modify that particular scanner.

For example, I could share a scanner with you but only give you read access to the scanner, and you would still be able to modify the agent.

mattmundell: Compare with `modify_alert` in `manage_sql_alerts.c`, which does it that way.

PR #2576, opened 2025-10-01
Add: Include <agent_group> XML element in get_tasks responses

ozgen:
## What

Added support for including agent group information in the `get_tasks` response.  
When the feature flag `ENABLE_AGENTS` is enabled,

## Why

This change improves API completeness and aligns agent groups with other task resources.

## References

GEA-1278

### DIFF
--- a/src/manage_sql_agent_groups.c
+++ b/src/manage_sql_agent_groups.c
@@ -859,6 +859,54 @@ trash_agent_group_in_use (agent_group_t agent_group)
 }
 
 /**
+ * @brief Return whether an agent_group is readable.
+ *
+ * @param[in]  agent_group  Row id in agent_groups table.
+ *
+ * @return 1 if readable, 0 otherwise.
+ */
+int
+agent_group_readable (agent_group_t agent_group)
+{
+  char *uuid;
+  agent_group_t found = 0;
+
+  if (agent_group == 0)
+    return 0;
+  uuid = agent_group_uuid (agent_group);
+  if (uuid == NULL)
+    return 0;
+  find_agent_group_with_permission (uuid, &found, "get_agent_groups");
+  g_free (uuid);
+  return found > 0;;
+}
+
+/**
+ * @brief Return whether a trashcan agent_group is readable.
+ *
+ * @param[in]  agent_group  Row id in agent_groups_trash.
+ *
+ * @return 1 if readable, 0 otherwise.
+ */
+int
+trash_agent_group_readable (agent_group_t agent_group)
+{
+  char *uuid;
+  agent_group_t found = 0;
+
+  if (agent_group == 0)
+    return 0;
+  uuid = trash_agent_group_uuid (agent_group);
+  if (find_trash ("agent_group", uuid, &found))
+    {
+      g_free (uuid);
+      return 0;
+    }
+  g_free (uuid);
+  return found > 0;
+}
+
+/**
  * @brief Return whether a agent_group is writable.
  *
  * @param[in]  agent_group  Agent Group row id.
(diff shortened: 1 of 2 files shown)


src/gmp.c
```diff
@@ -19500,6 +19501,69 @@ handle_get_tasks (gmp_parser_t *gmp_parser, GError **error)
...
+#if ENABLE_AGENTS
+          agent_group_t agent_group;
+          int agent_group_in_trash, agent_group_available;
+          char *task_agent_group_uuid = NULL, *task_agent_group_name = NULL;
+          gchar *task_agent_group_name_escaped = NULL;
+
+          agent_group = task_agent_group (index); /* 0 if none */
+          agent_group_in_trash = task_agent_group_in_trash (index);
+
+          if (agent_group || agent_group_in_trash)
```
timopollmeier: I would only check if the agent group is set here in case the agent group was removed from the task but the trash flag was not set to 0.
```suggestion
          if (agent_group)
```

src/gmp.c
```diff
@@ -19500,6 +19501,69 @@ handle_get_tasks (gmp_parser_t *gmp_parser, GError **error)
...
+                    "%s: GET_TASKS: error finding task agent group, aborting",
+                    __func__);
+
+                  agent_group_available = (found > 0);
+                }
+              else
+                {
+                  task_agent_group_uuid = NULL;
+                  task_agent_group_name = NULL;
+                }
```
timopollmeier: This would also be obsolete with the change of the outer condition.
```suggestion
```

src/gmp.c
```diff
@@ -19500,6 +19501,69 @@ handle_get_tasks (gmp_parser_t *gmp_parser, GError **error)
...
+                "%s"
+                "</agent_group>",
+                task_agent_group_uuid ? : "",
+                task_agent_group_name_escaped ? : "",
+                agent_group_in_trash,
+                agent_group_available ? "" : "<permissions/>");
+
+              free (task_agent_group_name);
+              free (task_agent_group_uuid);
+              g_free (task_agent_group_name_escaped);
```
timopollmeier: If you use `g_markup_printf_escaped` instead of `g_strdup_printf` you don't have to XML escape the name yourself.

src/gmp.c
```diff
@@ -19500,6 +19501,69 @@ handle_get_tasks (gmp_parser_t *gmp_parser, GError **error)
...
+                  task_agent_group_name = agent_group_name (agent_group);
+
+                  if (find_agent_group_with_permission (task_agent_group_uuid,
+                    &found,
+                    "get_agent_groups"))
+                    g_error (
+                    "%s: GET_TASKS: error finding task agent group, aborting",
+                    __func__);
+
+                  agent_group_available = (found > 0);
```
timopollmeier: I think this should be replaced with a `agent_group_readable` function analogous to the `trash_agent_group_readable` above.

src/gmp.c
```diff
@@ -19500,6 +19501,69 @@ handle_get_tasks (gmp_parser_t *gmp_parser, GError **error)
...
+              agent_group_available = 1;
+
+              if (agent_group_in_trash)
+                {
+                  task_agent_group_uuid = trash_agent_group_uuid (agent_group);
+                  task_agent_group_name = trash_agent_group_name (agent_group);
+                  agent_group_available = trash_agent_group_readable (
+                    agent_group);
+                }
+              else if (agent_group)
```
timopollmeier: The `if` part is redundant if you just check for the agent_group above.
```suggestion
              else
```

src/manage_sql_agent_groups.c
```diff
@@ -883,4 +898,63 @@ agent_group_scanner (agent_group_t agent_group)
...
+ */
+int
+trash_agent_group_readable (agent_group_t agent_group)
+{
+  char *uuid;
+  agent_group_t found = 0;
+
+  if (agent_group == 0)
+    return 0;
+  uuid = agent_group_uuid (agent_group);
```
timopollmeier:
```suggestion
  uuid = trash_agent_group_uuid (agent_group);
```

src/gmp.c
```diff
@@ -19500,6 +19501,47 @@ handle_get_tasks (gmp_parser_t *gmp_parser, GError **error)
...
+                  task_agent_group_uuid = trash_agent_group_uuid (agent_group);
+                  task_agent_group_name = trash_agent_group_name (agent_group);
+                  agent_group_readable = trash_agent_group_readable (
+                    agent_group);
+                }
+              else
+                {
+                  task_agent_group_uuid = agent_group_uuid (agent_group);
+                  task_agent_group_name = agent_group_name (agent_group);
+                  agent_group_readable = trash_agent_group_readable (
```
timopollmeier: This has to be its own function `agent_group_readable` (which also means you'll to have to rename the variable) because `trash_agent_group_readable` only checks the trash table.

PR #2578, opened 2025-10-02
Fix: Tests: always free suite

mattmundell:
## What

Always free the suite in the tests.

## Why

Makes the output clearer when running tests with `-fsanitize=address`.

## References

Like https://github.com/greenbone/gvm-libs/pull/939.


mattmundell:
I've updated the `gvm-libs` required version in gvmd/pull/2579, for the test to pass. But the CI images need to be updated too.

PR #2581, opened 2025-10-08
Add: Extends the task element in get_reports XML output to include agent_group block

ozgen:
## What

Add an <agent_group> block under each <task> when compiled with ENABLE_AGENTS and when the task has an associated agent group.

## Why

Expose agent group metadata in reports so GSA does not need an extra round trip to resolve the task’s agent group.

## References

GEA-1278

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -18265,23 +18265,62 @@ print_report_xml_start (report_t report, report_t delta, task_t task,
              "<trash>%i</trash>"
              "<name>%s</name>"
              "<comment>%s</comment>"
-             "</target>"
-             "<progress>%s</progress>",
+             "</target>",
              tsk_uuid,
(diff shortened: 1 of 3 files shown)


src/manage_sql.c
```diff
@@ -18265,24 +18265,63 @@ print_report_xml_start (report_t report, report_t delta, task_t task,
...
+                 ag_uuid ? ag_uuid : "",
+                 in_trash,
+                 ag_name ? ag_name : "",
+                 ag_comment ? ag_comment : "");
+
+          g_free (ag_uuid);
+          g_free (ag_name);
+          g_free (ag_comment);
+
+          progress_xml = g_strdup_printf ("%i", 100);
```
mattmundell, line 18318: `progress_xml` has been allocated above, so it needs to be freed before being assigned again.

PR #2599, opened 2025-10-27
Add new credential store credential type

a-h-abdelsalam:
## What
New credential types cs_* are now supported in create_credential, modify_credential and get_credentials commands.

## Why
This will be used for credentials retrieved from a credential store

Requires https://github.com/greenbone/gvmd/pull/2582 to be merged

## References
GEA-1324

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -3017,15 +3017,15 @@ credential_full_type (const char* abbreviation)
   else if (strcasecmp (abbreviation, "cc") == 0)
     return "client certificate";
 #if ENABLE_CREDENTIAL_STORES
-  else if (strcmp (abbreviation, "cs_krb5") ==0)
+  else if (strcasecmp (abbreviation, "cs_krb5") ==0)
     return "Credential store Kerberos 5";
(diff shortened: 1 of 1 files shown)


src/manage.c
```diff
@@ -3016,6 +3016,18 @@ credential_full_type (const char* abbreviation)
     return NULL;
   else if (strcasecmp (abbreviation, "cc") == 0)
     return "client certificate";
+#if ENABLE_CREDENTIAL_STORES
+  else if (strcmp (abbreviation, "cs_krb5") ==0)
```
mattmundell: Is there a reason to be case sensitive here? The others use `strcasecmp`.

mattmundell (changes requested):

src/manage.c
a-h-abdelsalam: Fixed now.

mattmundell (approved):
Looks OK. Might have been good to prevent the repetition in `CREDENTIAL_ITERATOR_COLUMNS`.

PR #2600, opened 2025-10-28
Add: Added second logic to get scanner preferences from the scanner.

jhelmold:
## What
The logic to get scanner preferences from the container image scanner has now also been added to the function
"add_container_image_scan_preferences (..)", because this function is also called, when working with the openvas scanner instead of the openvasd scanner.

## Why
This logic has to work also, when working with the openvas scanner instead of the openvasd scanner.

## References
GEA-1316

## Checklist

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -37,6 +37,10 @@ jobs:
               name: container scanning
               prefix: container-scanning
               build-args: "FEATURE_TOGGLE=-DOPENVASD=1 -DENABLE_CONTAINER_SCANNING=1"
+          - build:
+              name: credential stores
+              prefix: credential-stores
+              build-args: "FEATURE_TOGGLE= -DENABLE_CREDENTIAL_STORES=1"
     name: Build and Push Container Images (${{ matrix.build.name }})
     uses: greenbone/workflows/.github/workflows/container-build-push-gea.yml@main
     with:
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -27,17 +27,17 @@ find_package(Threads)
 ## might occur.
 
 pkg_check_modules(CJSON REQUIRED libcjson>=1.7.14)
-pkg_check_modules(LIBGVM_BASE REQUIRED libgvm_base>=22.29)
-pkg_check_modules(LIBGVM_UTIL REQUIRED libgvm_util>=22.29)
-pkg_check_modules(LIBGVM_OSP REQUIRED libgvm_osp>=22.29)
-pkg_check_modules(LIBGVM_GMP REQUIRED libgvm_gmp>=22.29)
+pkg_check_modules(LIBGVM_BASE REQUIRED libgvm_base>=22.30)
+pkg_check_modules(LIBGVM_UTIL REQUIRED libgvm_util>=22.30)
+pkg_check_modules(LIBGVM_OSP REQUIRED libgvm_osp>=22.30)
+pkg_check_modules(LIBGVM_GMP REQUIRED libgvm_gmp>=22.30)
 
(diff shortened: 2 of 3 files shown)


src/manage_container_image_scanner.c
```diff
@@ -298,6 +298,40 @@ get_bool_string (const char *preference_value)
...
+  char *value = NULL;
+  GSList *point = scan_prefs;
+
+  while (point)
+    {
+      http_scanner_param_t *param;
+      char *p_name;
+
+      param = point->data;
+      p_name = g_strdup (http_scanner_param_id (param));
```
mattmundell: Wondering why this needs to be duplicated.

src/manage_container_image_scanner.c
```diff
@@ -323,6 +360,25 @@ add_container_image_scan_preferences (GHashTable *scanner_options, task_t task)
...
                          get_bool_string (registry_allow_insecure));
 
+  http_scanner_parsed_scans_preferences (connector, &scan_prefs);
+
+  if (scan_prefs && !accept_invalid_certs)
+    {
+      accept_invalid_certs = get_preference_from_list ("accept_invalid_certs", scan_prefs);
+      if (accept_invalid_certs)
+        g_hash_table_insert (scanner_options, g_strdup ("accept_invalid_certs"),
+                             get_bool_string(accept_invalid_certs));
```
mattmundell:
```suggestion
                             get_bool_string (accept_invalid_certs));
```

src/manage_container_image_scanner.c
```diff
@@ -323,6 +360,25 @@ add_container_image_scan_preferences (GHashTable *scanner_options, task_t task)
...
+    }
+
+  if (scan_prefs && !registry_allow_insecure)
+    {
+      registry_allow_insecure = get_preference_from_list ("registry_allow_insecure", scan_prefs);
+      if (registry_allow_insecure)
+        g_hash_table_insert (scanner_options, g_strdup ("registry_allow_insecure"),
+                             get_bool_string(registry_allow_insecure));
```
mattmundell:
```suggestion
                             get_bool_string (registry_allow_insecure));
```

src/manage_container_image_scanner.c
```diff
@@ -298,6 +298,38 @@ get_bool_string (const char *preference_value)
...
 
+/*
+* @brief Get the value of a named scanner preference from a list of preferences.
+*
+* @param[in]  name        The name of the scanner preferences.
+* @param[in]  scan_prefs  The list of preferences.
+*
+* @return     The value of the preference.
+*/
+static char *
```
mattmundell:
```suggestion
static gchar *
```

src/manage_container_image_scanner.c
```diff
@@ -298,6 +298,38 @@ get_bool_string (const char *preference_value)
...
+get_preference_from_list (char *name, GSList *scan_prefs)
+{
+  char *value = NULL;
```
mattmundell:
```suggestion
  gchar *value = NULL;
```
But you could just return the g_strdup directly below.

src/manage_container_image_scanner.c
```diff
@@ -323,6 +358,25 @@ add_container_image_scan_preferences (GHashTable *scanner_options, task_t task)
     g_hash_table_insert (scanner_options, g_strdup ("registry_allow_insecure"),
                          get_bool_string (registry_allow_insecure));
 
+  http_scanner_parsed_scans_preferences (connector, &scan_prefs);
```
mattmundell: This depends on a scanner, so it could fail. Makes me think the return should be checked (and passed on to the caller). Otherwise the scan will just keep running even when there's a problem on the scanner side.

PR #2605, opened 2025-10-29
Fix the check for credential store preferences

a-h-abdelsalam:
## What

Fix the check when updating credential store preferences.

## Why
Was crashing the preference is not predefined.

## References
GEA-1257


a-h-abdelsalam:
Changes look good to me although I prefer the older formatting. I cannot approve it since I initially opened it

PR #2609, opened 2025-11-04
Change: switch to checksum-only validation and drop CPE handling.

ozgen:
## What
- Use checksum-only stream validation for agent installers
- JSON parser updated: remove CPE expectations/iterators and related deletes.

## Why

Agent-installer metadata no longer provides `file_size` and `cpes`, so callers need a
checksum-only path.

## References

GEA-1307

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -14,7 +14,7 @@ Prerequisites:
 * glib-2.0 >= 2.42
 * gnutls >= 3.2.15
 * gpgme
-* [gvm-libs](https://github.com/greenbone/gvm-libs/) >= 22.30
+* [gvm-libs](https://github.com/greenbone/gvm-libs/) >= 22.31
 * libical >= 1.0.0
 * libbsd
 * pkg-config


bjoernricks (commented):
Could you update the required gvm-libs version in the INSTALL.md file too?

src/manage_agent_installers.c
```diff
@@ -403,28 +349,19 @@ static int
...
   const char *creation_time_str, *modification_time_str;
 
   memset (data, 0, sizeof(agent_installer_data_t));
 
   GET_AGENT_INSTALLER_JSON_STR (uuid, "uuid");
   GET_AGENT_INSTALLER_JSON_STR (name, "name");
   GET_AGENT_INSTALLER_JSON_STR (description, "description");
-  GET_AGENT_INSTALLER_JSON_STR (content_type, "contentType");
-  GET_AGENT_INSTALLER_JSON_STR (file_extension, "fileExtension");
-  GET_AGENT_INSTALLER_JSON_STR (installer_path, "installerPath");
```
timopollmeier: I think for consistency with other JSON files in the feed we should keep using camelCase here.

src/manage_agent_installers.c
```diff
@@ -440,43 +377,21 @@ get_agent_installer_data_from_json (cJSON *json,
...
       return -1;
     }
   data->modification_time = parse_iso_time (modification_time_str);
   if (data->modification_time == 0)
     {
       g_warning ("%s: Field '%s' is not a valid ISO date-time",
-                 __func__, "lastModified");
+                 __func__, "last_modified");
       return -1;
     }
```
timopollmeier: These should also be left as camelCase.

PR #2616, opened 2025-11-05
Logging agent sync throttle

ozgen:
## What

- Refactored the periodic maintenance logic in `serve_and_schedule()` to remove duplicate code.
- Added helper functions for schedule, feed, agents, and queue jobs.
- Introduced `AGENT_SYNC_SCHEDULE_PERIOD` to let agent synchronization run on its own interval.

## Why

- The previous implementation duplicated the same periodic logic before and after `pselect()`.
- Centralizing this logic reduces maintenance effort and improves clarity.
- Separate scheduling for agent sync provides better control and flexibility.
- Improves overall readability and consistency of the gvmd main loop.

## References

GEA-1306

### DIFF
--- a/src/gvmd.c
+++ b/src/gvmd.c
@@ -1654,11 +1654,12 @@ fork_agents_sync ()
 /**
  * @brief Periodic timestamps for background jobs in the main loop.
  */
-typedef struct {
-  time_t last_schedule;      ///< Last time the scheduler management was executed.
-  time_t last_feed_sync;     ///< Last time the feed sync was executed.
-  time_t last_queue;         ///< Last time queued task actions were executed.
-  time_t last_agents_sync;   ///< Last time the agents sync was executed.
+typedef struct
+{
+  time_t last_schedule;    ///< Last time the scheduler management was executed.
+  time_t last_feed_sync;   ///< Last time the feed sync was executed.
+  time_t last_queue;       ///< Last time queued task actions were executed.
+  time_t last_agents_sync; ///< Last time the agents sync was executed.
 } periodic_times_t;
 
 /**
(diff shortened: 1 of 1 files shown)


src/gvmd.c
```diff
@@ -1619,25 +1651,163 @@ fork_agents_sync ()
...
+  if (!time_to_run (t->last_schedule, SCHEDULE_PERIOD, now))
+    return;
+
+  int rc = manage_schedule (fork_connection_for_scheduler,
+                            scheduling_enabled,
+                            sigmask_normal);
+  switch (rc)
+    {
+    case 0:
+      set_now (&t->last_schedule, now);
```
mattmundell: There's a slight difference with the previous behaviour here, because before it would use the time after the `manage_schedule`. Maybe this will have an effect if the schedule takes a long time?

src/gvmd.c
```diff
@@ -1682,42 +1857,11 @@ serve_and_schedule ()
...
-                       __func__, last_schedule_time);
-              break;
-            case 1:
-              break;
-            default:
-              gvm_close_sentry ();
-              exit (EXIT_FAILURE);
-          }
-
-      if ((time (NULL) - last_sync_time) >= SCHEDULE_PERIOD)
```
mattmundell, line 1702: Similar situation here. The new one uses the same time, so this will be skipped even if it is time to fork the feed sync because manage_schedule took some time. The effect is minimal though I think, because in 5s we'll check again.

src/gvmd.c
```diff
@@ -1651,17 +1821,22 @@ serve_and_schedule ()
...
+
   g_info ("gvmd is ready to accept GMP connections");
+
   while (1)
     {
       int ret, nfds;
       fd_set readfds, exceptfds;
       struct timespec timeout;
 
+      run_periodic_block (&times, time (NULL));
```
mattmundell: This second arg is always the same, maybe move it into the function?

PR #2619, opened 2025-11-06
Change: guard agent-related filter names behind ENABLE_AGENTS

ozgen:
## What

Wrap agent filter names with `ENABLE_AGENTS` so they only build when the feature is enabled.

## Why

To hide agent-related filters in non-agent builds.

## References

GEA-1318


mattmundell (approved):

mattmundell (changes requested):
Oh the else is now dangling

mattmundell (approved):

PR #2629, opened 2025-11-10
Change: Refactor adding cs credentials to target.

a-h-abdelsalam:
## What

Refactored adding cs credentials to target.
Added task preference to specify scan behavior when failing to retrieve credentials from credential store.

## Why
To allow the user to define how the scan should behave if it was not possible to query the credentials.

## References
GEA-1339
GEA-1347

### DIFF
--- a/src/manage_credential_store_cyberark.c
+++ b/src/manage_credential_store_cyberark.c
@@ -20,6 +20,12 @@
 #include <gvm/util/fileutils.h>
 #include <gvm/util/tlsutils.h>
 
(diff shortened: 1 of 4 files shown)


timopollmeier (commented):
This looks fine overall but I think the error handling would be a bit more readable if we used a shared enum type for the return values of all the credential retrieval functions (including general, fetching from DB and from credential store).
That way you could tell easily if an error code is about a failed credential store retrieval or something else without having to check the function documentation.

PR #2633, opened 2025-11-11
Change: Split target alive_tests into sub-elements

timopollmeier:
## What
Instead of using predefined combinations of alive tests given as a string, the alive tests are now given and returned as one sub-element per test.

## Why
This allows users to select any combination of alive tests excluding "Scan Config Default" and "Consider Alive", which can only be selected alone.

## References
GEA-1346

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -21332,9 +21332,15 @@ alive_test_from_array (GPtrArray *alive_tests)
     {
       const char *item = g_ptr_array_index (alive_tests, i);
       if (strcasecmp (item, "Scan Config Default") == 0)
-        return 0;
+        {
+          alive_test_bitfield = 0;
+          break;
+        }
       else if (strcasecmp (item, "Consider Alive") == 0)
-        return ALIVE_TEST_CONSIDER_ALIVE;
+        {
+          alive_test_bitfield = ALIVE_TEST_CONSIDER_ALIVE;
+          break;
+        }
       else if (strcasecmp (item, "ARP") == 0
           || strcasecmp (item, "ARP Ping") == 0)
         alive_test_bitfield |= ALIVE_TEST_ARP;


src/gmp.c
```diff
@@ -18525,6 +18551,88 @@ handle_get_tags (gmp_parser_t *gmp_parser, GError **error)
...
+ * @param[in]  gmp_parser   GMP parser.
+ * @param[in]  error        Error parameter.
+ */
+static void
+send_alive_tests_str (int alive_tests,
+                      gmp_parser_t *gmp_parser,
+                      GError **error)
+{
+  if (alive_tests == 0)
+    SEND_TO_CLIENT_OR_FAIL ("Scan Config Default");
```
mattmundell: On failure this macro will send an error response then return, but the caller will keep running and sending messages to the client. The macro only works in the top-level handle function. I think calling send_to_client directly and passing a return to the caller would be a fine alternative.

src/gmp.c
```diff
@@ -24431,6 +24555,12 @@ gmp_xml_handle_end_element (/* unused */ GMarkupParseContext* context,
       CLOSE (CLIENT_CREATE_TARGET, REVERSE_LOOKUP_ONLY);
       CLOSE (CLIENT_CREATE_TARGET, REVERSE_LOOKUP_UNIFY);
       CLOSE (CLIENT_CREATE_TARGET, ALIVE_TESTS);
+      case CLIENT_CREATE_TARGET_ALIVE_TESTS_ALIVE_TEST:
+        g_ptr_array_add (create_target_data->alive_tests,
+                         g_strstrip (create_target_data->alive_test));
```
mattmundell: I'm wondering if it's OK to do this. I think ->alive_test could be NULL if the user sends `<alive_test/>` (because `gmp_handle_text` will not be called).

src/manage_sql.c
```diff
@@ -21314,7 +21314,51 @@ validate_results_port (const char *port)
...
+  if (alive_tests->len == 0)
+    return 0;
+
+  for (int i = 0; i < alive_tests->len; i++)
+    {
+      const char *item = g_ptr_array_index (alive_tests, i);
+      if (strcasecmp (item, "Scan Config Default") == 0)
+        return 0;
+      else if (strcasecmp (item, "Consider Alive") == 0)
+        return ALIVE_TEST_CONSIDER_ALIVE;
```
mattmundell: Nitpick: recommend making these top two and the `else` case match.

src/manage_sql.c
```diff
@@ -23040,43 +23119,12 @@ DEF_ACCESS (target_iterator_reverse_lookup_only,
...
 
-/**
- * @brief Get the alive test description from a target iterator.
- *
- * @param[in]  iterator  Iterator.
- *
- * @return Reverse lookup unify of the target or NULL if iteration is complete.
- */
-const char*
+int
```
mattmundell, line 23135: All the others have doc comments.

src/gmp.c
timopollmeier: This should be fixed with 433fd5712d0d2b1ddd07c476a50f030ad315a8ac where I introduced a new macro that also takes a return value in case of a send error.

src/gmp.c
timopollmeier: Good catch. A NULL would cause an error with `g_strstrip`, so with  96701ce52a85e7dd82a0529aee6d14e1a44a6bb0 an empty string is added to the array.

src/manage_sql.c
timopollmeier: I've added the comment in 96701ce52a85e7dd82a0529aee6d14e1a44a6bb0.

src/manage_sql.c
timopollmeier: I'm not sure if I correctly understood what you meant, but I've rewritten those cases in f3cb78c5fee7d27bcf5c17a335c6a0c5720fc361

mattmundell: Yes that's what I meant. Mixing `return`s and `break`s inside the same `if` statement feels messy to me. Like with 433fd5712d0d2b1ddd07c476a50f030ad315a8ac the `return`s make the `else`s redundant (or put another way, there could be one `return 0` right at the end because the `else`s control the flow). I am nitpicking though :)

mattmundell (approved):

PR #2637, opened 2025-11-20
Fix: build failure in launch_openvasd_openvas_task due to missing credential functions

ozgen:
## What

This change fixes a build failure in launch_openvasd_openvas_task() when compiling with ENABLE_CREDENTIAL_STORES=1.

## Why

The legacy _db credential helpers are not part of the credential-store build configuration, so compiling them with ENABLE_CREDENTIAL_STORES=1 causes build failures.

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -7245,7 +7245,6 @@ launch_openvasd_openvas_task (task_t task, target_t target, const char *scan_id,
   openvasd_target_t *openvasd_target;
   GSList *openvasd_targets, *vts;
   GHashTable *vts_hash_table;
-  openvasd_credential_t *ssh_credential, *smb_credential, *esxi_credential;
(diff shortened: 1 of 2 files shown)


src/manage.c
```diff
@@ -7349,6 +7352,7 @@ launch_openvasd_openvas_task (task_t task, target_t target, const char *scan_id,
     (openvasd_credential_t *) target_osp_esxi_credential_db (target);
   if (esxi_credential)
     openvasd_target_add_credential (openvasd_target, esxi_credential);
+#endif
```
a-h-abdelsalam, line 7355: I think the other case also needs to be handled when `ENABLE_CREDENTIAL_STORES=1`. Similar to `launch_osp_openvas_task`

PR #2647, opened 2025-11-25
Fix: format specifier for sizeof

mattmundell:
## What

Use `%zu` to format `sizeof` return.

Clean a little code formatting in the area at the same time.

## Why

`%zd` is for signed. `sizeof` returns the unsigned `size_t`, so it must use `%zu`.

This is a very minor fix because it only affects certain portability, but we may as will be consistent with how we do it in other places, eg at [lsc_crypt.c 912](https://github.com/greenbone/gvmd/blob/7cef9fd2c1304149408257f59f453db3853bcd2e/src/lsc_crypt.c#L912).

## References

[Explanation on SO](https://stackoverflow.com/questions/27296011/correct-format-specifier-for-return-value-of-sizeof-in-c)


bjoernricks:
Hi @mattmundell, I merged your other PR first and it created a bigger merge conflict :-/ could you fix that?

mattmundell:
Moved to /pull/2651.

PR #2654, opened 2025-11-27
Add: Support Kerberos and SNMP v3 credentials from credential store.

a-h-abdelsalam:
## What
Support Kerberos and SNMP v3 credentials from credential store.

## Why
Required for CyberArk integration

## References
GEA-1341

### DIFF
--- a/src/manage_tests.c
+++ b/src/manage_tests.c
@@ -21,8 +21,12 @@
 #include <cgreen/cgreen.h>
 
 Describe (manage);
-BeforeEach (manage) {}
-AfterEach (manage) {}
+BeforeEach (manage)
+{
+}
+AfterEach (manage)
+{
+}
 
 /* truncate_certificate */
 
@@ -61,6 +65,47 @@ Ensure (manage, truncate_certificate_given_truncated)
 
   truncated = truncate_certificate (given);
   assert_that (truncated, is_equal_to_string (given));
+  g_free (truncated);
+}
+
+Ensure (manage, truncate_certificate_empty_string)
+{
+  const gchar *given;
+  gchar *truncated;
+
+  given = "";
+
+  truncated = truncate_certificate (given);
+  assert_that (truncated, is_null);
+  g_free (truncated);
+}
+
+Ensure (manage, truncate_certificate_invalid_certificate)
+{
+  const gchar *given;
+  gchar *truncated;
+
+  given = "foo bar baz";
+
+  truncated = truncate_certificate (given);
+  assert_that (truncated, is_null);
+  g_free (truncated);
+}
+
+Ensure (manage, truncate_certificate_extra_data)
+{
+  const gchar *given, *expected;
+  gchar *truncated;
+  given = "-----BEGIN CERTIFICATE-----\n"
+          "MIIEjTCCAvWgAwIBAgIMWtd9bxgrX+9SgEHXMA0GCSqGSIb3DQEBCwUAMGIxKjAo\n"
+          "-----END CERTIFICATE-----\n"
+          "u1UuTizi5guqzOf+57s4o7Q=\n";
+  expected =
+    "-----BEGIN CERTIFICATE-----\n"
+    "MIIEjTCCAvWgAwIBAgIMWtd9bxgrX+9SgEHXMA0GCSqGSIb3DQEBCwUAMGIxKjAo\n"
+    "-----END CERTIFICATE-----\n";
+  truncated = truncate_certificate (given);
+  assert_that (truncated, is_equal_to_string (expected));
   g_free (truncated);
 }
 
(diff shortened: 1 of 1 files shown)


src/manage_openvas.c
```diff
@@ -392,6 +394,63 @@ set_auth_data_up_from_credential_store (iterator_t *iter,
...
+  if (cyberark_login_password_credential_data (cred_store_uuid,
+                                               vault_id,
+                                               host_identifier,
+                                               &login,
+                                               &password))
+    {
+      g_debug ("%s: Error retrieving SNMP username and password from"
+               " CyberArk credential store '%s'.",
+               __func__, cred_store_uuid);
+      return 1;
```
ozgen, line 430: Can you free login, password (and privacy_password where set) before returning in the error branches to avoid memory leaks?

src/manage_openvas.c
```diff
@@ -392,6 +394,63 @@ set_auth_data_up_from_credential_store (iterator_t *iter,
...
+  if (cyberark_login_password_credential_data (cred_store_uuid,
+                                               vault_id,
+                                               privacy_host_identifier,
+                                               NULL,
+                                               &privacy_password))
+    {
+      g_debug ("%s: Error retrieving SNMP privacy password from"
+               " CyberArk credential store '%s'.",
+               __func__, cred_store_uuid);
+      return 1;
```
ozgen, line 445: Can you free login, password (and privacy_password where set) before returning in the error branches to avoid memory leaks?

src/manage_openvas.c
ozgen, line 430: we currently use only 0 or 1. If callers need to distinguish different failure cases (e.g., CyberArk errors), we could introduce an enum with specific error codes. If all call sites just check != 0, keeping a single generic error code is fine.

src/manage_openvas.c
```diff
@@ -648,6 +710,201 @@ target_osp_esxi_credential (target_t target,
...
+
+          osp_credential_set_auth_data (osp_credential,
+                                        "auth_algorithm",
+                                        credential_iterator_auth_algorithm (&iter)
+                                          ?: "");
+          osp_credential_set_auth_data (osp_credential,
+                                        "privacy_algorithm",
+                                        credential_iterator_privacy_algorithm
+                                          (&iter) ?: "");
+          osp_credential_set_auth_data (osp_credential,
```
ozgen: In `target_osp_snmp_credential()`, the helper `set_auth_data_snmp_from_credential_store()` already sets `privacy_password`. Later in the `cs_snmp` branch, `privacy_password` is set again from the DB iterator, overwriting the CyberArk value. Is this intentional?

src/manage.c
```diff
@@ -7424,13 +7424,14 @@ launch_openvasd_openvas_task (task_t task, target_t target, const char *scan_id,
...
     openvasd_target_add_credential (openvasd_target, esxi_credential);
-#endif
 
   snmp_credential =
-    (openvasd_credential_t *) target_osp_snmp_credential (target);
+    (openvasd_credential_t *) target_osp_snmp_credential_db (target);
   if (snmp_credential)
     openvasd_target_add_credential (openvasd_target, snmp_credential);
 
+ #endif
```
mattmundell:
```suggestion
#endif
```

src/manage_sql.c
```diff
@@ -25031,6 +25067,28 @@ modify_credential (const char *credential_id,
...
+          && !g_str_has_prefix (type, "cs_"))
+        {
+          g_warning ("%s: Unknown credential type: %s", __func__, type);
+          sql_rollback ();
+          cleanup_iterator (&iterator);
+          return -1;
+        }
+
+      if (g_str_has_prefix (type, "cs_")
+          && !feature_enabled (FEATURE_ID_CREDENTIAL_STORES))
```
ozgen, line 25097: This flag check is much clearer than the previous version of mine :+1:

src/manage_sql.c
```diff
@@ -25123,17 +25182,22 @@ modify_credential (const char *credential_id,
               if (used_algorithm == NULL)
                 used_algorithm = "";
 
-              // privacy password must be empty if algorithm is empty
+              /*
+              privacy password and host identifier
+              must be empty if algorithm is empty
+              */
```
```suggestion
              /* privacy password and host identifier
               * must be empty if algorithm is empty */
```

src/manage_sql.c
a-h-abdelsalam: was getting more complex with my added checks

src/manage_openvas.c
a-h-abdelsalam: Since it is only checked for success or failure, I am keeping it simple for now. Thought about handling feature disabled as a different error but not really sure it helps much since a cs_credential cannot be created if the credential stores feature is disabled.

src/manage_openvas.c
a-h-abdelsalam: Good catch

PR #2662, opened 2025-12-01
Change: move timezone init out of print_report_xml_start

mattmundell:
## What

Use a separate function to do the timezone info init for `print_report_xml_start`.

## Why

To make `print_report_xml_start` easier to maintain.

## References

Follows #2653.

## Testing

I got a single report with the PR, then got the same report on the `main` branch. I made sure they were the same. I did that with two timezones just to be sure (using the `timezone` term in the filter). 

```sh
o m m '<get_reports report_id="92a09ffd-a5bd-464b-9f2c-54b78ec5ebcf" details="1" filter="timezone=Pacific/Auckland rows=3 apply_overrides=1 min_qod=60"/>' > /tmp/r-auk-main
o m m '<get_reports report_id="92a09ffd-a5bd-464b-9f2c-54b78ec5ebcf" details="1" filter="timezone=Europe/Berlin rows=3 apply_overrides=1 min_qod=60"/>' > /tmp/r-ber-main
diff /tmp/r-auk /tmp/r-auk-main
diff /tmp/r-ber /tmp/r-ber-main
```

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -7828,7 +7828,7 @@ task_severity_double (task_t task, int overrides, int min_qod, int offset)
   report_t report;
 
   if (current_credentials.uuid == NULL
-      || task_target (task) == 0 /* Container task. */)
+      || task_target (task) == 0 /* import task. */)
     return SEVERITY_MISSING;
 
   report = sql_int64_0 ("SELECT id FROM reports"
@@ -9927,7 +9927,7 @@ process_report_import (report_t report)
  * @brief Create a report from an array of results.
  *
  * @param[in]   results       Array of create_report_result_t pointers.
- * @param[in]   task_id       UUID of container task, or NULL to create new one.
+ * @param[in]   task_id       UUID of import task, or NULL to create new one.
  * @param[in]   in_assets     Whether to create assets from the report.
  * @param[in]   scan_start    Scan start time text.
  * @param[in]   scan_end      Scan end time text.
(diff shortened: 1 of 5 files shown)


src/manage_sql.c
```diff
@@ -17526,6 +17526,44 @@ struct print_report_context
...
 
+/**
+ * @brief Init zone info for print_report_xml_start.
+ *
+ * @param[in]  ctx  Printing context.
+ *
+ * @return 0 on success, -1 error.
+ */
+static int
+print_report_init_zone (print_report_context_t *ctx)
```
ozgen, line 17537: In the old code, if `ctx.zone ` was empty, we set `ctx.tz = NULL `and c`tx.old_tz_override = NULL`

In this new helper, when `ctx->zone` is empty, we do nothing and return 0.

Can you confirm that the context (ctx) is always created with all fields set to NULL?
Otherwise, these two fields may contain wrong values.

ozgen (approved):

src/manage_sql.c
ozgen: Is the context always zero-initialised? If yes, you can resolve the comment @mattmundell

mattmundell: Yes, at the top of `print_report_xml_start` there is
```
  print_report_context_t ctx = {0};
```
which initializes all fields in the struct to zero.

mattmundell: @ozgen Also, part of the idea of collecting the vars in a struct is they can all be freed together on exit by one function. Some description in #2563.

ozgen: Thanks for the clarification @mattmundell

PR #2667, opened 2025-12-05
Fix: Check default scanner against openvasd feature flag

ozgen:
## What

Use the runtime openvasd feature flag when creating the default scanner, and add a warning if an existing default scanner conflicts with it.

## Why

This ensures the default scanner matches the runtime feature settings and helps users detect and fix configuration mismatches.

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -4199,8 +4199,8 @@ check_db_scanners ()
           /* Runtime: openvasd scanner should be used */
           if (sc_type != SCANNER_TYPE_OPENVASD)
             {
-              g_warning ("Default scanner with UUID %s already exists, "
-                         "but the openvasd scanner feature is enabled. "
+              g_warning ("The openvasd scanner feature is enabled, but the "
+                         "default scanner with UUID %s uses a different type. "
                          "Not updating the scanner automatically; please adjust "
                          "the scanner configuration manually if needed.",
                          SCANNER_UUID_DEFAULT);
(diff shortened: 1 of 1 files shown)


src/manage_sql.c
```diff
@@ -4168,24 +4168,56 @@ check_db_scanners ()
...
+      /* Check existing default scanner against runtime feature flag */
+      scanner_type_t sc_type = get_scanner_type_by_uuid (SCANNER_UUID_DEFAULT);
+
+      if (feature_enabled (FEATURE_ID_OPENVASD_SCANNER))
+        {
+          /* Runtime: openvasd scanner should be used */
+          if (sc_type != SCANNER_TYPE_OPENVASD)
+            {
+              g_warning ("Default scanner with UUID %s already exists, "
+                         "but the openvasd scanner feature is enabled. "
```
timopollmeier:
```suggestion
              g_warning ("The openvasd scanner feature is enabled., but the"
                         "default scanner with UUID %s uses a different type. "
```

src/manage_sql.c
```diff
@@ -4168,24 +4168,56 @@ check_db_scanners ()
...
+                         "Not updating the scanner automatically; please adjust "
+                         "the scanner configuration manually if needed.",
+                         SCANNER_UUID_DEFAULT);
+            }
+        }
+      else
+        {
+          /* Runtime: openvas scanner should be used */
```
timopollmeier:
```suggestion
          /* Runtime: ospd-openvas scanner should be used */
```

timopollmeier (approved):

PR #2670, opened 2025-12-09
Add: Functions using SQL prepared statement syntax

timopollmeier:
## What
Functions are added that run SQL statements using the prepared statement syntax with numbered parameters starting with a dollar sign, e.g "$1", and passing the values to the SQL library function.

## Why
These are meant to replace the current SQL statement functions that insert values directly into the SQL string in printf style. The new method is more secure because it does not require strings to be quoted explicitly.

## References
GEA-1430

### DIFF
--- a/src/sql.c
+++ b/src/sql.c
@@ -15,6 +15,7 @@
  */
 
 #include "sql.h"
+
 #include "utils.h"
 
 #include <assert.h>
@@ -41,10 +42,10 @@
 /* Headers of internal symbols defined in backend files. */
 
 int
-sql_prepare_internal (int, int, const char*, va_list, sql_stmt_t **);
+sql_prepare_internal (int, int, const char *, va_list, sql_stmt_t **);
 
 int
-sql_prepare_ps_internal (int, int, const char*, va_list, sql_stmt_t **);
+sql_prepare_ps_internal (int, int, const char *, va_list, sql_stmt_t **);
 
 int
 sql_exec_internal (int, sql_stmt_t *);
@@ -88,8 +89,8 @@ int log_errors = 1;
  *
  * @return Freshly allocated, quoted string. Free with g_free.
  */
-gchar*
-sql_nquote (const char* string, size_t length)
+gchar *
+sql_nquote (const char *string, size_t length)
 {
   gchar *new, *new_start;
   const gchar *start, *end;
@@ -100,7 +101,8 @@ sql_nquote (const char* string, size_t length)
   /* Count number of apostrophes. */
 
   start = string;
-  while ((start = strchr (start, '\''))) start++, count++;
+  while ((start = strchr (start, '\'')))
+    start++, count++;
 
   /* Allocate new string. */
 
@@ -110,13 +112,13 @@ sql_nquote (const char* string, size_t length)
 
   start = string;
   end = string + length;
-  for (; start < end; start++, new++)
+  for (; start < end; start++, new ++)
     {
       char ch = *start;
       if (ch == '\'')
         {
           *new = '\'';
-          new++;
+          new ++;
           *new = '\'';
         }
       else
@@ -133,8 +135,8 @@ sql_nquote (const char* string, size_t length)
  *
  * @return Freshly allocated, quoted string. Free with g_free.
  */
-gchar*
-sql_quote (const char* string)
+gchar *
+sql_quote (const char *string)
 {
   assert (string);
   return sql_nquote (string, strlen (string));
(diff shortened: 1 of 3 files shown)


src/sql.h
```diff
@@ -27,6 +27,92 @@ typedef struct
...
+/**
+ * @brief Enum for defining the type of SQL prepared statement parameters.
+ */
+typedef enum
+{
+  SQL_PARAM_TYPE_NULL = 0,  ///< null value
+  SQL_PARAM_TYPE_DOUBLE,    ///< double precision floating point number
+  SQL_PARAM_TYPE_INT,       ///< integer
+  SQL_PARAM_TYPE_STRING,    ///< string
+  SQL_PARAM_TYPE_RESOURCE,  ///< resource rowid (resource_t)
```
ozgen: Could you change `rowid` as `row_id`?

src/sql.h
```diff
@@ -27,6 +27,92 @@ typedef struct
...
+
+/**
+ * @brief Union type for SQL prepared statement parameters values.
+ */
+typedef union
+{
+  double      double_value;   ///< double precision floating point value
+  int         int_value;      ///< integer value
+  char        *str_value;     ///< string value
+  resource_t  resource_value; ///< resource rowid (resource_t) value
```
ozgen: same here, Could you change rowid as row_id?

src/sql_pg.c
```diff
@@ -466,6 +466,93 @@ sql_prepare_internal (int retry, int log, const char* sql, va_list args,
...
+ * @param[in]  retry  Whether to keep retrying while database is busy or locked.
+ * @param[in]  log    Whether to keep retrying while database is busy or locked.
+ * @param[in]  sql    SQL statement template in prepared statement syntax.
+ * @param[in]  args   Parameters passed as a va_list.
+ * @param[out] stmt   Statement return.
+ *
+ * @return 0 success, 1 gave up, -1 error.
+ */
+int
+sql_prepare_ps_internal (int retry, int log, const char* sql, va_list args,
```
ozgen: `retry` is never used inside the function.

src/sql_pg.c
```diff
@@ -466,6 +466,93 @@ sql_prepare_internal (int retry, int log, const char* sql, va_list args,
...
+  int n_params = 0;
+
+  *stmt = (sql_stmt_t*) g_malloc (sizeof (sql_stmt_t));
+  sql_stmt_init (*stmt);
+  (*stmt)->sql = g_strdup (sql);
+
+  if (log)
+    g_debug ("   sql: %s", (*stmt)->sql);
+
+  va_copy (args_copy, args);
```
ozgen, line 493: You never call` va_end(args_copy)` after either `va_copy`

src/sql_pg.c
```diff
@@ -466,6 +466,93 @@ sql_prepare_internal (int retry, int log, const char* sql, va_list args,
...
+          pq_value = g_strdup_printf ("%d", param->value.int_value);
+          if (log)
+            g_debug ("sql param $%d (int) = %d",
+                     i+1, param->value.int_value);
+          break;
+        case SQL_PARAM_TYPE_STRING:
+          pq_value = g_strdup (param->value.str_value ?: "");
+          if (log)
+            g_debug ("sql param $%d (string) = %s",
+                     i+1, param->value.str_value);
```
ozgen: we could correctly handle NULL in `g_strdup` using ?: "", or we can use `pq_value` in the log.

src/sql_pg.c
```diff
@@ -466,6 +466,93 @@ sql_prepare_internal (int retry, int log, const char* sql, va_list args,
...
+ *
+ * @see sql_param_t for more info about passing the parameters.
+ *
+ * @param[in]  retry  Whether to keep retrying while database is busy or locked.
+ * @param[in]  log    Whether to keep retrying while database is busy or locked.
+ * @param[in]  sql    SQL statement template in prepared statement syntax.
+ * @param[in]  args   Parameters passed as a va_list.
+ * @param[out] stmt   Statement return.
+ *
+ * @return 0 success, 1 gave up, -1 error.
```
ozgen, line 475: Return value does not match the code, I mean, the code always returns 0

src/sql.c
```diff
@@ -309,6 +377,84 @@ sql_error (char* sql, ...)
...
+ * Return on error, instead of aborting.
+ *
+ * @param[in]  sql    Format string for SQL statement.
+ * @param[in]  ...    Arguments for format string.
+ *
+ * @return 0 success, 2 reserved (lock unavailable),
+ *         3 unique constraint violation, -1 error.
+ */
+int
+sql_error (char* sql, ...)
```
ozgen: Can we use `const char*` instead of `char*`?

src/sql_pg.c
timopollmeier: I think we should refactor this and `sql_prepare_internal ` to a `void` function in a separate PR.

src/sql_pg.c
timopollmeier: I would also change this in a separate PR.

src/sql.c
timopollmeier: Yes, but I'd try refactoring this for all the SQL functions in a separate PR.

ozgen (dismissed):

PR #2676, opened 2025-12-12
Add: public_key_info in get_credentials response

timopollmeier:
## What
The get_credentials response now contains a public_key_info element for PGP and S/MIME public key credentials, containing the fingerprint. 

## Why
This allows users to check if a key is present and identify it.

## References
GEA-1452

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -275,11 +275,11 @@ get_pgp_public_key_info (const char *key_str, gpgme_protocol_t protocol,
   int ret = 0;
   gpgme_ctx_t ctx;
   char gpg_temp_dir[] = "/tmp/gvmd-gpg-XXXXXX";
-  const gpgme_data_type_t openpgp_types_ptr[2] = {GPGME_DATA_TYPE_X509_CERT,
-                                                  GPGME_DATA_TYPE_PGP_KEY};
-  const gpgme_data_type_t cms_types_ptr[3] = {GPGME_DATA_TYPE_X509_CERT,
-                                              GPGME_DATA_TYPE_CMS_OTHER,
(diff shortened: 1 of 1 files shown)


src/gmp.c
```diff
@@ -259,6 +259,73 @@ check_certificate_smime (const char *cert_str)
...
+                                              GPGME_DATA_TYPE_CMS_OTHER,
+                                              GPGME_DATA_TYPE_PGP_KEY};
+
+  GArray *key_types;
+  gpgme_error_t err;
+  gpgme_key_t key;
+
+  *fingerprint = NULL;
+
+  if (mkdtemp (gpg_temp_dir) == NULL)
```
ozgen, line 290: you create tmp_dir, but never delete it. That means imported keys remain on disk, we should delete this tmp file after our job completed.

src/gmp.c
```diff
@@ -259,6 +259,73 @@ check_certificate_smime (const char *cert_str)
...
+static int
+get_pgp_public_key_info (const char *key_str, gpgme_protocol_t protocol,
+                         gchar **fingerprint)
+{
+  int ret = 0;
+  gpgme_ctx_t ctx;
+  char gpg_temp_dir[] = "/tmp/gvmd-gpg-XXXXXX";
+  const gpgme_data_type_t openpgp_types_ptr[2] = {GPGME_DATA_TYPE_X509_CERT,
+                                                  GPGME_DATA_TYPE_PGP_KEY};
+  const gpgme_data_type_t cms_types_ptr[3] = {GPGME_DATA_TYPE_X509_CERT,
```
ozgen: naming `*_ptr` is misleading, can we user  `*_arr` instead.

PR #2682, opened 2025-12-15
Fix: Validating 'sort_field' in GMP command 'get_nvts'

robindittmar:
## What

Validate `sort_field` in `handle_get_nvts()`

## Why

Not validating `sort_field` resulted in a possible SQL injection attack

## References

GEA-1420

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -14900,7 +14900,7 @@ handle_get_nvts (gmp_parser_t *gmp_parser, GError **error)
             }
         }
       else if (get_nvts_data->sort_field
-               && !validate_sort_field ("nvts", get_nvts_data->sort_field))
+               && validate_sort_field ("nvts", get_nvts_data->sort_field))
         {
           SEND_TO_CLIENT_OR_FAIL (
             XML_ERROR_SYNTAX ("get_nvts", "Invalid sort field"));
--- a/src/manage.h
+++ b/src/manage.h
@@ -247,7 +247,7 @@ current_encryption_key_uid (gboolean);
 void
 set_current_encryption_key_uid (const char *new_uid);
 
-gboolean
+int
 validate_sort_field (const gchar*, const gchar*);
 
 void
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -3134,12 +3134,12 @@ set_current_encryption_key_uid (const char *new_uid)
 /**
  * @brief Validates sort_field for nvts table
  *
- * @return 1 success, 0 invalid
+ * @return 0 success, 1 invalid
  */
-gboolean
+int
 validate_sort_field (const gchar *table, const gchar *sort_field)
 {
-  return db_table_has_column ("public", table, sort_field);
+  return !db_table_has_column ("public", table, sort_field);
 }
 
 


src/manage_sql_nvts.h
```diff
@@ -115,6 +115,38 @@
    { NULL, NULL, KEYWORD_TYPE_UNKNOWN }                                     \
  }
 
+#define NVT_VALID_SORTBY_COLUMNS \
```
a-h-abdelsalam: Just an idea, have you considered other approaches ? Other than hard-coding the column names, perhaps using `information_schema` inside an sql function that can be also reused elsewhere if needed. If it could be done in a safe way

rdttmr: I updated the code to dynamically fetch the column names when `check_db_nvts ()` is called.
As far as my testing goes this memory seems to be initialized prior to forking (for GMP handling) and works fine.

Since prepared statements are used, I assume this should be safe.

src/manage_pg.c
```diff
@@ -79,6 +79,58 @@ manage_db_empty ()
...
+  while (next (&column_it))
+    {
+      g_ptr_array_add (column_names,
+                       g_strdup (iterator_string (&column_it, 0)));
+    }
+  g_ptr_array_add (column_names, NULL);
+  cleanup_iterator (&column_it);
+
+  return (gchar **) g_ptr_array_free (column_names, FALSE);
+}
```
a-h-abdelsalam: I think this function is not needed any more.

src/manage_sql.h
```diff
@@ -187,6 +187,12 @@ typedef long long int rowid_t;
 int
 manage_db_empty ();
 
+gchar**
+get_db_table_column_names (const gchar*, const gchar*);
```
a-h-abdelsalam: Same here. This could be removed.
```suggestion
```

src/manage_sql.c
```diff
@@ -546,16 +546,18 @@ array_find_string (array_t *array, const gchar *string)
  *
  * @return The string from the vector if found, else NULL.
  */
-static const gchar*
+const gchar *
```
a-h-abdelsalam: I agree with removing the else but since this function is not used in the scope of the PR anymore, I suggest to keep it unchanged. Up to you.

src/manage_pg.c
```diff
@@ -79,6 +79,58 @@ manage_db_empty ()
...
+
+/**
+ * @brief Check whether a given table has a certain column
+ *
+ * @param schema The schema the table is located in (e.g. "public")
+ * @param table Name of the table to check column for
+ * @param column The column o check for existence
```
a-h-abdelsalam:
```suggestion
 * @param column The column to check for existence
```

src/manage_sql.h
```diff
@@ -187,6 +187,9 @@ typedef long long int rowid_t;
 int
 manage_db_empty ();
 
+int
+db_table_has_column(const gchar*, const gchar*, const gchar*);
```
mattmundell:
```suggestion
db_table_has_column (const gchar*, const gchar*, const gchar*);
```

src/manage_sql_nvts.c
```diff
@@ -2092,3 +2092,14 @@ nvts_oids_all_discovery_cached (GSList *oids)
...
   return TRUE;
 }
+
+/**
+ * @brief Validates sort_field for nvts table
+ *
+ * @return 0 success, -1 invalid
+ */
+int
+validate_nvts_sort_field (const char* sort_field)
```
mattmundell:
```suggestion
validate_nvts_sort_field (const gchar* sort_field)
```

src/manage_sql_nvts.h
```diff
@@ -160,4 +160,7 @@ nvts_discovery_oid_cache_reload ();
 gboolean
 nvts_oids_all_discovery_cached (GSList *oids);
 
+int
+validate_nvts_sort_field (const char*);
```
mattmundell:
```suggestion
validate_nvts_sort_field (const gchar*);
```

src/manage_pg.c
```diff
@@ -79,6 +79,27 @@ manage_db_empty ()
          == 0;
 }
 
+/**
+ * @brief Check whether a given table has a certain column
+ *
+ * @param schema The schema the table is located in (e.g. "public")
+ * @param table Name of the table to check column for
+ * @param column The column to check for existence
+ * @return 1 if column is present, else 0.
```
mattmundell, line 89:
```suggestion
 *
 * @return 1 if column is present, else 0.
```
I'm nitpicking but it's nice to format them all the same if you have the time. Also the `@params` could be aligned like the others.

src/gmp.c
```diff
@@ -14885,6 +14886,12 @@ handle_get_nvts (gmp_parser_t *gmp_parser, GError **error)
               return;
             }
         }
+      else if (get_nvts_data->sort_field
+               && validate_nvts_sort_field (get_nvts_data->sort_field))
```
mattmundell: I wonder if the idea is to do this for every GET command that can take a sort_field. If so it might be simpler to have `validate_sort_field ("nvts", get_nvts_data->sort_field)`.

src/gmp.c
```diff
@@ -103,6 +103,7 @@
 #include "manage_report_configs.h"
 #include "manage_report_formats.h"
 #include "manage_runtime_flags.h"
+#include "manage_sql_nvts.h"
```
mattmundell: gmp.c should not be importing a manage_sql* file. The idea is that there is a "manage" layer that the GMP layer uses. The manage layer exports from the non-SQL files like manage.h and manage_nvts.h. The manage_sql* stuff is internal to the manage layer.

The idea is to keep the parts that require SQL usage together, and to keep a path (at least in theory) for replacing all of the SQL with some other form of backend.

mattmundell: I'm also seeing some row ID things creeping into the "manage" layer API, eg `struct agent_group_data` includes a field called `row_id`. So does `struct agent_data` in manage_agents.h.

This breaks the existing pattern where the resource types are seen as handles on the resource. Yes internally they are rowids but the idea is that the layers above the "manage" layer see them as types/handles/instances/references.

All the existing code works like this. `ticket_in_use` takes a ticket.

mattmundell: My general pattern has always been that an arg like `target` means the handle/rowid and an arg like `target_id` means the UUID.

mattmundell: I'm also wondering about `agent_group_id_by_uuid` which breaks the `find_resource_with_permissions` or `find_resource_no_acl` pattern. I'm assuming the agent groups can't have specific permissions like the other resources, but it still seems like this is weakening the ACL layer by not going through one of those functions.

Also `agent_group_by_uuid` would fit the general `ticket`, `target` not `row_id` pattern better.

src/gmp.c
rdttmr: I did adopt your idea, I think it makes sense to make this function more generic

src/gmp.c
rdttmr: I re-arranged the code so that `#include "manage_sql_nvts.h"` is not required anymore in `gmp.c`.

Also I understand the separation of concerns and abstraction idea, however I am not quite sure what to do with the row ID information, I think this reaches beyond this PR?

src/manage.h
```diff
@@ -247,6 +247,9 @@ current_encryption_key_uid (gboolean);
 void
 set_current_encryption_key_uid (const char *new_uid);
 
+gboolean
+validate_sort_field (const gchar*, const gchar*);
```
mattmundell, line 251: Changing to a gboolean actually breaks the convention. The are many validate_* functions, like `validate_username` in manage.c. They generally return 0 on success, and some number on fail. Using an `int` allows returning more details about why the validation failed, see eg `validate_email_data`.

src/gmp.c
mattmundell: > I re-arranged the code so that #include "manage_sql_nvts.h" is not required anymore in gmp.c.

Great, thanks!

> Also I understand the separation of concerns and abstraction idea, however I am not quite sure what to do with the row ID information, I think this reaches beyond this PR?

Yes, definitely a separate PR. I was trying to communicate info, more than asking for a change.

PR #2699, opened 2026-01-02
fix: prevent authentication failure when auth_cache is locked.

ozgen:
## What

This PR prevents authentication from failing when the auth_cache table is locked.
If the cache cannot be accessed, authentication continues without using the cache.

## Why

The authentication cache is an optimization and should not block logins or API access.
Previously, temporary database locks could cause authentication failures.

## References

GEA-1489

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -5989,6 +5989,7 @@ authenticate_any_method (const gchar *username, const gchar *password,
 {
   int ret;
   gchar *hash;
+  gboolean use_cache = TRUE;
 
   sql_begin_immediate ();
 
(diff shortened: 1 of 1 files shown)


src/manage_sql.c
```diff
@@ -5997,7 +5998,9 @@ authenticate_any_method (const gchar *username, const gchar *password,
     {
       sql_rollback ();
       g_warning ("%s: Failed to acquire auth_cache lock", __func__);
-      return -1;
+      use_cache = FALSE;
+      // this begin is needed for other queries.
+      sql_begin_immediate ();
```
a-h-abdelsalam, line 6003: As the function is not returning here anymore, is the rollback still needed ? If not, I think this begin could also be removed.

ozgen: `sql_table_lock_wait()` can fail if the table is busy.
When it fails, the database marks the current transaction as broken.

`sql_rollback() ` is needed to reset the transaction.
Without it, all following SQL queries would fail.
@a-h-abdelsalam

PR #2702, opened 2026-01-06
Change: Move agent_installer_t and oci_image_target_t

mattmundell:
## What

Move `agent_installer_t` and `oci_image_target_t`.

Also sort the resource list.

## Why

All the other resource types are in `manage_resources.h`.

### DIFF
--- a/src/gmp_credential_stores.c
+++ b/src/gmp_credential_stores.c
@@ -11,6 +11,7 @@
  */
 
 #include "gmp_credential_stores.h"
+#include "gmp_get.h"
 #include "manage_credential_stores.h"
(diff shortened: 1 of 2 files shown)


bjoernricks (approved):

bjoernricks:
Hi @mattmundell, this PR breaks the build currently. Could you take a look?

mattmundell:
> Hi @mattmundell, this PR breaks the build currently. Could you take a look?

Builds now. I think only 5c6119d was required but I'm leaving 7c51c65 in because it's cleaner.

bjoernricks (approved):

PR #2703, opened 2026-01-06
Fix: Validate URL in HTTP GET alert

timopollmeier:
## What
The HTTP GET alert method now checks if the URL has a valid scheme prefix and is only used as the final positional argument for wget.

## Why
This prevents passing other command line options to wget instead of an actual URL.

## References
GEA-1474

### DIFF
--- a/src/manage_alerts.c
+++ b/src/manage_alerts.c
@@ -1044,8 +1044,8 @@ http_get (const char *url)
   cmd[3] = g_strdup ("--");
   cmd[4] = g_strdup (url);
   cmd[5] = NULL;
-  g_debug ("%s: Spawning in /tmp/: %s %s %s %s",
-           __func__, cmd[0], cmd[1], cmd[2], cmd[3]);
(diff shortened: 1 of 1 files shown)


src/manage_alerts.c
```diff
@@ -1030,12 +1030,20 @@ http_get (const char *url)
...
+  cmd = (gchar **) g_malloc (6 * sizeof (gchar *));
   cmd[0] = g_strdup ("/usr/bin/wget");
   cmd[1] = g_strdup ("-O");
   cmd[2] = g_strdup ("-");
-  cmd[3] = g_strdup (url);
-  cmd[4] = NULL;
+  cmd[3] = g_strdup ("--");
+  cmd[4] = g_strdup (url);
+  cmd[5] = NULL;
   g_debug ("%s: Spawning in /tmp/: %s %s %s %s",
```
jhelmold: Hi Timo, do you now also want to log cmd[4] here?

timopollmeier: Yes, I've overlooked that part and added it now.

PR #2715, opened 2026-01-14
Add: Allow pause/resume and excluding images in OCI targets.

a-h-abdelsalam:
## What
Parse `host_start` and `host_end` for container image results and add support for excluded images.
Extend create, modify, copy and get GMP commands for OCI image targets to support excluded images.

Uses https://github.com/greenbone/gvm-libs/pull/1033 

## Why
Allow pausing / resuming container image tasks and excluding images form an OCI target.

## References
GEA-1504

### DIFF
--- a/src/manage_sql_oci_image_targets.c
+++ b/src/manage_sql_oci_image_targets.c
@@ -40,7 +40,6 @@ create_oci_image_target (const char* name,
                          oci_image_target_t* oci_image_target,
                          gchar **error_message)
 {
-  gchar *quoted_name, *quoted_urls, *quoted_comment, *quoted_exclude_images;
   oci_image_target_t new_oci_image_target;
   credential_t credential = 0;
 
@@ -60,13 +59,10 @@ create_oci_image_target (const char* name,
       return CREATE_OCI_IMAGE_TARGET_EXISTS_ALREADY;
     }
 
-  quoted_name = sql_quote (name ?: "");
-
   if (g_str_equal (image_references, "")
       || !validate_oci_image_references (image_references, error_message))
     {
       sql_rollback ();
-      g_free (quoted_name);
       return CREATE_OCI_IMAGE_TARGET_INVALID_IMAGE_URLS;
     }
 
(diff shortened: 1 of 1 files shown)


src/manage_sql_oci_image_targets.c
```diff
@@ -118,19 +131,20 @@ create_oci_image_target (const char* name,
...
           g_free (quoted_name);
           g_free (quoted_urls);
+          g_free (quoted_exclude_images);
           g_free (quoted_comment);
-          return 3;
+          return CREATE_OCI_IMAGE_TARGET_INVALID_CREDENTIAL;
         }
     }
 
   sql ("INSERT INTO oci_image_targets"
```
rdttmr: Not sure if it makes sense to address this immediately, but `sql_ps()` could be used here to
a) avoid all the `sql_quote()` calls
b) use prepared statements, which are less prone to injection attacks

src/manage_sql_oci_image_targets.c
```diff
@@ -346,9 +360,42 @@ modify_oci_image_target (const char *oci_image_target_id, const char *name,
       g_free (quoted_image_references);
     }
 
+  if (exclude_images)
+    {
+      if (g_str_equal (exclude_images, ""))
+        {
+          sql ("UPDATE oci_image_targets SET"
```
rdttmr: Same as above, `sql_ps()` could be used here

src/manage_sql_oci_image_targets.c
```diff
@@ -346,9 +360,42 @@ modify_oci_image_target (const char *oci_image_target_id, const char *name,
...
+
+          if (!validate_oci_image_references (exclude_images, error_message))
+            {
+              sql_rollback ();
+              return MODIFY_OCI_IMAGE_TARGET_INVALID_EXCLUDE_IMAGES;
+            }
+
+          quoted_exclude_images = sql_quote (exclude_images);
+
+          sql ("UPDATE oci_image_targets SET"
```
rdttmr: Same as above, `sql_ps()` could be used here

src/manage_sql_oci_image_targets.c
```diff
@@ -819,6 +880,21 @@ oci_image_target_image_references (oci_image_target_t oci_image_target)
...
+ *
+ * @param[in]  oci_image_target  OCI Image Target.
+ *
+ * @return Newly allocated comma separated list of excluded images
+ *         if available, else NULL.
+ */
+char*
+oci_image_target_exclude_images (oci_image_target_t oci_image_target)
+{
+  return sql_string ("SELECT exclude_images FROM oci_image_targets"
```
rdttmr: Same as above, `sql_string_ps()` could be used here

PR #2719, opened 2026-01-19
Add idea for token based authentication

bjoernricks:
## What

Add idea for token based authentication

## Why
Draft an idea for a token based authentication in gvmd.

### DIFF
--- a/docs/token-based-authentication.md
+++ b/docs/token-based-authentication.md
@@ -1,0 +1,233 @@
+# Token Based Authentication <!-- omit in toc -->
+
+In general it is desirable to use token based authentication and
+[JSON Web Tokens (JWT)][JWT]
+for the format of the tokens. It allows to implement *stateless* authentication
+and therefore to split the authentication into another service.
+
+A service getting the JWT from the client has to verify the token
+signature only and can afterwards trust its content. The content of the JWT can
+contain arbitrary data which in our case should be the user's permissions.
+
+- [JWT Format](#jwt-format)
+- [GMP Authentication](#gmp-authentication)
+- [GMP Requests with Token](#gmp-requests-with-token)
+- [Authentication Workflow](#authentication-workflow)
+  - [Step 1](#step-1)
+  - [Step 2](#step-2)
+
+## JWT Format
+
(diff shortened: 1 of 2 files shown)


docs/token-based-authentication.md
```diff
@@ -0,0 +1,232 @@
...
+add claims for the user roles and permissions. The roles and permissions can be
+used in the services to decide wether a user is allowed to access a route.
+
+Proposed JWT format
+
+```json
+{
+  "sub": "<username>",
+  "exp": "<unix timestamp of current datetime + X>",
+  "iat": "<unix timestamp of current datetime>",
```
ozgen, line 31: Could we add `user_id` to JWT token proposed format?

bjoernricks: That's already mentioned. Should we define sub should be user id already?

> The JWT should use the username (or maybe better the user id) as subject

ozgen: Or we can add another field e.g.: `user_id :<user_uuid>`, Does this make sense?

docs/token-based-authentication.md
```diff
@@ -0,0 +1,233 @@
...
+- [GMP Authentication](#gmp-authentication)
+- [GMP Requests with Token](#gmp-requests-with-token)
+- [Authentication Workflow](#authentication-workflow)
+  - [Step 1](#step-1)
+  - [Step 2](#step-2)
+
+## JWT Format
+
+The JWT should use the user id as subject (`sub`) and add [claims](https://en.wikipedia.org/wiki/JSON_Web_Token#Standard_fields)
+for the user name, roles and permissions. The roles and permissions can be used
```
mattmundell, line 22: First, if we're talking about gvmd permissions as they currently are (as returned by GET_PERMISSIONS), then technically you only need the permissions to decide if the route is allowed. The list of permissions includes the permissions granted to the roles. So you don't also need the roles.

Second, if you do want to send the roles anyway, then maybe you also want to send the groups (because the groups also determine the permissions).

bjoernricks: Thanks for the hint. Added groups in https://github.com/greenbone/gvmd/pull/2738

PR #2721, opened 2026-01-20
Change: Parse image digest, host details, host_start and host_end

a-h-abdelsalam:
## What
Parse image digest, host details, host_start and host_end for container scanning.

## Why
They are now sent from the scanner and to support pausing and resuming container image tasks.

## References
GEA-1519

### DIFF
--- a/src/manage_container_image_scanner.c
+++ b/src/manage_container_image_scanner.c
@@ -133,19 +133,7 @@ add_container_image_scan_result (http_scanner_result_t res,
   type = res->type;
   test_id = res->oid;
 
-  host = NULL;
-  if (res->ip_address)
-    {
-      gchar *ip_suffix = g_strstr_len (res->ip_address, -1, "@sha256:");
-      if (ip_suffix)
-        {
-          host = ip_suffix + 1;
-        }
-      else
-        {
-          host = res->ip_address;
-        }
-    }
+  host = extract_sha256_digest_if_found (res->ip_address);
   hostname = res->hostname;
   port = res->port;
   desc = res->message;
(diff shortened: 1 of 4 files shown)


src/manage_container_image_scanner.c
```diff
@@ -129,53 +129,92 @@ add_container_image_scan_result (http_scanner_result_t res,
...
+  type = res->type;
   test_id = res->oid;
-  // hostname is used as host since there is no IP
-  if (res->hostname && g_str_has_prefix (res->hostname, "oci://"))
-    host = res->hostname + strlen ("oci://");
-  else
-    host = res->hostname;
+
+  host = NULL;
+  if (res->ip_address)
```
ozgen: Extracting the digest logic is duplicated in multiple places; maybe we can move this logic into a helper function in `manage_utils`.

src/manage_sql.c
```diff
@@ -14294,6 +14294,54 @@ set_scan_host_start_time_ctime (report_t report, const char* host,
...
+ * @param[in]  report     Report associated with the scan.
+ * @param[in]  host       Host.
+ * @param[in]  timestamp  End time. In ISO format.
+ */
+void
+set_scan_host_end_time_isotime (report_t report, const char* host,
+                                const char* timestamp)
+{
+  gchar *quoted_host;
+  quoted_host = sql_quote (host);
```
ozgen, line 14318: We could validate the inputs before proceed e.g.

```
  gchar *quoted_host;
  int end_time;

  if (host == NULL || timestamp == NULL)
    return;

  end_time = parse_iso_time (timestamp);
  if (end_time <= 0)
    return;

  quoted_host = sql_quote (host);
```

src/manage_sql.c
```diff
@@ -41592,14 +41641,27 @@ check_http_scanner_result_exists (report_t report,
...
+          const char *desc, *type, *severity = NULL, *host = NULL;
           const char *hostname, *port = NULL;
           gchar *quoted_desc, *quoted_type, *quoted_host;
           gchar *quoted_hostname, *quoted_port;
           double severity_double = 0.0;
           int qod_int = get_http_scanner_nvti_qod (res->oid);
 
-          host = res->ip_address;
+          // ip_address for container scanning is an image digest
+          if (res->ip_address)
```
ozgen: Duplicated logic of extracting digest.

src/manage_container_image_scanner.c
a-h-abdelsalam: Moved to a helper function.

src/manage_sql.c
a-h-abdelsalam: Moved the validation of `host` and `timestamp` inside the function.

ozgen (approved):

PR #2724, opened 2026-01-21
Add: Propagate agent_update_available and updater_update_available

ozgen:
## What

Updates **gvmd** to handle and expose the `agent_update_available` and `updater_update_available` fields in agent processing and responses.

## Why

This ensures agent and updater update availability information is correctly propagated to consumers.

## References

GEA-1523


timopollmeier (changes requested):
It looks like the new columns are missing in the `CREATE TABLE` statements in `create_tables`.

timopollmeier (approved):

PR #2725, opened 2026-01-21
Removed hosts_ordering option.

jhelmold:
## What
The hosts ordering option has never been used, so it has been removed from the sources.

## Why
The hosts ordering option was never used.

## References
GEA-1499

## Checklist

### DIFF
--- a/src/manage_agents.c
+++ b/src/manage_agents.c
@@ -884,3 +884,4 @@ agent_response_to_string (agent_response_t code)
 }
 
 #endif // ENABLE_AGENTS
+


src/gmp.c
```diff
@@ -7101,6 +7099,7 @@ gmp_xml_handle_start_element (/* unused */ GMarkupParseContext* context,
             gvm_append_string (&modify_task_data->comment, "");
             set_client_state (CLIENT_MODIFY_TASK_COMMENT);
           }
+        /* HOSTS_ORDERING kept for reasons of compatibility */
```
mattmundell: I'd prefer if these were removed.

GMP will just "read over" any foreign elements. See the `ELSE_READ_OVER` that's everywhere in gmp.c, and the set_read_over at the end of `gmp_xml_handle_start_element`.

For example, if you run this the comments is still modified:
```xml
o m m '<modify_task task_id="8831937d-0e87-439a-807c-a9a2cf976f6a"><random_element>abc</random_element><comment>123</comment></modify_task>'
```

src/manage_migrators.c
```diff
@@ -3576,6 +3576,39 @@ migrate_265_to_266 ()
   return 0;
 }
 
+/**
+ * @brief Migrate the database from version 266 to version 267.
+ *
+ * @return 0 success, -1 error.
+ */
+int
+migrate_266_to_267 ()
```
mattmundell, line 3533: This number is used now.

mattmundell (changes requested):

PR #2726, opened 2026-01-21
Change: Improve validation for OCI images and exclude images.

a-h-abdelsalam:
## What
Improve validation of OCI images and exclude images. Trim spaces and replace newlines with commas

## Why
Resolve unnecessary validation errors blocking file upload.

## References
GEA-1504

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -8067,7 +8067,7 @@ fork_agent_controller_scan_handler (task_t task, agent_group_t agent_group,
     *report_id_return = NULL;
 
   current_scanner_task = task;
-  set_task_run_status (task, TASK_STATUS_DONE);
+  set_task_run_status (task, TASK_STATUS_REQUESTED);
 
   switch (fork ())
     {


src/manage_oci_image_targets.c
```diff
@@ -203,6 +203,70 @@ valid_oci_url (const gchar *oci_url)
...
+                                            g_free, NULL);
+  GPtrArray *clean_array = make_array ();
+
+  while (*point)
+    {
+      g_strstrip (*point);
+      if (**point != '\0')
+        {
+          gchar *key = g_strdup (*point);
+          if (!g_hash_table_contains (seen, key))
```
ozgen, line 245: In here we do extra lookup on hash table by using `g_hash_table_contains` and `g_hash_table_add`
could be reduced only using `g_hash_table_add` because it returns `true`  if the key did not exist yet.
ref: https://docs.gtk.org/glib/type_func.HashTable.add.html

a-h-abdelsalam: @ozgen If `g_hash_table_add` is used, then as mentioned in the docs, "if the key already exists in the hash table, then the old copy of key in the hash table is freed and key replaces it in the table." which is also some extra work.

Additionally, lookup in a hash table is not expensive, so there isn't much of an extra gain.

PR #2738, opened 2026-01-27
Add groups to the JWT token format

bjoernricks:
## What
Add groups to the JWT token format

## Why

Maybe we just stick with the permissions in future but for now we add all three user related content: roles, groups and permissions to the JWT format.

## References

https://github.com/greenbone/gvmd/pull/2719#discussion_r2727613566

### DIFF
--- a/docs/token-based-authentication.md
+++ b/docs/token-based-authentication.md
@@ -19,8 +19,9 @@ contain arbitrary data which in our case should be the user's permissions.
 ## JWT Format
 
 The JWT should use the user id as subject (`sub`) and add [claims](https://en.wikipedia.org/wiki/JSON_Web_Token#Standard_fields)
-for the user name, roles and permissions. The roles and permissions can be used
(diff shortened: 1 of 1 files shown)


docs/token-based-authentication.md
```diff
@@ -36,6 +37,12 @@ Proposed JWT format
     "User",
     "..."
   ],
+  "groups": [
+    "<groups of the user, for example>",
+    "SomeGroup",
```
mattmundell, line 42: Multiple groups may have the same name. Names are only unique per user. And the groups of multiple users can affect the permissions. So I guess these should be UUIDs instead of strings. Or perhaps something like:
```json
{ "id": "9029423b-9512-4691-94ae-a4ecfc6ecaff",
  "name": "SomeGroup" }
```
This format could even mirror the GET_GROUPS response, but in JSON instead of XML.

Same for roles.

PR #2743, opened 2026-01-29
Clear password from memory

janlucasschwoerer-greenbone:
## What
Before the memory related to a password (and other private values) is freed, we will now clear them from memory.

## Why
This will ensure that auth/credential related values are available in memory for as short as absolutely necessary. This will prevent potential that an attacker could get access to the auth/credential values.

## References
https://jira.greenbone.net/browse/GEA-1158

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -554,6 +554,9 @@ typedef struct
 static void
 create_credential_data_reset (create_credential_data_t *data)
 {
+  memset (data->password, 0, strlen (data->password));
+  memset (data->privacy_password, 0, strlen (data->privacy_password));
+
   free (data->allow_insecure);
   free (data->certificate);
   free (data->comment);
@@ -1211,6 +1214,8 @@ typedef struct
 static void
 create_user_data_reset (create_user_data_t * data)
 {
+  memset (data->password, 0, strlen (data->password));
+
   g_free (data->copy);
   array_free (data->groups);
   g_free (data->name);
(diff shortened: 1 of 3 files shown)


src/manage_openvas.c
```diff
@@ -268,7 +268,7 @@ set_auth_data_ssh_from_credential_store (iterator_t *iter,
     = credential_iterator_host_identifier (iter);
 
   if (cyberark_login_password_credential_data (cred_store_uuid,
-                                               vault_id,
+                                               vault_id,gm
```
timopollmeier, line 271:
```suggestion
                                               vault_id,
```

src/gmp.c
mattmundell, line 570: Would be nice if these were combined into a single call, like:

```suggestion
  free_password (data->password);
```

src/gmp.c
```diff
@@ -2644,6 +2649,9 @@ typedef struct
 static void
 modify_credential_data_reset (modify_credential_data_t *data)
 {
+  memset (data->password, 0, strlen (data->password));
+  memset (data->privacy_password, 0, strlen (data->password));
```
mattmundell, line 2653: Strlen should be on same field.

src/manage_openvas.c
```diff
@@ -440,6 +444,9 @@ set_auth_data_snmp_from_credential_store (iterator_t *iter,
       g_debug ("%s: Error retrieving SNMP privacy password from"
                " CyberArk credential store '%s'.",
                __func__, cred_store_uuid);
+
+      password (password, 0, strlen (password));
```
mattmundell, line 448: did you mean memset?

src/manage_alerts.c
```diff
@@ -4402,6 +4417,9 @@ trigger_to_vfire (alert_t alert, task_t task, report_t report, event_t event,
       get_data_reset (alert_filter_get);
       g_free (alert_filter_get);
     }
+
+  memset (password, 0, strlen (password));
```
mattmundell, line 4421: `password` technically can be NULL, from `credential_encrypted_value`

mattmundell (changes requested):

PR #2751, opened 2026-02-02
Clarify roles and groups in JWT for new auth

bjoernricks:
## What

Clarify roles and groups in JWT for new auth

## Why

Currently it is not clear if adding roles and groups to the JWT for the new token based authentication is needed or even useful. Therefore declare them as optional for now and also use objects as we might need both the ids and names.

### DIFF
--- a/docs/token-based-authentication.md
+++ b/docs/token-based-authentication.md
@@ -19,9 +19,9 @@ contain arbitrary data which in our case should be the user's permissions.
 ## JWT Format
 
 The JWT should use the user id as subject (`sub`) and add [claims](https://en.wikipedia.org/wiki/JSON_Web_Token#Standard_fields)
-for the user name, roles, groups and permissions. The roles, groups and
-permissions can be used in the services to decide wether a user is allowed to
-access a route.
+for the user name and permissions. Optionally the roles and groups could be
+added to if necessary. The permissions, roles and groups can be used in the
+services to decide whether a user is allowed to access a route.
 
 Proposed JWT format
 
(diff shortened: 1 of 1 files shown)


docs/token-based-authentication.md
```diff
@@ -19,9 +19,9 @@ contain arbitrary data which in our case should be the user's permissions.
 ## JWT Format
 
 The JWT should use the user id as subject (`sub`) and add [claims](https://en.wikipedia.org/wiki/JSON_Web_Token#Standard_fields)
-for the user name, roles, groups and permissions. The roles, groups and
-permissions can be used in the services to decide wether a user is allowed to
-access a route.
+for the user name and permissions. Optionally the roles and groups could be
+added to if necessary. The permissions, roles and groups can be used in the
+services to decide wether a user is allowed to access a route.
```
mattmundell:
```suggestion
services to decide whether a user is allowed to access a route.
```

docs/token-based-authentication.md
```diff
@@ -31,24 +31,22 @@ Proposed JWT format
...
-  ],
-  "groups": [
-    "<groups of the user, for example>",
-    "SomeGroup",
-    "AnotherGroup",
-    "..."
-  ]
   "permissions": [
     "<permissions of the user, for example>",
     "get_tasks",
```
mattmundell, line 36: These will also need to be more detailed.
For example, there are permissions on specific resources, and there are the Super permissions.

mattmundell (dismissed):

docs/token-based-authentication.md
bjoernricks: Currently I only have the permissions in mind which are returned by the [`<help>`](https://docs.greenbone.net/API/GMP/gmp-22.7.html#command_help) command.

bjoernricks: The permissions on specific resources are currently out of scope for me as they are (still) returned in all the resource responses.

mattmundell: OK. If the gvmd clients only want to know if the user is allowed to run one of the GMP commands, then that list is enough. Something to consider: the list may become stale when permissions are modified.

PR #2757, opened 2026-02-04
Change: move stray permission headers

mattmundell:
## What

Move permission headers to right file.

Also add missing log domains, and clean a couple comments. 

## Why

Headers where in old main file.

## Testing

Ran gvmd with verbose logs to check domain.
For create_permissions*: gave Full and Fast a new UUID in the db, then checked that --rebuild-gvmd-data=configs created a new Full and Fast with permissions.


mattmundell:
> @mattmundell could you resolve the conflict? thanks

@bjoernricks thanks, done

PR #2764, opened 2026-02-11
Change: remove some unused functions

mattmundell:
## What

Remove out of use functions.

These are shown by the `UNUSED_FUNCTION_FLAGS`. I searched the code for each one, in case the calls were hidden behind `#if`s. I also checked why they were unused, by searching for refs in past commits.

## Why

Cleaner.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -351,9 +351,9 @@ set(
 )
 set(LINKER_HARDENING_FLAGS "-Wl,-z,relro -Wl,-z,now")
 
-#set(UNUSED_FUNCTIONS_FLAGS "")
+set(UNUSED_FUNCTIONS_FLAGS "")
 # Enable this to find unused functions, and search for "removing" in output.
-set(UNUSED_FUNCTIONS_FLAGS "-flto -fwhole-program -ffunction-sections -Wl,--gc-sections -Wl,--print-gc-sections")
+#set(UNUSED_FUNCTIONS_FLAGS "-flto -fwhole-program -ffunction-sections -Wl,--gc-sections -Wl,--print-gc-sections")
 set(
   CMAKE_C_FLAGS
   "${CMAKE_C_FLAGS} ${UNUSED_FUNCTIONS_FLAGS} -Wall -D_BSD_SOURCE -D_ISOC99_SOURCE -D_SVID_SOURCE -D_DEFAULT_SOURCE -D_FILE_OFFSET_BITS=64 ${COVERAGE_FLAGS}"


CMakeLists.txt
```diff
@@ -351,10 +351,12 @@ set(
 )
 set(LINKER_HARDENING_FLAGS "-Wl,-z,relro -Wl,-z,now")
 
-# To find unused functions, add: -flto -fwhole-program -ffunction-sections -Wl,--gc-sections -Wl,--print-gc-sections
+set(UNUSED_FUNCTIONS_FLAGS "")
+# Enable this to find unused functions, and search for "removing" in output.
+#set(UNUSED_FUNCTIONS_FLAGS "-flto -fwhole-program -ffunction-sections -Wl,--gc-sections -Wl,--print-gc-sections")
```
bjoernricks, line 356: Should we maybe enable this by default for gsad and gvmd?

bjoernricks: I was thinking about something like this yesterday btw. because in rust it is a default of the compiler to complain about unused code.

mattmundell: It would be nice, but would take some adjustment
1. Sometimes functions are used behind feature flag #ifs (we could move them inside the #ifs)
2. Sometimes functions are added for API completeness, eg get_feed_lock_timeout in gvmd a932f70f (we could just remove these)

The output is like this, so we'd have to filter for `.*removing unused section '\.text\..*'`
```
/usr/bin/ld: removing unused section '.rodata.cst4' in file '/usr/lib/gcc/x86_64-linux-gnu/14/../../../x86_64-linux-gnu/Scrt1.o'
/usr/bin/ld: removing unused section '.data' in file '/usr/lib/gcc/x86_64-linux-gnu/14/../../../x86_64-linux-gnu/Scrt1.o'
/usr/bin/ld: removing unused section '.text.get_broker_address' in file '/tmp/cccaX7gk.ltrans0.ltrans.o'
/usr/bin/ld: removing unused section '.text.get_feed_lock_timeout' in file '/tmp/cccaX7gk.ltrans0.ltrans.o'
```

There are also a bunch of unused functions in gvmd manage_pg_server.c that say `Dummy function to allow restoring gvmd-9.0 dumps.` and `This function will be removed once direct migration compatibility with gvmd 9.0 is no longer required`. I'm unsure when these can be removed.

mattmundell: Another one:
 3. Sometimes functions are used only in the tests, like `gsad_reset_validator` (we could move the function into the test)

bjoernricks: Ok. It's an improvement for the future.

bjoernricks (approved):

PR #2769, opened 2026-02-16
Add get_timezones GMP command and ensure users' timezones are supported

timopollmeier:
## What
The GMP command GET_TIMEZONES is added and before setting the timezone for a user, it is checked whether it is
supported.

## Why
The new command can be used by clients to ensure only supported timezones can be
set when changing user settings. 
The added check will then enforce this and also prevent SQL errors in case an unsupported timezone is already set in the users database table.

## References
GEA-1573

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -20335,7 +20335,7 @@ handle_get_timezones (gmp_parser_t *gmp_parser, GError **error)
                           " status=\"" STATUS_OK "\""
                           " status_text=\"" STATUS_OK_TEXT "\">");
 
-  timezones = manange_get_timezones ();
+  timezones = manage_get_timezones ();
   for (int i = 0; i < timezones->len; i++)
     {
       const char *name = timezones->pdata[i];
--- a/src/manage.c
+++ b/src/manage.c
@@ -6190,12 +6190,12 @@ gvm_migrate_secinfo (int feed_type)
 /* Time zone info. */
 
 /**
- * @brief Get all a list of supported timezones
+ * @brief Get a list of all supported timezones
  *
  * @return An array of supported timezones. Caller must free.
  */
 array_t *
-manange_get_timezones ()
+manage_get_timezones ()
 {
   array_t *tzs_out = make_array ();
   iterator_t pg_iterator;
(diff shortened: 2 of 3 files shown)


src/manage.c
```diff
@@ -6186,6 +6186,48 @@ gvm_migrate_secinfo (int feed_type)
...

+/* Time zone info. */
+
+/**
+ * @brief Get all a list of supported timezones
+ *
+ * @return An array of supported timezones. Caller must free.
+ */
+array_t *
+manange_get_timezones ()
```
ozgen:
```suggestion
manage_get_timezones ()
```

src/manage.c
```diff
@@ -6186,6 +6186,48 @@ gvm_migrate_secinfo (int feed_type)
   return ret;
 }
 
+
+/* Time zone info. */
+
+/**
+ * @brief Get all a list of supported timezones
```
ozgen:
```suggestion
 * @brief Get a list of supported timezones
```

src/manage.h
```diff
@@ -3333,6 +3333,14 @@ manage_asset_snapshot_delete_stale (int);
 int
 manage_dump_asset_snapshot_counts (GSList *, const db_conn_info_t *);
 
+
+/* Timezone info. */
+array_t *
+manange_get_timezones ();
```
ozgen:
```suggestion
manage_get_timezones ();
```

src/gmp.c
```diff
@@ -20315,6 +20321,35 @@ handle_get_tasks (gmp_parser_t *gmp_parser, GError **error)
...
+ */
+static void
+handle_get_timezones (gmp_parser_t *gmp_parser, GError **error)
+{
+  array_t *timezones;
+  SEND_TO_CLIENT_OR_FAIL ("<get_timezones_response"
+                          " status=\"" STATUS_OK "\""
+                          " status_text=\"" STATUS_OK_TEXT "\">");
+
+  timezones = manange_get_timezones ();
```
timopollmeier:
```suggestion
  timezones = manage_get_timezones ();
```

PR #2791, opened 2026-02-24
Add: oci_image element in GMP results

timopollmeier:
## What
The results of the get_results and get_reports GMP commands can now be filtered by various OCI image fields and can also contain a new sub-element with OCI image information if available.

## Why
This allows better filtering of container image scan reports and simplifies handling of the new filterable values.

## References
GEA-1582

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -12192,11 +12192,12 @@ result_iterator_oci_image_path (iterator_t* iterator)
 }
 
 /**
- * @brief Get container image organisation from a result iterator.
+ * @brief Get container image short name (last part including tag)
+ *        from a result iterator.
  *
  * @param[in]  iterator  Iterator.
  *
- * @return The container image organisation.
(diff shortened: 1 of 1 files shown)


src/manage_sql.c
```diff
@@ -12105,6 +12135,76 @@ result_iterator_compliance (iterator_t* iterator)
...
+  if (iterator->done) return 0;
+  return iterator_string (iterator, GET_ITERATOR_COLUMN_COUNT + 39);
+}
+
+/**
+ * @brief Get container image organisation from a result iterator.
+ *
+ * @param[in]  iterator  Iterator.
+ *
+ * @return The container image organisation.
```
a-h-abdelsalam:
```suggestion
 * @brief Get container image short name (last part including tag) from a result iterator.
 *
 * @param[in]  iterator  Iterator.
 *
 * @return The container image short name.
```

PR #2795, opened 2026-02-24
Fix: Add exclude_images to trash OCI image targets

timopollmeier:
## What
The previously missing field exclude_images is added to the oci_image_targets_trash table. The delete and restore functions are adjusted accordingly.

## Why
This fixes various issues with OCI image targets in the trashcan not working correctly.

## References
GEA-1619

### DIFF
--- a/src/manage_sql_port_lists.c
+++ b/src/manage_sql_port_lists.c
@@ -2165,7 +2165,7 @@ init_port_range_iterator (iterator_t* iterator, port_list_t port_list,
     }
   else
     init_iterator (iterator,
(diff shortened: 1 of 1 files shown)


src/manage_migrators.c
```diff
@@ -3677,6 +3677,40 @@ migrate_268_to_269 ()
   return 0;
 }
 
+/**
+ * @brief Migrate the database from version 264 to version 265.
```
ozgen:
```suggestion
 * @brief Migrate the database from version 269 to version 270.
```

timopollmeier: I've amended my commit to fix this.

PR #2859, opened 2026-04-01
Add: Option to auto-update relays from file

timopollmeier:
## What
The command llne option --relays-path is added which makes gvmd update the relay_host and relay_port fields of scanners automatically from a JSON file with the given path.

## Why
This is meant as a replacement for the deprecated --relay-mapper option.

## References
GEA-1685

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -50,6 +50,7 @@
 #include "manage_oci_image_targets.h"
 #include "manage_http_scanner.h"
 #include "manage_runtime_flags.h"
+#include "manage_scanner_relays.h"
 #include "manage_sql.h"
 #include "manage_sql_assets.h"
 #include "manage_sql_resources.h"
@@ -91,6 +92,7 @@
 #include <gvm/util/uuidutils.h>
 #include <gvm/util/versionutils.h>
 #include <gvm/gmp/gmp.h>
+
 
 #undef G_LOG_DOMAIN
 /**
@@ -4218,6 +4220,8 @@ manage_sync (sigset_t *sigmask_current,
 
   reinit_manage_process ();
   manage_session_init (current_credentials.uuid);
+
+  sync_scanner_relays ();
 
   if (feed_sync_required ())
     {
--- a/src/manage.c
+++ b/src/manage.c
@@ -93,7 +93,6 @@
 #include <gvm/util/versionutils.h>
 #include <gvm/gmp/gmp.h>
 
-
 #undef G_LOG_DOMAIN
 /**
  * @brief GLib log domain.
--- a/src/manage_scanner_relays.c
+++ b/src/manage_scanner_relays.c
@@ -129,7 +129,7 @@ extract_relay_fields_from_json_item (cJSON *json_item,
   item_out->original_port = 0;
   item_out->relay_host = NULL;
   item_out->relay_port = 0;
-  item_out->protocol = NULL;
+  item_out->scanner_type = NULL;
 
   if (json_item == NULL)
     {
@@ -145,12 +145,26 @@ extract_relay_fields_from_json_item (cJSON *json_item,
                           &item_out->relay_host);
   gvm_json_obj_check_int (json_item, "relay_port",
                           &item_out->relay_port);
-  gvm_json_obj_check_str (json_item, "protocol",
-                          &item_out->protocol);
+  gvm_json_obj_check_str (json_item, "scanner_type",
+                          &item_out->scanner_type);
 
   if (item_out->original_host == NULL)
     {
       g_debug ("%s: mandatory original_host field missing or not a string",
+               __func__);
+      return -1;
+    }
+
+  if (item_out->relay_host == NULL)
+    {
+      g_debug ("%s: mandatory relay_host field missing or not a string",
+               __func__);
+      return -1;
+    }
+
+  if (item_out->scanner_type == NULL)
+    {
+      g_debug ("%s: mandatory scanner_type field missing or not a string",
                __func__);
       return -1;
     }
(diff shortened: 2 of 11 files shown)


src/manage.c
```diff
@@ -92,6 +93,7 @@
 #include <gvm/util/versionutils.h>
 #include <gvm/gmp/gmp.h>
 
+
```
ozgen: redundant space

src/manage_scanner_relays.c
```diff
@@ -0,0 +1,371 @@
...
+  gvm_json_obj_check_int (json_item, "original_port",
+                          &item_out->original_port);
+  gvm_json_obj_check_str (json_item, "relay_host",
+                          &item_out->relay_host);
+  gvm_json_obj_check_int (json_item, "relay_port",
+                          &item_out->relay_port);
+  gvm_json_obj_check_str (json_item, "protocol",
+                          &item_out->protocol);
+
+  if (item_out->original_host == NULL)
```
ozgen, line 151: Maybe validating the port is not necessary if socket-based connections are supported. Should we also need to validate other fields, such as `relay_host`?

src/manage_scanner_relays.c
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+  if (update_all_scanner_relays_from_json (parsed_file, stat_buffer.st_mtime))
+    {
+      g_warning ("%s: could not update relays", __func__);
+      cJSON_Delete (parsed_file);
+      return -1;
+    }
+
+  cJSON_Delete (parsed_file);
```
ozgen, line 288: we need to add `g_free` for `file_contents`

src/manage_scanner_relays.c
```diff
@@ -0,0 +1,371 @@
...
+  if (relay_host == NULL || relay_port == NULL)
+    {
+      g_warning ("%s: output parameters must not be NULL", __func__);
+      return -1;
+    }
+
+  *relay_host = NULL;
+  *relay_port = 0;
+
+  if (! g_file_get_contents (relays_path, &file_contents, &file_size, &error))
```
ozgen, line 352: same here `file_contents` is not freed

src/manage_scanner_relays.h
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * @brief  Data structure describing an item from a scanner relays list.
+ */
+typedef struct {
+  char *original_host;  ///< The original host / socket path of the scanner
+  int original_port;    ///< Original port of the scanner
+  char *relay_host;     ///< Host or socket path of the relay
+  int relay_port;       ///< Port of the relay
+  char *protocol;       ///< Protocol string defining the scanner type.
```
ozgen: I would prefer to use `scanner_type` instead of `protocol` which causes misunderstanding because `protocol` is already used in the scanner context to refer to `http` or `https`, which makes this naming confusing.

src/manage_scanner_relays.c
```diff
@@ -0,0 +1,371 @@
...
+      g_warning ("%s: Failed to stat scanner relays file: %s",
+                 __func__, strerror (errno));
+      return -1;
+    }
+
+  db_mod_time = get_scanner_relays_db_update_time ();
+  if (stat_buffer.st_mtime <= db_mod_time)
+    {
+      g_debug ("%s: relays file not modified since last sync", __func__);
+      return -1;
```
ozgen: That looks like a normal no-op case, so returning success would be more appropriate.

src/manage_scanner_relays.c
```diff
@@ -0,0 +1,371 @@
...
+          && (parsed_item.original_port == 0
+              || parsed_item.original_port == original_port)
+          && (strcmp (original_host, parsed_item.original_host) == 0))
+        {
+          g_free (*relay_host);
+          *relay_host = g_strdup (parsed_item.relay_host);
+          *relay_port = parsed_item.relay_port;
+        }
+    }
+  return 0;
```
ozgen, line 390: I think we need to cleanup JSON object before returning success

src/manage_sql_scanner_relays.c
```diff
@@ -0,0 +1,144 @@
...
+          NULL);
+}
+
+/**
+ * @brief Perform actions at the start of updating all scanner relays.
+ *
+ * The relay hosts and ports for all scanners are reset.
+ */
+void
+update_all_scanner_relays_start ()
```
ozgen, line 60: This means only scanners present in the file are reassigned, any scanners not listed end up with cleared relay settings. That may be intended, but it should be documented very explicitly because it is broader than just updating matching entries

src/manage_scanner_relays.c
```diff
@@ -0,0 +1,371 @@
...
+  GHashTableIter scanners_iter;
+  gchar *scanner_id;
+
+  if (extract_relays_fields_from_root_cjson (parsed_file, &relays))
+    return -1;
+
+  updated_scanners
+    = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
+
+  update_all_scanner_relays_start ();
```
ozgen, line 200: I think we need transaction handling or rollback mechanism here, if one of them failed, we should rollback all

ozgen (approved):

src/manage_scanner_relays.c
```diff
@@ -0,0 +1,391 @@
...
+      if (extract_relay_fields_from_json_item (json_item, &parsed_item))
+        continue;
+      if (scanner_type_matches_relay (scanner_type,
+                                      parsed_item.scanner_type)
+          && (parsed_item.original_port == 0
+              || parsed_item.original_port == original_port)
+          && (strcmp (original_host, parsed_item.original_host) == 0))
+        {
+          g_free (*relay_host);
+          *relay_host = g_strdup (parsed_item.relay_host);
```
mattmundell, line 385: Hey @timopollmeier This is allocating to *relay_host:
 - I think the doc should mention that the caller must free *relay_host
 - I think it's leaking in `create_scanner`: the other cases of `used_relay_host` come from `relay_host` which is a const passed in but `file_relay_host` is allocated.
 - I think `modify_scanner` is also leaking in the success case (but it was already, before this PR).

mattmundell: Fixed in /pull/2917.

PR #2876, opened 2026-04-15
[GEA-1613] DB table and helper functions for Security Intelligence Export

robindittmar:
## What

Add SQL table `export_reports` and queue reports for export on scan completion

## Why

Make Security Intelligence export possible

## References

GEA-1613


src/manage_pg.c
```diff
@@ -3114,6 +3115,19 @@ create_tables ()
        "  value text,"
        "  UNIQUE (report_config, name));");
 
+  if (feature_enabled(FEATURE_ID_SECURITY_INTELLIGENCE_EXPORT))
```
ozgen: personally, we could remove the feature flag checks while creating tables, for the consistency I always prefer to exists in the db even if feature flag disabled.

src/manage_pg.c
```diff
@@ -3546,6 +3560,19 @@ create_tables ()
        "                     'tls_certificate_origins',"
        "                     'origin_id, origin_type')");
 
+  if (feature_enabled(FEATURE_ID_SECURITY_INTELLIGENCE_EXPORT))
```
ozgen: same here

src/manage_sql_report_exports.c
```diff
@@ -0,0 +1,174 @@
...
+ *        all exports that are due now
+ *
+ * @param[out] iterator iterator to initialize
+ *
+ * @return 0 on success
+ */
+int
+init_report_export_iterator_due_exports (iterator_t *iterator)
+{
+  init_iterator (iterator, "SELECT id, report_id, status, reason,"
```
ozgen: nit: for readability, status could be checked e.g.:

```

init_iterator (iterator, "SELECT id, report_id, status, reason,"
                              "    retry_count, next_retry_time,"
                              "    creation_time, modification_time"
                              " FROM report_exports"
                              " WHERE next_retry_time <= m_now()"
                              " AND status IN ('report_export_requested',"
                              "                'report_export_failed')");
```

src/manage_sql.c
```diff
@@ -13390,6 +13391,13 @@ set_report_scan_run_status (report_t report, task_status_t status)
        report);
   if (setting_auto_cache_rebuild_int ())
     report_cache_counts (report, 0, 0, NULL);
+
+  if (feature_enabled (FEATURE_ID_SECURITY_INTELLIGENCE_EXPORT)
```
ozgen: I think this is the wrong place for this logic.

`set_report_scan_run_status()` looks like a low-level status update helper, but this change also makes it trigger export-related workflow when the status is set to `TASK_STATUS_DONE`. That adds a hidden side effect and bad for readability.

It would be clearer to queue the export where the report is marked as `DONE`, or in the higher-level scan completion handling.

src/manage_sql_report_exports.c
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * @brief Get the creation time from a report export iterator.
+ *
+ * @param[in]  iterator  Iterator.
+ *
+ * @return Creation time, or 0 if iteration is complete. Freed by
+ *         cleanup_iterator.
+ */
+long long
```
ozgen: Can we use `time_t` instead of `long long` here? Since this represents a timestamp, `time_t` would be more appropriate and clearer.

src/manage_sql_report_exports.c
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * @brief Get the modification time from a report export iterator.
+ *
+ * @param[in]  iterator  Iterator.
+ *
+ * @return Modification time, or 0 if iteration is complete. Freed by
+ *         cleanup_iterator.
+ */
+long long
```
ozgen: use `time_t`  instead of` long long `

src/manage_sql_report_exports.c
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * @brief Get the next retry time from a report export iterator.
+ *
+ * @param[in]  iterator  Iterator.
+ *
+ * @return Next retry time, or 0 if iteration is complete. Freed by
+ *         cleanup_iterator.
+ */
+long long
```
ozgen: use `time_t` instead of `long long`

src/manage_sql_report_exports.c
```diff
@@ -0,0 +1,174 @@
...
+          "   SET status = $1, reason = $2"
+          " WHERE report_id = $3",
+          SQL_STR_PARAM (status),
+          SQL_STR_PARAM (reason),
+          SQL_INT_PARAM (report),
+          NULL);
+}
+
+void
+set_report_export_next_retry_time (report_t report, long long next_retry_time)
```
ozgen: use `time_t` instead of `long long`

src/manage_sql_report_exports.c
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * @brief Get the retry count from a report export iterator.
+ *
+ * @param[in]  iterator  Iterator.
+ *
+ * @return Retry count, or 0 if iteration is complete. Freed by
+ *         cleanup_iterator.
+ */
+long long
```
ozgen: `long long` feels a bit unusual for a retry count. Is there a reason not to use a more specific integer type here, since this value is just a small counter?

src/manage_sql_report_exports.c
ozgen: Should we add a maximum retry limit here? Otherwise we might keep retrying indefinitely. A configurable threshold (e.g. default 10 via env variable) could make sense.

src/manage_report_exports.h
```diff
@@ -0,0 +1,57 @@
...
+report_t
+report_export_iterator_report_id (iterator_t *iterator);
+
+const char *
+report_export_iterator_status (iterator_t *iterator);
+
+const char *
+report_export_iterator_reason (iterator_t *iterator);
+
+long long
```
ozgen: Do we want to keep `long long` for all of these return types?

`retry_count` seems like it could use a narrower counter type, while `next_retry_time`, `creation_time`, and `modification_time` may be better expressed as `time_t` if they represent timestamps. That would make the intent clearer.

src/manage_report_exports.h
```diff
@@ -0,0 +1,57 @@
...
+set_report_export_retry_count (report_t report, int retry_count);
+
+
+int
+init_report_export_iterator_due_exports (iterator_t *iterator);
+
+report_t
+report_export_iterator_report_id (iterator_t *iterator);
+
+const char *
```
ozgen: could we use `gchar` instead of `char`?

src/manage_report_exports.h
```diff
@@ -0,0 +1,57 @@
...
+report_export_iterator_status (iterator_t *iterator);
+
+const char *
```
ozgen: could we use `gchar` instead of `char`?

ozgen (changes requested):

src/manage_sql_report_exports.c
robindittmar: I changed the SQL to your more readable version.

Regarding the retry limit: In theory we would retry indefinitely, however the retry_count will cause the next_retry_time to increase exponentially, right? So at some point we'll add more than like ten years or so to the next_retry_time, which imo is good enough and less complex

src/manage_sql.c
robindittmar: I moved the logic to all the places where `set_report_scan_run_status(..., TASK_STATUS_DONE)` is called instead

src/manage_sql_report_exports.c
```diff
@@ -0,0 +1,224 @@
...
+}
+
+/**
+ * @brief Update report export next retry time
+ *
+ * @param  report           The report ID that is in export queue
+ * @param  next_retry_time  The new next retry time
+ */
+void
+set_report_export_next_retry_time (report_t report, long long next_retry_time)
```
ozgen: still `long long` used in the function parameter

src/manage_sql_report_exports.c
```diff
@@ -0,0 +1,224 @@
...
+/**
+ * @brief Checks if the automatic export to security intelligence is enabled
+ *        for given report
+ *
+ * @param  report  The report to check
+ *
+ * @return TRUE if enabled, FALSE otherwise
+ */
+gboolean
+export_enabled_for_report (report_t report)
```
ozgen: The naming is a bit confusing here. This does not really check whether export is enabled *for the report*, but whether it is enabled in the owning user's settings.

Would a name like `export_enabled_for_report_owner()` be clearer?

src/manage_sql_report_exports.c
ozgen: That makes sense, and exponential backoff definitely reduces the practical impact.

My concern is just that this still does not define a clear stopping point. A max retry count would make the retry policy more explicit, whereas with backoff alone the job remains retryable indefinitely in theory.

 this might be worth aligning on at team level. could you bring this up for discussion?

src/manage_osp.c
```diff
@@ -1319,6 +1326,10 @@ handle_osp_scan_end (task_t task, int handle_progress_rc, gboolean discovery)
       asset_snapshots_target (global_current_report, task, discovery);
       set_task_run_status (task, TASK_STATUS_DONE);
       set_report_scan_run_status (global_current_report, TASK_STATUS_DONE);
+      if (feature_enabled (FEATURE_ID_SECURITY_INTELLIGENCE_EXPORT))
```
ozgen, line 1325: We may need to extend this functionality to other scanner types as well, like openvasd, Agent Control, and potentially container scanning.

src/manage_osp.c
```diff
@@ -1113,6 +1116,10 @@ handle_osp_scan_start (task_t task, target_t target, const char *scan_id,
       report_add_result (global_current_report, result);
       set_task_run_status (task, TASK_STATUS_DONE);
       set_report_scan_run_status (global_current_report, TASK_STATUS_DONE);
+      if (feature_enabled (FEATURE_ID_SECURITY_INTELLIGENCE_EXPORT))
```
ozgen: This might be the wrong place because at the end we return `-1`. We should enqueue when the scan completed successfully.

src/manage_osp.c
ozgen: Just to be sure, we have a queue mechanism for OSP scan tasks. Could you check whether this case is covered?

ozgen (changes requested):

src/manage_osp.c
robindittmar: You're right. There was another place in agent controllers just like that.

No idea why the task/report is marked as "done" in an error case, but I removed the queuing from those places

src/manage_osp.c
robindittmar: I don't really understand your concern?
The queuing is happening in the following functions:
```
cve_scan_handler
openvasd_scan_handler
agent_controller_scan_handler
container_image_scan_handler
handle_osp_scan_end
```
Which do you think is missing?

ozgen: >We may need to extend this functionality to other scanner types as well, like openvasd, Agent Control, and potentially container scanning. 

To clarify, the current change seems to only cover OSP scan tasks.

However, other scanner types follow different completion paths:

* Agent Controller scans are handled in `fork_agent_controller_scan_handler`
* OpenVASD scans are handled in `fork_openvasd_scan_handler`
* Container image scans are handled in `fork_container_image_scan_handler`

you have already covered these cased with the new changes in this PR.



>Just to be sure, we have a queue mechanism for OSP scan tasks. Could you check whether this case is covered?

for this concern, we currently have two different queue mechanisms, and they serve different purposes:

1. There is a queue mechanism for OSP scan tasks (when enabled by the user). This controls how scans are scheduled and executed, and is handled in `manage_scan_handler.c` (implemented by Timo). Queued scans are processed via `update_osp_scan` in `manage_osp.c`.

2. The queue introduced here is for report export, which is a separate concern.

I think these two might be getting mixed up in this context. The existing OSP scan queue does not automatically cover the export flow.

So we likely need to ensure that the export queueing logic is also integrated consistently across the relevant scan handlers / flows.

src/manage_container_image_scanner.c
```diff
@@ -1011,6 +1012,10 @@ fork_container_image_scan_handler (task_t task,
       report_add_result (global_current_report, result);
       set_task_run_status (task, TASK_STATUS_DONE);
       set_report_scan_run_status (global_current_report, TASK_STATUS_DONE);
+      if (feature_enabled (FEATURE_ID_SECURITY_INTELLIGENCE_EXPORT))
```
ozgen: please also remove your changes because the scan is not completed successfully

src/manage.c
```diff
@@ -7493,6 +7498,10 @@ fork_openvasd_scan_handler (task_t task, target_t target, int from,
       report_add_result (global_current_report, result);
       set_task_run_status (task, TASK_STATUS_DONE);
       set_report_scan_run_status (global_current_report, TASK_STATUS_DONE);
+      if (feature_enabled (FEATURE_ID_SECURITY_INTELLIGENCE_EXPORT))
```
ozgen: This might be the wrong place because at the end we return -1. We should enqueue when the scan completed successfully.

src/manage_osp.c
robindittmar: After talking to Timo, no queuing in `update_osp_scan ()` is required, as the task status is only set to `TASK_STATUS_RUNNING` in there.

This ultimately causes `handle_osp_scan_end ()` to be called, which means that this case is already covered in this PR

src/manage_sql_report_exports.c
robindittmar: Since the team was in favor of an explicit policy, I added the environment variable `GVMD_REPORT_EXPORT_MAX_RETRIES` (default: 10) which is used to exclude entries with a retry_count bigger than or equal to that value

src/manage_sql_report_exports.c
```diff
@@ -112,13 +112,23 @@ set_report_export_retry_count (report_t report, int retry_count)
...
 init_report_export_iterator_due_exports (iterator_t *iterator)
 {
-  init_iterator (iterator, "SELECT id, report_id, status, reason,"
-                              "    retry_count, next_retry_time,"
-                              "    creation_time, modification_time"
-                              " FROM report_exports"
-                              " WHERE   next_retry_time < m_now()"
-                              " AND     status IN ('report_export_requested',"
-                              "                    'report_export_failed')");
+  static int max_retry_count = 0;
```
ozgen, line 115: Instead of caching this value, should we read it dynamically from the environment variable? This would allow changes without restarting gvmd. Also this environment variable is currently undocumented and should be properly documented (including its purpose, accepted values, and default behavior).

ozgen: I am not sure which place is suitable for this, maybe you should ask the team?

ozgen: Could you confirm with GOS-DEV if this approach is convenient for them?

robindittmar: I have
- Removed the `static` keyword, now the env variable is read each time you initialize the iterator and query the DB
- Added an additional config section/key to `gvmd.conf`
- Documented the config value + env variable in `README.md`
- Asked GOS-DEV, and they said they can "live with it" :D

ozgen (approved):

PR #2886, opened 2026-04-22
Add/get report applications

ozgen:
## What

* Add report applications support with severity aggregation in the management and SQL layers
* Implement `get_report_applications` GMP command
* Return application name, host count, occurrence count, severity, and threat
* Support `details` flag to return either full application data or only count
* Add GMP documentation with request/response examples

## Why

* Enable efficient retrieval of application-level insights from reports
* Avoid over-fetching data by supporting count-only responses when details are not needed
* Provide a dedicated endpoint instead of relying on full `get_reports` parsing
* Improve frontend integration with structured and ready-to-use application data

## References

GEA-1710


src/manage_report_applications.c
```diff
@@ -0,0 +1,135 @@
...
+      double *severity_ptr;
+      report_application_t app = report_application_new ();
+
+      app->application_name = g_strdup (
+        report_app_iterator_application_name (&report_apps));
+      app->hosts_count = report_app_iterator_host_count (&report_apps);
+      app->occurrences = report_app_iterator_occurrences (&report_apps);
+      app->severity_double = 0.0;
+
+      if (!str_blank (app->application_name))
```
timopollmeier: If the app struct is only used if the application name isn't blank, it would make sense to check the name directly from the iterator first before allocating the struct and copying the string.

timopollmeier (approved):

PR #2889, opened 2026-04-23
[GEA-1614] Add report export scheduler

robindittmar:
## What

Add a report export scheduler, which regularly exports all due reports.

## Why

Add functionality to automatically export reports to OpenVAS Security Intelligence

## References

GEA-1614


src/manage_report_export_scheduler.c
```diff
@@ -0,0 +1,178 @@
+/* Copyright (C) 2025 Greenbone AG
```
ozgen:
```suggestion
/* Copyright (C) 2026 Greenbone AG
``

File: src/manage_report_export_scheduler.c [hunk 2]
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * @brief  Run report export scheduler, which fetches all due exports
+ *         and tries to export them accordingly
+ *
+ * @return 0 on success, -1 on failure
+ */
+int
+manage_report_export_scheduler ()
+{
+  iterator_t it;
```
ozgen: `it` is not meaningful for me could we renamed as `report_exports`?

src/gvmd.c
```diff
@@ -1529,6 +1530,81 @@ fork_queued_task_actions ()
...
+        }
+
+      if (manager_socket_2 > -1)
+        {
+          close (manager_socket_2);
+          manager_socket_2 = -1;
+        }
+
+      /* Run report export scheduler */
+      init_report_export_scheduler_from_config ();
```
ozgen: If the intention is to always use the latest config values, should `init_report_export_scheduler_from_config()` be called inside `manage_report_export_scheduler()` instead?

src/manage_report_export_scheduler.h
```diff
@@ -0,0 +1,37 @@
+/* Copyright (C) 2026 Greenbone AG
+*
```
ozgen:
```suggestion
 * 
``

File: src/manage_report_export_scheduler.c [hunk 5]
```diff
@@ -0,0 +1,178 @@
...
+
+  g_debug ("%s: iterating over due exports", __func__);
+  init_report_export_iterator_due_exports (&it);
+
+  while (next (&it))
+    {
+      report_t report = report_export_iterator_report_id (&it);
```
ozgen: Maybe we can move this logic into a helper like `process_report_export(report, retry_count)` to handle each export in one place.
It would make the loop cleaner and easier to maintain.

Also, we could wrap the final status/retry updates in a single transaction

e.g:

```
mark STARTED
run export
transaction {
  update final status
  update retry time
  update retry count
}
```
This avoids keeping a transaction open while `export_report()` is running.

ozgen (changes requested):

src/manage_report_export_scheduler.c
rdttmr: Done

src/manage_report_export_scheduler.c
rdttmr: Done

src/gvmd.c
rdttmr: Moved it as you suggested

src/manage_report_export_scheduler.h
rdttmr: Done

src/manage_report_export_scheduler.c
rdttmr: Done.

Moved the inner loop logic to its own function called `process_report_export (...)` and the final status update is now wrapped in a transaction.

src/manage_report_export_scheduler.c
```diff
@@ -0,0 +1,187 @@
...
+  gvmd_config_resolve_int ("GVMD_REPORT_EXPORT_RETRY_MULTIPLIER", TRUE,
+                           retry_multiplier, &multiplier);
+  gvmd_config_resolve_int ("GVMD_REPORT_EXPORT_RETRY_MAX_DELAY", TRUE,
+                           retry_max_delay, &max_delay);
+
+  return time (NULL)
+         + MIN (base_delay * (long) pow (multiplier, retry_count), max_delay);
+}
+
+static void
```
ozgen, line 176: Can you add brief comment in this `process_report_export` func?

src/manage_report_export_scheduler.c
```diff
@@ -0,0 +1,187 @@
...
+int
+manage_report_export_scheduler ()
+{
+  iterator_t report_exports;
+
+  init_report_exports_from_config ();
+  init_report_export_scheduler_from_config ();
+
+  g_debug ("%s: iterating over due exports", __func__);
+  init_report_export_iterator_due_exports (&report_exports);
```
ozgen: **One concern:** `STARTED` exports are not considered by the due-export iterator. If a worker dies after setting the status to `STARTED`, the export may stay stuck forever. Should we add handling for stale `STARTED` exports?

src/manage_report_export_scheduler.h
```diff
@@ -0,0 +1,37 @@
...
+} export_report_result_t;
+
+void
+init_report_export_scheduler_from_config ();
+
+int
+manage_report_export_scheduler ();
+
+export_report_result_t
+export_report (report_t report);
```
ozgen: Do we need to expose `export_report()` in the header?

src/manage_report_export_scheduler.c
rdttmr: Good point!
I have added some logic to handle stale exports.

src/manage_report_export_scheduler.c
rdttmr: Done

src/manage_report_export_scheduler.h
rdttmr: I cleaned up the header :)

ozgen (approved):

PR #2890, opened 2026-04-23
Change: remove unused macros

mattmundell:
## What

Remove unused macros.

## Why

Cleaner.

## Testing

Compile with `-Wunused-macros`, warnings are gone.

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -4882,17 +4882,6 @@ set_read_over (gmp_parser_t *gmp_parser)
     }                                                           \
   break
 
-/**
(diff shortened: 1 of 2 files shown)


bjoernricks (approved):
Should we add this compiler flag by default?

mattmundell:
> Should we add this compiler flag by default?

It's tricky because it warns about valid macros like G_LOG_DOMAIN and _XOPEN_SOURCE.

I think it would be better as something that the CI checks, with filtering of these valid cases.

PR #2892, opened 2026-04-27
docs: add initial terms and keywords to glossary

ozgen:
## Checklist

- [ ] Tests


docs/glossary.md
```diff
@@ -0,0 +1,106 @@
+## Glossary
+
+| Term                             | Definition                                                                                                                                               |
+|----------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **Target**                       | A collection of hosts (IPs, domains, or ranges) that you want to scan. It also includes scan-related options like excluded hosts and alive test methods. |
+| **Asset**                        | A long-term representation of a host. It combines data from multiple scans and keeps history like last seen time, severity counts, and identifiers.      |
```
rdttmr: I would prefer something more decoupled and general, like:

`An asset is any digital or physical element, like a server, IP address, or software, that a vulnerability scanner can identify, scan or report on. Essentially, anything that can be assessed for security risks.`

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Report Host**                  | A host entry inside a report, including its findings.                                                                                                    |
+| **Host Asset**                   | The asset representation of a host, maintained across multiple scans.                                                                                    |
+| **User**                         | A system account that interacts with the platform.                                                                                                       |
+| **Role**                         | A collection of permissions assigned to users.                                                                                                           |
+| **Permission**                   | A rule defining allowed actions.                                                                                                                         |
+| **Group**                        |                                                                                                                                                          |
```
rdttmr: Suggestion:

`A collection of users. Groups inherit Roles and Permissions to all Users within a Group.`

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Alert**                        | A notification triggered by events (e.g., scan finished, high severity found).                                                                           |
```
rdttmr: I think it makes sense to move the examples for events to the `Alert Event` definition

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Agent Installer**              | Tool or package used to deploy agents.                                                                                                                   |
+| **Port List**                    | A predefined set of ports to scan.                                                                                                                       |
+| **Container Scanning**           | Scanning container images (e.g., Docker) for vulnerabilities.                                                                                            |
+| **OCI Image Target**             | A container image defined as a scan target.                                                                                                              |
+| **OCI Image**                    | A container image following OCI standards (e.g., Docker image).                                                                                          |
+| **Sensor**                       | A scanner that works on specific network.                                                                                                                |
```
rdttmr: Based on this definition, to me it's not really clear what a Sensor actually is. And I don't know what a sensor is, so I have no proposal how to change it 😁

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Master**                       | The main scanner of the system.                                                                                                                          |
+| **Feed Key**                     | A key used to authenticate and download feed data.                                                                                                       |
```
rdttmr: Maybe it makes sense to add the word "license" here somewhere?
To make it clear, that this is something you can only obtain by contacting Greenbone? Just throwing ideas out there.

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Note**                         | A comment added to a result or resource.                                                                                                                 |
+| **Tag**                          | A label used to organize and categorize resources.                                                                                                       |
+| **Filter**                       | A query rule used to search or limit data (e.g., severity > 7).                                                                                          |
+| **User Setting**                 | Preferences specific to a user account.                                                                                                                  |
+| **Resource / Entity Type**       | A category of objects (e.g., task, report, asset).                                                                                                       |
+| **TLS Certificate**              | A TLS/SSL certificate discovered during a scan on a host or service (e.g., HTTPS).                                                                       |
```
rdttmr: In our domain, do we really only talk about TLS certificates when "discovered during a scan"?

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Security Information**         | General security-related data collected or processed by the system.                                                                                      |
+| **Feed Owner**                   | The user responsible for maintaining the feed.                                                                                                           |
+| **Agent Owner**                  | The user  responsible for an agent.                                                                                                                      |
+| **Admin**                        | A user with elevated permissions.                                                                                                                        |
+| **Super Admin**                  | A user with full system control.                                                                                                                         |
+| **Feature Flag**                 | A toggle used to enable or disable features without redeploying.                                                                                         |
```
rdttmr: I think we can cut out "without redeploying", as it is not really relevant for the definition.
Maybe instead, if my assumptions is correct, we could specify that this is used only for features that are not yet polished?

docs/glossary.md
timopollmeier: That sounds like a good definition to me. Maybe we should also mention that in some contexts like other products it can specifically refer to what we call a host asset.

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
+## Glossary
+
+| Term                             | Definition                                                                                                                                               |
+|----------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **Target**                       | A collection of hosts (IPs, domains, or ranges) that you want to scan. It also includes scan-related options like excluded hosts and alive test methods. |
+| **Asset**                        | A long-term representation of a host. It combines data from multiple scans and keeps history like last seen time, severity counts, and identifiers.      |
+| **Task**                         | A scan definition that links a target, scan configuration, and schedule. It controls how and when scans are executed.                                    |
```
timopollmeier: This term is somewhat ambiguous and can also refer to a task specifically for vulnerability scans in contrast to audits which are about compliance scans.

ozgen: that's a good point!

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Scan**                         | A single execution of a task. It runs vulnerability tests against the target and produces results.                                                       |
```
timopollmeier: This can also include compliance tests:
```suggestion
| **Scan**                         | A single execution of a task. It runs vulnerability or compliance tests against the target and produces results.                                                       |
```

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Asset**                        | A long-term representation of a host. It combines data from multiple scans and keeps history like last seen time, severity counts, and identifiers.      |
+| **Task**                         | A scan definition that links a target, scan configuration, and schedule. It controls how and when scans are executed.                                    |
+| **Scan**                         | A single execution of a task. It runs vulnerability or compliance tests against the target and produces results.                                                       |
+| **Result**                       | A single finding from a scan, such as a detected vulnerability or informational message.                                                                 |
+| **Host Detail**                  | Additional structured information about a host discovered during a scan (e.g., OS, services, applications, ports).                                       |
+| **Host**                         | A single machine (IP address or hostname) within a target that is scanned individually.                                                                  |
```
timopollmeier:
```suggestion
| **Host**                         | A single physical or virtual network device, usually identified by its IP address and/or hostname.                                                                  |
```

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Port**                         | A network port (TCP/UDP) on a host that can be scanned for open services or vulnerabilities.                                                             |
+| **Credential**                   | Authentication data (e.g., username/password, SSH key, SNMP, Kerberos) used to perform deeper, authenticated scans.                                      |
+| **Scan Config**                  | A predefined set of vulnerability tests (VTs) and settings that define what checks are executed during a scan.                                           |
```
timopollmeier:
```suggestion
| **Scan Config**                  | A predefined set of vulnerability tests (VTs) and preferences that define what checks are executed during a scan and how.                                           |
```

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Preference**                   | Scanner settings that adjust how specific tests or the scanner behave.                                                                                   |
+| **Scanner**                      | The application that performs the actual vulnerability checks (e.g., OpenVAS, openvasd, agent-controller, container-scanning, sensors).                  |
```
timopollmeier:
```suggestion
| **Scanner**                      | An instance of the application that performs the actual vulnerability checks (e.g., OpenVAS, openvasd, agent-controller, container-scanning, sensors) or that application in general.                  |
```

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Scanner Type**                 | The communication type or implementation of the scanner (e.g., ospd-openvas, openvasd, agent-controller, container-scanning, sensors).                   |
+| **Alive Test**                   | A method used before scanning to check if a host is reachable (e.g., ICMP ping, TCP connect).                                                            |
+| **Report**                       | The stored output of a scan, including all results, summaries, and metadata.                                                                             |
```
timopollmeier:
```suggestion
| **Report**                       | The stored output of a scan, including all results, host details, summaries, and metadata.                                                                             |
```

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Compliance Report**            | A report focused on compliance checks against policies or standards (e.g., CIS benchmarks).                                                              |
+| **CIS Benchmarks**               | Simple security rules that tell you how to set up a system safely (like a checklist).                                                                    |
+| **CIS**                          | Center for Internet Security.                                                                                                                            |
+| **Schedule**                     | A configuration that defines when a task runs automatically in a specific period of time.                                                                |
+| **Audit**                        |                                                                                                                                                          |
```
timopollmeier:
```suggestion
| **Audit**                        | A subtype of task for defining the parameters (target, scan config, etc.) for compliance scans.|
```

docs/glossary.md
timopollmeier: We should also mention that tasks are also used to group reports and add an entry for "Import Task", which should mention that it was previously named "Container Task" before the introduction of OCI image scans.

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **VT Metadata**                  | Information about a test (OID, name, severity, references like CVE/CPE).                                                                                 |
+| **QoD**                          | Quality of Detection (0–100%). Indicates how reliable a finding is. Higher means more confidence.                                                        |
+| **Score**                        | A numeric value used to measure or rate something based on defined criteria.                                                                             |
+| **CVSS**                         | A standardized score (0.0–10.0) representing how severe a vulnerability is based on impact and exploit.                                                  |
+| **EPSS**                         | A score that estimates the likelihood that a vulnerability will be exploited in the real world.                                                          |
+| **Severity**                     | A simplified category derived from scores (Low, Medium, High, Critical).                                                                                 |
```
timopollmeier: Severity can refer to both the numeric score and level (Low to critical). Without specification it is usually the score.
Perhaps we should have separate entries for "Severity Level" (using the current category defintion) and "Severity Score", which is currently a CVSS score.

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **NVT Family**                   | A logical group of related vulnerability tests.                                                                                                          |
+| **VT Metadata**                  | Information about a test (OID, name, severity, references like CVE/CPE).                                                                                 |
+| **QoD**                          | Quality of Detection (0–100%). Indicates how reliable a finding is. Higher means more confidence.                                                        |
+| **Score**                        | A numeric value used to measure or rate something based on defined criteria.                                                                             |
+| **CVSS**                         | A standardized score (0.0–10.0) representing how severe a vulnerability is based on impact and exploit.                                                  |
+| **EPSS**                         | A score that estimates the likelihood that a vulnerability will be exploited in the real world.                                                          |
```
timopollmeier: Technically CVSS and EPSS are the scoring systems but in some contexts the terms are used as shorthand for the scores.

docs/glossary.md
ozgen: I like this idea!

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Severity**                     | A simplified category derived from scores (Low, Medium, High, Critical).                                                                                 |
+| **Threat**                       | Another representation of risk level, often aligned with severity.                                                                                       |
+| **CVE**                          | A unique public identifier for a known vulnerability (e.g., CVE-2024-12345).                                                                             |
+| **CPE**                          | A standardized way to identify software, hardware, or operating systems.                                                                                 |
```
timopollmeier: I think links to the specifications would be nice.

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **SBOM**                         | Software Bill of Materials. A structured list of all components (libraries, packages, dependencies) inside an application or container image.            |
+| **SCAP Data**                    | Standardized security data used for automated vulnerability and compliance checks.                                                                       |
+| **CERT Advisory**                | Security advisories published by organizations about vulnerabilities or threats.                                                                         |
```
timopollmeier:
```suggestion
| **CERT Advisory**                | Security advisories published by organizations (Computer Emergency Response Teams) about vulnerabilities or threats.                                                                         |
```

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Closed CVE**                   | A vulnerability that is resolved, mitigated, or no longer relevant for a host.                                                                           |
+| **Feed**                         | External data source that provides updated VTs, CVEs, and security information.                                                                          |
```
timopollmeier:
```suggestion
| **Feed**                         | External data source that provides updated VTs, predefined data objects, and security information like CVEs.                                                                        |
```

docs/glossary.md
timopollmeier: I think `Roles and Permissions are inhertited by all Users within a Group.` would be more correct.

docs/glossary.md
timopollmeier: We should also separate conditions like the severity threshold mentioned as an example from events.

docs/glossary.md
timopollmeier: I'd define it as something like this:
```suggestion
| **Sensor**                       |  A scanner running on a device other than the local GVM instance (master) to allow scanning a network that is otherwise not accessible.                                                                                                                |
```
Which also reminds me that we should define "GVM" (Greenbone Vulnerability Manager) and components (gvmd, gsad, GSA).

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Port List**                    | A predefined set of ports to scan.                                                                                                                       |
+| **Container Scanning**           | Scanning container images (e.g., Docker) for vulnerabilities.                                                                                            |
+| **OCI Image Target**             | A container image defined as a scan target.                                                                                                              |
+| **OCI Image**                    | A container image following OCI standards (e.g., Docker image).                                                                                          |
+| **Sensor**                       | A scanner that works on specific network.                                                                                                                |
+| **Master**                       | The main scanner of the system.                                                                                                                          |
```
timopollmeier:
```suggestion
| **Master**                       | A GVM instance controlling external scanners (sensors) and scanners running on the same device.                                                                                                                          |
```

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Filter**                       | A query rule used to search or limit data (e.g., severity > 7).                                                                                          |
+| **User Setting**                 | Preferences specific to a user account.                                                                                                                  |
+| **Resource / Entity Type**       | A category of objects (e.g., task, report, asset).                                                                                                       |
+| **TLS Certificate**              | A TLS/SSL certificate discovered during a scan on a host or service (e.g., HTTPS).                                                                       |
+| **Operating System**             | The OS detected on a host (e.g., Linux, Windows).                                                                                                        |
+| **Security Information**         | General security-related data collected or processed by the system.                                                                                      |
```
timopollmeier: This is actually more a specific term to refer to SCAP data, CERT advisories and VT metadata.

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Agent Owner**                  | The user  responsible for an agent.                                                                                                                      |
+| **Admin**                        | A user with elevated permissions.                                                                                                                        |
+| **Super Admin**                  | A user with full system control.                                                                                                                         |
+| **Feature Flag**                 | A toggle used to enable or disable features without redeploying.                                                                                         |
+| **Performance Report**           | A report showing system or scan performance metrics.                                                                                                     |
+| **GXR Report**                   | A specialized report format used within the system.                                                                                                      |
```
timopollmeier: I think we should remove this if we don't want to add entries for all report formats.

docs/glossary.md
```diff
@@ -0,0 +1,106 @@
...
+| **Pheme**                        | An internal or system-specific component for report.                                                                                                     |
+| **Authentication Method**        | The way users log in (password, jwt, LDAP, RADIUS, etc.).                                                                                                |
+| **Trashcan**                     | A temporary storage for deleted resources before permanent removal.                                                                                      |
+| **Alert Condition**              | A rule that defines when an alert should trigger.                                                                                                        |
+| **Alert Event**                  | The event that activates an alert.                                                                                                                       |
```
timopollmeier: For "Alert Condition" it should be clarified that it needs to be met in addition to the event. Giving examples would also be helpful.

bjoernricks (approved):

PR #2897, opened 2026-05-04
Fix: fall back to global setting, else sql_int_ps aborts

mattmundell:
## What

In `export_enabled_for_report_owner` fall back to the global setting of `SETTING_UUID_SECURITY_INTELLIGENCE_EXPORT` to ensure that a row will be returned.

## Why

A row must be returned, else `sql_int_ps` aborts.

## Testing

Start gvmd with `GVMD_ENABLE_SECURITY_INTELLIGENCE_EXPORT=1`. Run a CVE scan. Look for abort in logs.

### DIFF
--- a/src/manage_sql_report_exports.c
+++ b/src/manage_sql_report_exports.c
@@ -27,12 +27,14 @@ gboolean
 export_enabled_for_report_owner (report_t report)
 {
   return !!sql_int_ps (
-    "SELECT s.value FROM reports AS r"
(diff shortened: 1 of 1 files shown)


mattmundell:
This check (both the original and the fixed versions) bypasses our ACL. It would be nice to use the settings API instead, but that's a larger change because `setting_value_int` etc use `current_credential` (and here we're using the report owner).

ozgen (approved):

PR #2899, opened 2026-05-05
Add/get report closed cves

ozgen:
## What

- Added SQL and manage-layer helpers for report closed CVEs.
- Added counting logic for closed CVEs.
- Added the new `get_report_closed_cves` GMP command.
- Added GMP command documentation for `get_report_closed_cves`.
- Verified the SQL query manually with a test report.

## Why

- To provide closed CVE data through a dedicated report command.
- To avoid relying on full get_report host details parsing for closed CVEs.
- To make closed CVE loading more modular and easier to integrate in GSA.

## References

GEA-1710


src/gmp_report_closed_cves.c
```diff
@@ -0,0 +1,237 @@
...
+                             attribute_names,
+                             attribute_values);
+
+  if (find_attribute (attribute_names, attribute_values,
+                      "report_id", &attribute))
+    {
+      get_report_closed_cves_data.report_id = g_strdup (attribute);
+
+      get_data_set_extra (&get_report_closed_cves_data.get, "report_id",
+                          g_strdup (attribute));
```
mattmundell: This should just pass attribute, else it leaks (get_data_set_extra is doing g_strdup on 3rd arg).

Probably a copy past error, because it's like this in other places too.

ozgen: Thanks, you are right. Since `get_data_set_extra()` already does `g_strdup()` internally, we should pass `attribute` directly here. I will fix this and check the similar places too.

timopollmeier (dismissed):

src/gmp_report_closed_cves.c
```diff
@@ -0,0 +1,237 @@
...
+{
+  const gchar *attribute;
+
+  get_data_parse_attributes (&get_report_closed_cves_data.get,
+                             "report_closed_cve",
+                             attribute_names,
+                             attribute_values);
+
+  if (find_attribute (attribute_names, attribute_values,
+                    "report_id", &attribute))
```
mattmundell:
```suggestion
                     "report_id", &attribute))
```

timopollmeier (approved):

PR #2902, opened 2026-05-06
Add/get report cves

ozgen:
## What

- Added the new `get_report_cves` GMP command.
- Added logic to return CVEs from report results.
- Each CVE includes host, CVE name, NVT info, severity and threat.
- Added support for `details=0` count-only responses.
- Added GMP docs and examples for the new command.

## Why

- To load report CVE data through a dedicated command.
- To reduce dependency on the full `get_report` response.
- To keep CVE handling consistent with the other report-specific commands.

## References

GEA-1710


src/manage_sql_resources.c
```diff
@@ -799,6 +799,20 @@ resource_count (const char *type, const get_data_t *get)
         }
       return 0;
     }
+  else if (strcmp (type, "report_cve") == 0)
+    {
+      report_t report;
+      const gchar *report_uuid = get_data_get_extra (get, "report_id");
+      if (!str_blank (report_uuid))
+        {
+          find_report_with_permission (report_uuid, &report, "get_reports");
```
mattmundell:
```suggestion
          if (find_report_with_permission (report_uuid, &report, "get_reports"))
            return 0;
```
like /pull/2903.

timopollmeier (approved):

PR #2914, opened 2026-05-13
Add/improve asset counting for authenticated scans

ozgen:
## What

- Added `asset_snapshot_identifiers` to store multiple identifiers per asset snapshot.
- Moved agent IDs and container digests into the new identifier-based model.
- Changed target asset matching to use identifier maps instead of single `ip_address`, `hostname`, and `mac_address` fields.
- Updated target candidate loading to read from `asset_snapshot_identifiers`.
- Removed unused target match flags because identifier maps are now used directly.
- Updated merge logic so equivalent target candidates are merged under one `asset_key`.

## Why

- A target asset can have multiple IPs, hostnames, and MAC addresses.
- Matching on only one IP/hostname/MAC can create duplicate assets or miss valid matches.
- Storing identifiers in a separate table makes matching more flexible and consistent for targets, agents, and container images.
- Equivalent snapshots with the same identifiers should share the same `asset_key`.
- The new structure prepares asset counting for more reliable authenticated scan results.

## References

GEA-1773

## Checklist

- [x] Tests


src/manage_sql_assets.c
```diff
@@ -1430,16 +1746,19 @@ asset_snapshots_set_target_asset_keys (scanner_t scanner)
...
+        continue;
 
-      /* Delegate all logic to the merge apply helper */
-      asset_target_merge_apply_sql (row_id, &obs, scanner);
+      obs = asset_target_obs_new ();
+
+      asset_target_obs_load_from_snapshot (obs, row_id);
+      asset_target_merge_apply_sql (row_id, obs, scanner);
+
+      asset_target_obs_free (obs);
```
mattmundell, line 1761: This is leaking obs. asset_target_obs_free should be freeing obs itself.

ozgen: Yes, you are right. The current function only destroys the identifiers hash table and resets the fields
fixed: https://github.com/greenbone/gvmd/pull/2914/changes/a7e489d8880a62667081f83e9a5af0c139f47be6

timopollmeier (approved):

src/manage_asset_keys.c
```diff
@@ -26,121 +26,399 @@
...
+{
+  float score = 0.0f;
+
+  if (!candidate || !obs)
+    return 0.0f;
+
+  score += (float) asset_identifier_map_intersection_count_for_type (
+    candidate->identifiers,
+    obs->identifiers,
+    ASSET_IDENTIFIER_TYPE_MAC) * SCORE_WEAK;
```
g-w, line 391: This differs from the method description: 

```suggestion
    ASSET_IDENTIFIER_TYPE_MAC) * SCORE_STRONG;
```

src/manage_asset_keys.c
```diff
@@ -157,92 +435,190 @@ asset_merge_decision_init (asset_merge_decision_t *out)
...
+      float score;
 
-      if (!c->asset_key || !*c->asset_key)
+      if (!asset_candidate_has_key (candidate))
         continue;
 
-      float s = candidate_score (c);
-      if (s <= 0.0f)
+      if (!asset_identifier_map_is_subset (obs->identifiers,
+                                           candidate->identifiers))
```
g-w, line 542: Why do we have to skip these? The conditions should check if there is no overlap, i.e. if the intersection of the two sets is not empty.

src/manage_asset_keys.c
```diff
@@ -157,92 +435,190 @@ asset_merge_decision_init (asset_merge_decision_t *out)
...
+   * Reuse the best matching existing candidate asset_key.
+   */
   for (size_t i = 0; i < candidates_len; i++)
     {
-      const asset_candidate_t *c = &candidates[i];
+      const asset_candidate_t *candidate = &candidates[i];
+      float score;
 
-      if (!c->asset_key || !*c->asset_key)
+      if (!asset_candidate_has_key (candidate))
```
g-w, line 538: Shouldn't this not always be true?

src/manage_asset_keys.c
```diff
@@ -157,92 +435,190 @@ asset_merge_decision_init (asset_merge_decision_t *out)
...
+    }
+}
+
+/**
+ * @brief Decide the asset_key for a target observation.
+ *
+ * Rules:
+ * - If an existing candidate contains all observed identifiers, reuse the best
+ *   matching candidate key.
+ * - Else if the observation contains one or more existing candidates, caller
```
g-w, line 498: This is wrong. Always the best match should be used (if there are any). If the new observation connects two existing assets they should be merged into the best match.

ozgen: @g-w 
You are right, the comment is misleading.

When the new observation matches existing candidates, we do not create a new `asset_key`. We keep the best matching existing key and merge the other matching candidates into it.

A new `asset_key` is only created when nothing matches.

src/manage_asset_keys.c
ozgen: @g-w , Yes, normally this should always be true.

The check is only there to be defensive. If a candidate is empty, partially initialized, or has no asset_key for any reason, we skip it instead of using broken input. So it is not part of the expected path, but it makes the code safer. :smile:

src/manage_asset_keys.c
ozgen: @g-w The subset check is intentional here.

At this point, we are not only checking whether two assets are somehow related. We are deciding whether it is safe to reuse an existing `asset_key`.

If we only check for any overlap, one shared value would already be enough. For example, the same IP or the same hostname would count as a match. But these values are not always reliable. They can change, be reused, or point to a different machine later.

So the subset check is safer. We reuse the existing `asset_key` only when the candidate already contains all identifiers from the new observation.

src/manage_asset_keys.c
ozgen: @g-w, fixed in https://github.com/greenbone/gvmd/pull/2921

src/manage_asset_keys.c
ozgen: fixed in https://github.com/greenbone/gvmd/pull/2921

PR #2928, opened 2026-05-21
Add/add agent group scheduler cron time

ozgen:
## What

* Added `scheduler_cron_time` support for agent groups.
* Updated agent group SQL handling to read and store scheduler cron time.
* Added logic to sync agent controller cron configuration from the current group DB state.
* Updated create and modify agent group handling to use synced helper functions.
* Added rollback handling for create and modify when agent controller sync fails.
* Updated modify handling to sync all affected agents, including added, kept, and removed agents.
* Added helper logic to read the previous agent group state before modification.
* Exposed the update-list resync helper where needed.
* Updated GMP documentation for `create_agent_group`, `modify_agent_group`, and `get_agent_groups`.
* Reorganized agent helper functions and cleaned up agent header declarations.
* Verified the changes with local builds/tests and manual checks of the generated controller payload.

## Why

* Agent group cron configuration needs to be stored in gvmd and synchronized with the agent controller.
* The controller should receive the final cron state after an agent group is created or modified.
* Modifying an agent group can affect both newly added and removed agents, so all affected agents need to be resynced.
* Rollback handling helps avoid inconsistent state between gvmd and the agent controller when synchronization fails.
* Reading the cron state from the database avoids duplicating cron merge logic in GMP handlers.
* The GMP documentation needs to match the new request and response structure.
* The code cleanup improves readability without changing behavior.

## References

GEA-1791


src/manage_sql_agent_groups.c
```diff
@@ -1059,4 +1066,67 @@ trash_agent_group_comment (agent_group_t agent_group)
...
+{
+  iterator_t it;
+
+  if (schedule_cron_times == NULL)
+    return -1;
+
+  *schedule_cron_times = g_ptr_array_new_with_free_func (g_free);
+
+  if (agent_uuid == NULL || *agent_uuid == '\0')
+    return -1;
```
mattmundell, line 1216: Leaking *schedule_cron_times. I think it should be before the g_ptr_array_new_with_free_func.

mattmundell: Probably good to add "on success" to the schedule_cron_times doc too.

ozgen: fixed: https://github.com/greenbone/gvmd/pull/2928/changes/cb42b835503b2cbde30bc5289fe6c044303e7ed4

bjoernricks (dismissed):

bjoernricks (approved):

PR #2938, opened 2026-06-03
fix: return agent group sync validation errors

ozgen:
## What

- Passed agent controller sync errors back when creating or modifying agent groups.
- Included validation error details in the GMP response when agent group sync fails.

## Why

- Users should see the real validation error instead of a generic invalid input error.
- This makes agent group create/modify failures easier to understand and debug.

## References

GEA-1701

### DIFF
--- a/src/gmp_agent_groups.c
+++ b/src/gmp_agent_groups.c
@@ -292,6 +292,7 @@ create_agent_group_run (gmp_parser_t *gmp_parser, GError **error)
   agent_group_data_t group_data;
   agent_uuid_list_t agent_uuids = NULL;
   agent_group_t new_agent_group;
+  GPtrArray *errs = NULL;
 
   if (!acl_user_may ("create_agent_group"))
     {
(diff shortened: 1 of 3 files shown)


src/gmp_agent_groups.c
```diff
@@ -465,8 +466,35 @@ create_agent_group_run (gmp_parser_t *gmp_parser, GError **error)
...
 
     case AGENT_GROUP_RESP_INVALID_ARGUMENT:
-      SEND_TO_CLIENT_OR_FAIL (XML_ERROR_SYNTAX ("create_agent_group", "Invalid input"));
-      log_event_fail ("agent_group", "Agent Group", NULL, "created");
+      if (errs)
+        {
+          gchar *status_text = concat_error_messages (
+            errs, "; ", "Validation failed for config: ");
+          if (!status_text)
+            status_text = g_markup_escape_text ("Validation failed for config.",
```
a-h-abdelsalam: Is this needed ? It is also checked below in line 481.
The same applies for the modify function.

a-h-abdelsalam (approved):

PR #2949, opened 2026-06-09
Fix: cleanup iterators in check_cpe_match_rule

mattmundell:
## What

Cleanup all three iterators in `check_cpe_match_rule`.

## Why

They were leaking.

## Testing

Ran GMP commands on main, saw Asan leak warnings.
Ran same commands with patch, Asan warnings gone.

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -1814,9 +1814,9 @@ check_cpe_match_rule (long long int node, gboolean *match, gboolean *vulnerable,
       child_node = cpe_match_node_childs_iterator_id (&cpe_match_node_childs);
       check_cpe_match_rule (child_node, match, vulnerable, report_host, host_cpe);
       if (strcmp (operator, "AND") == 0 && !(*match))
-        goto cleanup;
+        goto cleanup_node_childs;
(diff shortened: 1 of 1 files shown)


src/manage.c
```diff
@@ -1873,11 +1874,14 @@ check_cpe_match_rule (long long int node, gboolean *match, gboolean *vulnerable,
...
       g_free (vee);
       if (strcmp (operator, "AND") == 0 && !(*match))
-        goto cleanup;
+        goto cleanup2;
       if (strcmp (operator, "OR") == 0 && (*match) && (*vulnerable))
-        goto cleanup;
+        goto cleanup2;
     }
- cleanup:
+ cleanup2:
```
ozgen: Could we avoid names like `cleanup1` and `cleanup2` here? They are not very descriptive.

Since the labels represent which iterators have already been initialized, names like `cleanup_match_ranges` and `cleanup_child_nodes` would make the `goto` flow easier to read.

mattmundell: Yes, much better, fixed, thanks.

ozgen (approved):

PR #2983, opened 2026-06-22
Add: Add a user setting for a maintenance window.

a-h-abdelsalam:
## What
Add a user setting for a maintenance window.

## Why
Required to define a time slot in which no scans are executed.

## References
GEA-1846

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -61,6 +61,7 @@
 #include "manage_sql_secinfo.h"
 #include "manage_sql_targets.h"
 #include "manage_sql_tickets.h"
+#include "manage_sql_settings.h"
 #include "manage_sql_tls_certificates.h"
 #include "manage_tags.h"
 #include "manage_web_application_targets.h"
@@ -2639,6 +2640,42 @@ static int
 run_agent_control_task (task_t task, char **report_id);
 #endif
 
+/*
+  * @brief Check if scans are currently blocked by the
+  *        user's maintenance window.
+  *
+  * @return 1 if scans are blocked, 0 if not.
+*/
+static int
+scans_blocked_by_maintenance_window (const char *user_uuid)
+{
+  gchar *timezone, *maintenance_ical;
+  time_t now;
+
+  if (user_uuid == NULL)
+    {
+      g_warning ("%s: user_uuid is NULL, skipping maintenance window check",
+                 __func__);
+      return 0;
+    }
+
+  timezone = setting_timezone_for_user (user_uuid);
+  maintenance_ical = setting_maintenance_window_for_user (user_uuid);
+  now = time (NULL);
+
+  int window_active = 0;
+  if (timezone && maintenance_ical && maintenance_ical[0] != '\0')
+    window_active = icalendar_is_window_active (maintenance_ical, now, timezone);
+  else
+    g_debug ("%s: timezone or maintenance window setting not set,"
+             " skipping maintenance window check", __func__);
+
+  g_free (timezone);
+  g_free (maintenance_ical);
+
+  return window_active;
+}
+
 /**
  * @brief Start or resume a task.
  *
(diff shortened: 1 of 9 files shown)


src/manage.c
```diff
@@ -2639,6 +2640,38 @@ static int
...
+  time_t now;
+
+  if (user_uuid == NULL)
+    {
+      g_warning ("%s: user_uuid is NULL, skipping maintenance window check",
+                 __func__);
+      return 0;
+    }
+
+  timezone = setting_timezone_for_user (user_uuid);
```
mattmundell, line 2662: Should be freed.

src/manage.c
```diff
@@ -2639,6 +2640,38 @@ static int
...
+  maintenance_ical = setting_maintenance_window_for_user (user_uuid);
```
mattmundell, line 2663: Also should be freed.

src/manage_sql_settings.c
```diff
@@ -614,6 +647,25 @@ modify_setting (const gchar *uuid, const gchar *name,
             }
         }
 
+      if (strcmp (uuid, SETTING_UUID_MAINTENANCE_WINDOW) == 0)
+        {
+          /* Maintenance Window */
+          gchar *error_out = NULL;
+          if (value
+              && strcmp (value, "")
+              && icalendar_from_string (value,
```
mattmundell: On success this returns an ical component, which should be freed with icalcomponent_free.

src/manage_sql_settings.c
```diff
@@ -614,6 +647,25 @@ modify_setting (const gchar *uuid, const gchar *name,
...
+                                        icaltimezone_get_utc_timezone (),
+                                        &error_out)
```
mattmundell: It looks like this is technically also leaking on success, but I think that is an existing error in `icalendar_from_string`.

a-h-abdelsalam: where do you see the leak for this one on success ?

mattmundell: `icalendar_from_string` is always assigning the contents of the `warnings_buffer` string to `error` when `error` is given, even on success (see line 581 of `manage_utils.c`).

mattmundell: Looking closer, that is intentional, and the `error` return arg from `icalendar_from_string` should always be freed, including on success.

This is intentional, because for example `handle_modify_schedule` sends the error output (it gets it via `modify_schedule`) as `<status_details>` on success.

src/manage_sql_settings.c
mattmundell: You sure your patch for this is right? Seems like it will now call `icalendar_from_string` even when value is NULL or "".

a-h-abdelsalam: Overlooked this. Should be fixed now

src/manage_utils.c
```diff
@@ -1128,6 +1128,63 @@ icalendar_first_time_from_vcalendar (icalcomponent *vcalendar,
...
+   }
+
+  start = icalendar_next_time_from_vcalendar (vcalendar,
+                                              current_time,
+                                              tzid,
+                                              -1);
+
+  duration = icalendar_duration_from_vcalendar (vcalendar);
+
+  icalcomponent_free (vcalendar);
```
rdttmr, line 1180: According to Matt's comments, we should also free `error` here I assume?

robindittmar (approved):

src/manage_utils.c
```diff
@@ -1128,6 +1128,63 @@ icalendar_first_time_from_vcalendar (icalcomponent *vcalendar,
...
+  tz = icalendar_timezone_from_string (tzid);
+  if (tz == NULL)
+   {
+     g_warning ("%s: A timezone is required to check the maintenance window,"
+                " but timezone '%s' could not be found.", __func__, tzid);
+     return FALSE;
+   }
+
+  char *error = NULL;
+  vcalendar = icalendar_from_string (ical_string, tz, &error);
```
mattmundell, line 1164: This one is also missing the free of error on success.

mattmundell: Ah you did it already, nice.

PR #2988, opened 2026-06-25
Support WAS scanner preferences

a-h-abdelsalam:
## What
Add support for web application scanner preferences.

## Why
Preferences are now supported by the scanner.

## References
GEA-1872


src/manage_sql_configs.c
```diff
@@ -2040,6 +2041,24 @@ set_task_preferences (task_t task, array_t *preferences)
...
+                      && type == SCANNER_TYPE_WEB_APPLICATION)
+                    {
+                      return 5;
+                    }
+                  if ((strcmp (pair->name, "ajax_spider_timeout") == 0)
+                      && type == SCANNER_TYPE_WEB_APPLICATION)
+                    {
+                      gchar *endptr = NULL;
+                      errno = 0;
+                      long timeout = strtol(pair->value, &endptr, 10);
```
ozgen, line 2056: Could we also check that at least one digit was parsed? Currently, an empty value is accepted because `strtol() `returns `0` and `*endptr` is already `'\0'`.

ozgen (dismissed):

ozgen (approved):

PR #2991, opened 2026-06-25
Fix: quote NVT OID in SQL statements

imurd:
## What
Wrap NVT OIDs in sql_quote() before interpolating them into SQL in
insert_nvt(), insert_vt_severities(), insert_vt_refs() and the OSP /
openvasd preference deletes.

## Why
The OID comes straight from the feed/scanner (nvti_oid) and was
concatenated into the SQL unquoted, while every other string in the same
statements already goes through sql_quote(). An OID containing a single
quote breaks the generated statement, so the NVT update/rebuild fails,
and it allows SQL injection from feed data.

## Tests
Added a regression test covering OID quoting.

## Checklist

- [ ] Tests


bjoernricks:
Hi, without a description why this is necessary and what actually is broken without your changes this PR can't be merged. Please add the PR description and update your commit message to add why the change has been done (for example with `git rebase -i`)

PR #2994, opened 2026-06-29
[GEA-1178] Add target_type to tasks table

rdttmr:
## What

Add `target_type` column to `tasks` table.
Remove `agent_group`, `oci_image_target` and `web_application_target` (+ `*_location`) columns from `tasks` table

This PR only refactors the "backend", it intentionally does not modify any APIs, refactors behavior or simplifies other low hanging fruit that could be achieved after the refactoring.
It only intents to swap out database columns, while not changing any behavior. (The old columns are not yet dropped, which is a future endeavor once the changes have been working in production)

One side effect of this is, that the task trend feature is now available for all target types.

## Why

Modernize code- and database, make it easier to add different types of targets in the future.

## References

GEA-1178


src/manage_migrators.c
```diff
@@ -4195,6 +4195,109 @@ migrate_279_to_280 ()
...
+              " WHERE id = $1",
+              SQL_RESOURCE_PARAM (id),
+              SQL_RESOURCE_PARAM (new_target),
+              SQL_INT_PARAM (new_location),
+              SQL_INT_PARAM (new_target_type),
+              NULL);
+    }
+  cleanup_iterator (&tasks);
+
+  /* Drop old, unused target columns */
```
ozgen, line 4470: I would prefer to drop these columns in a separate migration after the full migration has been completed and verified, and after all task flows work properly. Maybe we can drop them in the next snapshot release. This feels safer because we avoid removing the old target data too early.

src/manage_pg.c
```diff
@@ -1285,7 +1285,32 @@ manage_create_sql_functions ()
             "$$ LANGUAGE SQL;");
 
       /* column hostname in table report_hosts was added in version 273 */
```
timopollmeier: I think this comment should be moved to the `else if` below and a new one about the version 282 changes should be added here.

src/manage_pg.c
```diff
@@ -1531,7 +1578,139 @@ manage_create_sql_functions ()
                TASK_STATUS_DONE);
         }
 
-      sql ("CREATE OR REPLACE FUNCTION task_trend (integer, integer, integer)"
+      if (current_db_version >= 282)
```
timopollmeier, line 1583: This should also have a comment about the relevant version 282 changes.

src/manage_tasks.h
```diff
@@ -7,5 +7,15 @@
...
 
+typedef enum tasks_target_type
+{
+  TASKS_TARGET_TYPE_UNDEFINED = -1,
+  TASKS_TARGET_TYPE_IMPORT_TASK = 0,
+  TASKS_TARGET_TYPE_REGULAR = 1,
+  TASKS_TARGET_TYPE_AGENT_GROUP = 2,
+  TASKS_TARGET_TYPE_OCI_IMAGE = 3,
+  TASKS_TARGET_TYPE_WEB_APPLICATION = 4,
+} tasks_target_type_t;
```
timopollmeier: For me it would feel more consistent to use the singular `task` in the type and value names instead of `tasks`.

src/manage_pg.c
rdttmr: Comment has been added

src/manage_pg.c
rdttmr: Comment has been added

src/manage_tasks.h
rdttmr: I changed the enum (along with all references) to

```
typedef enum task_target_type
{
  TASK_TARGET_TYPE_UNDEFINED = -1,
  TASK_TARGET_TYPE_IMPORT_TASK = 0,
  TASK_TARGET_TYPE_REGULAR = 1,
  TASK_TARGET_TYPE_AGENT_GROUP = 2,
  TASK_TARGET_TYPE_OCI_IMAGE = 3,
  TASK_TARGET_TYPE_WEB_APPLICATION = 4,
} task_target_type_t;
```

PR #3004, opened 2026-07-03
Fix/filtering issues in report details

ozgen:
## What

Apply the explicit host filter to report hosts, operating systems, ports, TLS certificates, applications, CVEs, and closed CVEs.

The host value is now exposed by the report filter controls and passed to the related report data queries.

## Why

Previously, the host filter was not consistently applied by the backend when `result_hosts_only` was disabled.

This change keeps the filtered report data consistent with the selected host.

## References

GEA-1681


src/manage_filters.h
```diff
@@ -25,12 +25,15 @@ manage_filter_controls (const gchar *, int *, int *, gchar **, int *);
 void
 manage_report_filter_controls (const gchar *, int *, int *, gchar **, int *,
                                int *, gchar **, gchar **, gchar **, gchar **,
-                               gchar **, int *, int *, int *, int *, gchar **);
+                               gchar **, gchar **, int *, int *, int *, int *,
+                               gchar **);
+
 int
-manage_report_filter_controls_from_get (const get_data_t *, gchar **, int *,
+manage_report_filter_controls_from_get (const get_data_t *get, gchar **, int *,
```
mattmundell:
```suggestion
manage_report_filter_controls_from_get (const get_data_t *, gchar **, int *,
```

src/manage_report_applications.c
```diff
@@ -95,15 +96,47 @@ get_report_applications (report_t report,
...
+                                                NULL,
+                                                NULL,
+                                                NULL,
+                                                NULL,
+                                                NULL,
+                                                NULL);
+
+  if (ret)
+    {
+      g_free (host_filter);
```
mattmundell, line 133: This could g_free NULL if some future situation called get_report_application with get NULL.

It's theoretical and minor, but we could init host_filter to NULL. (Even better manage_report_filter_controls_from_get should always set it to NULL.)

ozgen: fixed with: https://github.com/greenbone/gvmd/pull/3004/changes/1c3318171b62cf7f03f28647530488ccb92f4794

timopollmeier (approved):

PR #3009, opened 2026-07-06
More detailed error handling for function "stop_task (...)".

jhelmold:
## What
The returned error codes of the function "stop_task (...)" are handled in more detail now.

## Why
This is part of a bug-fix.

## References
GEA-1642

### DIFF
--- a/src/manage.c
+++ b/src/manage.c
@@ -2334,6 +2334,8 @@ stop_osp_task (task_t task)
   if (!connection)
     goto end_stop_osp;
   ret = osp_delete_scan (connection, scan_id);
+  if (ret)
+    ret = 4;
   osp_connection_close (connection);
 
 end_stop_osp:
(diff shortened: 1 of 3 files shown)


src/manage.c
```diff
@@ -2414,7 +2414,7 @@ stop_task (const char *task_id)
 
   if (scanner_type (task_scanner (task)) == SCANNER_TYPE_OPENVAS
       || scanner_type (task_scanner (task)) == SCANNER_TYPE_OSP_SENSOR)
-    return stop_osp_task (task);
+    return stop_osp_task (task) * 10;
```
timopollmeier, line 2417: I think we have enough return codes for `stop_task` where using an enum and a small helper function that maps the return values from `stop_osp_task` would make things more readable.

src/manage_sql_filters.c
```diff
@@ -866,6 +866,8 @@ columns_select_column_single (column_t *select_columns,
     *type = KEYWORD_TYPE_UNKNOWN;
   if (select_columns == NULL)
     return NULL;
+  if (filter_column == NULL)
+    return NULL;
```
timopollmeier, line 870: This appears to be unrelated to the error handler improvements.

PR #3012, opened 2026-07-06
Fix openvasd verification

a-h-abdelsalam:
## What
Add scanner verification for openvasd and container image scanners.
Verification for web application scanner will be added separately.

## Why
Verification was missing.

## References
GEA-1861

### DIFF
--- a/src/manage_container_image_scanner.c
+++ b/src/manage_container_image_scanner.c
@@ -1239,10 +1239,12 @@ verify_container_image_scanner_connection (scanner_t scanner)
   else if (response->code == -1)
     {
       gboolean has_relay = scanner_has_relay (scanner);
+      char *host = scanner_host (scanner, has_relay);
       g_warning ("%s: failed to connect to %s:%d",
                  __func__,
-                 scanner_host (scanner, has_relay),
+                 host,
                  scanner_port (scanner, has_relay));
+      g_free (host);
       ret = 1;
     }
   else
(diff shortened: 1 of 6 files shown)


src/manage_container_image_scanner.c
```diff
@@ -1201,4 +1201,60 @@ stop_container_image_task (task_t task)
...
+  if (response->code == 200)
+   {
+      ret = 0;
+   }
+  else if (response->code == -1)
+    {
+      gboolean has_relay = scanner_has_relay (scanner);
+      g_warning ("%s: failed to connect to %s:%d",
+                 __func__,
+                 scanner_host (scanner, has_relay),
```
ozgen: `scanner_host()` seems to return an allocated `char *`. Should we store it in a temporary variable and free it?

src/manage_openvasd.c
```diff
@@ -784,4 +784,59 @@ handle_openvasd_scan_end (task_t task, int handle_progress_rc, gboolean discover
...
+  if (response->code == 200)
+    {
+      ret = 0;
+    }
+  else if (response->code == -1)
+    {
+      gboolean has_relay = scanner_has_relay (scanner);
+      g_warning ("%s: failed to connect to %s:%d",
+                 __func__,
+                 scanner_host (scanner, has_relay),
```
ozgen: same here.

src/manage_container_image_scanner.c
a-h-abdelsalam: Should be fixed in [8f255c4](https://github.com/greenbone/gvmd/pull/3012/changes/8f255c42e6480f44ae1b695c182a842868629c1c)

ozgen (approved):

PR #3020, opened 2026-07-08
Add: Add scanner verification for web application scanner.

a-h-abdelsalam:
## What
Add scanner verification for web application scanner. 
Use HTTP_SCANNER_PATH_PREFIX instead of HTTP_SCANNER_SCAN_PREFIX.

## Why
Verification was missing.
Path prefix is used for all endpoints. Scan prefix is only used for scan endpoints.

Requires https://github.com/greenbone/gvm-libs/pull/1088

## References
GEA-1861

### DIFF
--- a/src/manage_sql.c
+++ b/src/manage_sql.c
@@ -24843,8 +24843,9 @@ verify_scanner (const char *scanner_id, char **version)
            || scanner_iterator_type (&scanner) == SCANNER_TYPE_OPENVASD_SENSOR)
     {
       scanner_t scanner_row_id = get_iterator_resource (&scanner);
-      int res = verify_openvasd_scanner_connection (scanner_row_id);
       cleanup_iterator (&scanner);
+
+      int res = verify_openvasd_scanner_connection (scanner_row_id);
       if (res)
         return 2;
       return 0;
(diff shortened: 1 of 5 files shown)


src/manage_sql.c
```diff
@@ -24878,11 +24879,11 @@ verify_scanner (const char *scanner_id, char **version)
 #if ENABLE_WEB_APPLICATION_SCANNING
   else if (scanner_iterator_type (&scanner) == SCANNER_TYPE_WEB_APPLICATION)
     {
-      // TODO: replace with actual version from the scanner
-      if (version)
-        *version = g_strdup ("TestVersion");
-
+      scanner_t scanner_row_id = get_iterator_resource (&scanner);
+      int res = verify_web_application_scanner_connection (scanner_row_id);
       cleanup_iterator (&scanner);
```
ozgen, line 24885: I prefer  to call the `cleanup_iteratior()` one like above 

like :

```
scanner_t scanner_row_id = get_iterator_resource (&scanner);
cleanup_iterator (&scanner);

int res = verify_web_application_scanner_connection (scanner_row_id);
if (res)
  return 2;
return 0;

```

ozgen (dismissed):

src/manage_sql.c
a-h-abdelsalam: yes makes sense. I updated it for others as well.

ozgen (approved):

PR #3021, opened 2026-07-10
Fix/security intelligence tls config

ozgen:
## What

- Upgrade the `gvm-auth-lib` version.
- Adapt the authentication interface to the updated library API.
- Improve Security Intelligence error handling and logging.
- Preserve raw response errors and pass them back to the caller.
- Fix `GPtrArray` NULL handling during report export.

## Why

- The updated authentication library requires interface changes.
- Better error propagation makes failed exports easier to diagnose.
- The pointer handling fix prevents crashes when no error array is available.

## References

GEA-1920

### DIFF
--- a/src/manage_report_export_scheduler.c
+++ b/src/manage_report_export_scheduler.c
@@ -277,7 +277,7 @@ process_report_export (report_t report, int retry_count,
         reason = g_strdup ("The request has timed out");
       else if (result == EXPORT_REPORT_RESULT_TOKEN_GENERATION_FAILED)
         reason = g_strdup ("Could not generate access_token");
-      if (errors && errors->len > 0)
+      else if (errors && errors->len > 0)
         {
           reason = errors_to_string (errors);
         }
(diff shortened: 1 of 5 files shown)


src/manage_report_export_scheduler.c
```diff
@@ -231,9 +274,13 @@ process_report_export (report_t report, int retry_count,
   else
     {
       if (result == EXPORT_REPORT_RESULT_TIMEOUT)
-          reason = g_strdup ("The request has timed out");
+        reason = g_strdup ("The request has timed out");
       else if (result == EXPORT_REPORT_RESULT_TOKEN_GENERATION_FAILED)
         reason = g_strdup ("Could not generate access_token");
+      if (errors && errors->len > 0)
```
rdttmr: Shouldn't this be an `else if`?
What if the result is some variant of `EXPORT_REPORT_RESULT_` indicating an error, but `errors` is also set?
Then we would leak memory, I believe

robindittmar (approved):

PR #3038, opened 2026-07-15
Changed handling of oci_image_target_id when editing a Container Image Task.

jhelmold:
## What
Changed the handling of the oci_image_target_id when editing a Container Image Task to make the storage of the task without misleading error messages possible again.

## Why
This fixes a bug.

## References
GEA-1605

## Checklist


src/manage_sql.c
```diff
@@ -19213,7 +19213,9 @@ modify_task (const gchar *task_id, const gchar *name,
 
       if ((task_run_status (task) != TASK_STATUS_NEW)
           && (task_alterable (task) == 0))
-        return MODIFY_TASK_TARGET_STATUS_MUST_BE_NEW;
+        {
```
a-h-abdelsalam, line 19205: I think we need to keep this to prevent changing the target if the task is not new. 

Instead we can add the check:
```
      if (strcmp (oci_image_target_id, "0") == 0)
        {
          /* Leave it as is. */
        }
```
You can take a look how it is done for `target_id` a few lines above.

GSA then needs to send `oci_image_target_id` and `scanner_id` as 0

a-h-abdelsalam (changes requested):

src/manage.h
a-h-abdelsalam, line 1001: I think this is out of scope of this change and is already fixed on the main branch

src/manage_sql.c
```diff
@@ -19211,7 +19211,11 @@ modify_task (const gchar *task_id, const gchar *name,
     {
       oci_image_target_t oci_image_target = 0;
 
-      if ((task_run_status (task) != TASK_STATUS_NEW)
+      if (strcmp (oci_image_target_id, "0") == 0)
+        {
+          /* Leave it as is. */
+        }
+      else if ((task_run_status (task) != TASK_STATUS_NEW)
```
a-h-abdelsalam, line 19208: Just a nitpick, could you adjust the indentation in the line below ?

a-h-abdelsalam (approved):

PR #3040, opened 2026-07-15
Fix: Add per-object permission checks for modify/delete asset.

a-h-abdelsalam:
## What
Add per-object permission checks for modifying and deleting assets.

Refactor delete_asset to use helper functions for resolving asset type and checking permissions.

## Why
Only role-level capability was checked which allowed any user with read-only access to modify/delete an asset.

## References
GEA-1877

### DIFF
--- a/src/manage_sql_assets.c
+++ b/src/manage_sql_assets.c
@@ -162,6 +162,19 @@ typedef enum
   AS_COL_SCANNER = 8
 } asset_snapshot_col_t;
 
+/**
+ * @brief Types of assets that can be deleted including unknown and error.
+ */
+typedef enum
+{
+  DELETE_ASSET_TYPE_UNKNOWN = 0,
+  DELETE_ASSET_TYPE_HOST_IDENTIFIER = 1,
+  DELETE_ASSET_TYPE_HOST_OS = 2,
+  DELETE_ASSET_TYPE_OS = 3,
+  DELETE_ASSET_TYPE_HOST = 4,
+  DELETE_ASSET_TYPE_ERROR = -1,
+} delete_asset_type_t;
+
 static int
 report_host_dead (report_host_t);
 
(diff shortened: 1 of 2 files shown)


src/manage_sql_assets.c
```diff
@@ -3592,218 +3587,297 @@ delete_report_assets (const char *report_id)
...
-        break;
-      case 1:        /* Too few rows in result of query. */
-        asset = 0;
+      sql_rollback ();
+      return auth_ret;
+    }
+  if (auth_ret == 1)
+    {
+      sql_rollback ();
+      return 2;
```
ozgen: Should this return `99` for permission denied, as in the old implementation, instead of `2` for asset not found?

src/manage_sql_assets.c
```diff
@@ -3446,65 +3474,32 @@ modify_asset (const char *asset_id, const char *comment)
...
-       " modification_time = m_now ()"
-       " WHERE id = %llu;",
-       quoted_comment, asset);
-
-  g_free (quoted_comment);
+  sql_ps ("UPDATE hosts SET"
+          " comment = $1,"
+          " modification_time = m_now ()"
+          " WHERE id = $2;",
+          SQL_STR_PARAM(comment),
```
ozgen: Should this use `SQL_STR_PARAM (comment ?: "")` to keep the old behavior when `comment` is `NULL`?

src/manage_sql_assets.c
a-h-abdelsalam: I had this initially returning 99 but then changed it to 2 after checking what most modify_x functions do when there is no permission.
I agree to better keep it as old implementation. Changed it.

src/manage_sql_assets.c
a-h-abdelsalam: yes good catch

ozgen (approved):

PR #3043, opened 2026-07-17
Add/get scan report command

ozgen:
## What

- Add a new `get_scan_report` command
- Return a structured report summary
- Include task, target, progress, counts, severity, and scan information
- Add GMP XML output
- Fix memory cleanup
- Add command documentation

## Why

- Separate report retrieval from report export
- Avoid creating temporary report files

## References

GEA-1937


src/gmp_scan_report.c
```diff
@@ -0,0 +1,675 @@
...
+
+  if (send_report_xml (
+    gmp_parser,
+    "<comment>%s</comment>",
+    task->comment ? task->comment : ""))
+    return TRUE;
+
+  target = task->target;
+
+  if (target
```
a-h-abdelsalam, line 208: I think this will miss other target types like container scanning or web application scanning. Is this intended ?

src/gmp_scan_report.c
```diff
@@ -0,0 +1,675 @@
...
+  const gchar *attribute;
+
+  get_data_parse_attributes (&get_scan_report_data.get,
+                             "report",
+                             attribute_names,
+                             attribute_values);
+
+  if (find_attribute (attribute_names,
+                      attribute_values,
+                      "report_id",
```
timopollmeier, line 551: Wouldn't it make sense to name the attribute `scan_report_id` for consistency?

src/gmp_scan_report.c
```diff
@@ -0,0 +1,675 @@
...
+                      &attribute))
+    {
+      get_scan_report_data.report_id = g_strdup (attribute);
+
+      get_data_set_extra (&get_scan_report_data.get,
+                          "report_id",
+                          attribute);
```
timopollmeier, line 558: Is this really needed? For me this only makes sense for getting objects inside the report like `get_report_hosts`.

src/manage_resources.c
```diff
@@ -324,7 +325,8 @@ type_owned (const char* type)
          && strcasecmp (type, "report_operating_system")
          && strcasecmp (type, "report_port")
          && strcasecmp (type, "report_tls_certificate")
-         && strcasecmp (type, "report_vuln");
+         && strcasecmp (type, "report_vuln")
+         && strcasecmp (type, "scan_report");
```
timopollmeier, line 329: The reports do have owners, so this should not be in here even though it may require some changes to ensure the permission checks use the "subtype".

src/gmp_scan_report.c
ozgen: fixed with https://github.com/greenbone/gvmd/pull/3043/changes/f3f212fa0a4faa09800df890318b99cdb07b553c

a-h-abdelsalam (approved):

PR #3045, opened 2026-07-20
Add: ec dependencies

pascalholthaus:
## What
Add: install-ec-dependencies.sh

## Why
Add curl, graphviz and pycha.
These components are required for specific reports in the enterprise containers.

If someone knows something better as pip for pycha, please commit to this pr.

"libcurl4t64" gets updated during the curl installation; hopefully, that doesn't break any dependencies.

## References
DEVOPS-2201


bjoernricks (commented):
Just put into curl and graphviz into the `runtime-dependencies.list` if they are required for some report format. for which report format do we need pycha?

y0urself:
> Just put into curl and graphviz into the `runtime-dependencies.list` if they are required for some report format. for which report format do we need pycha?

It looks, that pycha seems to be part of GOS ... I don't know if it is for a specific report-format.

https://github.com/greenbone/gsm-python3-pycha

When testing report formats on the OEC, it was noticed, that some libraries are missing.

timopollmeier:
> for which report format do we need pycha?

It's used in the GSR-PDF and GXR-PDF report formats.

.github/install-ec-dependencies.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# GSR/GXR PDF
+# pycha
+
+apt-get update && \
+apt-get install -y --no-install-recommends python3-pip && \
+pip install --no-cache-dir --break-system-packages pycha && \
+apt-get purge -y python3-pip && \
+apt-get autoremove -y --purge && \
+rm -rf /var/lib/apt/lists/*
```
bjoernricks: rm is already done in the docker file calling the script
```suggestion
apt-get autoremove -y --purge
```

bjoernricks (approved):

PR #3058, opened 2026-07-24
Change/split common functionality from get scan report

ozgen:
## What

- add shared report models and lifecycle helpers
- extract common SQL report loading functions
- extract common GMP XML serialization helpers
- rename the scan-specific summary model
- keep scan-specific result handling in scan report modules

## Why

- reduce duplication between scan and audit reports
- keep shared and report-specific logic separated
- make the upcoming get_audit_report implementation easier
- improve maintainability and readability

## References

GEA-1958


src/manage_scan_report.c
mattmundell: Looks like summary->base should be freed here too.

mattmundell: Also here.

Aside: I wonder why these functions need NULL checks at all. They're using glib functions which we accept everywhere else will just abort when we're out of memory. Also, if these functions can return NULL I think their docs should say so.

ozgen: thanks, fixed with: https://github.com/greenbone/gvmd/pull/3058/changes/86df1104a47b81519fb0667cf7b254a184943b46

ozgen: fixed with: https://github.com/greenbone/gvmd/pull/3058/changes/86df1104a47b81519fb0667cf7b254a184943b46

mattmundell: Thanks I'll resolve, but these NULL checks should really go, in my opinion. I think in a separate PR.

We intentionally skip glib memory allocation checks everywhere because we know that glib just aborts for these. This is convenient and keeps things clean. We should do the same here. Why do any of these _new functions need NULL checks? They're all just using glib for allocation.

Now `scan_report_summary_new` can return NULL, so the caller has to worry about doing a NULL check.

It would be more consistent too. Because `scan_report_summary_new` says it can return NULL but `report_resource_summary_new` is missing this info, and `report_summary_base_new` is treated as always succeeding (like they all should be).

src/gmp_report_common.c
```diff
@@ -0,0 +1,299 @@
...
+send_report_base_start (gmp_parser_t *gmp_parser,
+                        report_summary_base_t base)
+{
+  const gchar *creation_time;
+  const gchar *modification_time;
+
+  if (gmp_parser == NULL || base == NULL)
+    return TRUE;
+
+  creation_time = iso_if_time (base->creation_time);
```
mattmundell: These two lines have removed the g_strdup from here, but iso_if_time return a pointer to static memory. So the second call is overwriting the first one.

src/manage_scan_report.c
ozgen: `NULL` checks already covered in here:
https://github.com/greenbone/gvmd/pull/3058/changes#diff-c0374e051c226c4bd0a7b5b73779b14e25f62e660f7a96726eb63ac7fb791d13R245

src/gmp_report_common.c
ozgen: fixed with :https://github.com/greenbone/gvmd/pull/3058/changes/a7ab57f7a8d3ad89f566127c8964c875f89a3175

src/manage_scan_report.c
ozgen: I misunderstood the issue @mattmundell , fixed with https://github.com/greenbone/gvmd/pull/3058/changes/9850c9544314796212730f386e5208a73b2c44fb

timopollmeier (approved):

PR #3060, opened 2026-07-24
[GEA-1810] Task scheduler changes

a-h-abdelsalam:
## What

- Move periodic jobs and scheduler to their own separate processes forked from `serve_and_schedule`
- Scheduler now processes previously forked runners each tick (`scheduled_task_check_runners_startup_status`).
If child startup reports DB connection failure, task is re-queued for retry.
- Runner status is checked in a nonblocking way (nonblocking pipe read + `waitpid(..., WNOHANG)`) to keep the scheduler loop responsive.
- Replace start/stop list handling with a queue and hash set for deduplication.

## Why

Scheduled tasks were sporadically skipped.
This re-work of the scheduler should fix that.

## References

GEA-1810

## Checklist

- [x] Tests (Regression test locally)

### DIFF
--- a/src/gvmd.c
+++ b/src/gvmd.c
@@ -309,9 +309,19 @@ static int feed_version_check_in_progress = 0;
 static pid_t periodic_job_runner_pid = -1;
 
 /**
+ * @brief PID of the dedicated scheduler, or -1.
+ */
+static pid_t scheduler_pid = -1;
+
+/**
  * @brief Set when the periodic worker terminates.
  */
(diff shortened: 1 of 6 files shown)


robindittmar (commented):
I addressed Ahmed's open ToDos:

- Moved scheduler to its own long running process, with no custom SIGCHLD handler
- `run_schedule ()` is run from the scheduler process regularly

- Tested periodic runner and scheduler locally, everything works as expected

timopollmeier (approved):

PR #3067, opened 2026-07-30
Performance: reuse query plans instead of re-planning every statement

TheBeast85:
## What

- Cache server side prepared statements per connection in `sql_pg.c`, keyed by the statement text. The existing `_ps` API only used `PQexecParams`, which sends the text every time and plans it every time.
- Bind the values of the queries that run once per row of a listing: the Super clause and `acl_user_owns_uuid` in `manage_acl.c`, `find_resource_with_permission`, `resource_tag_count` and `target_in_use`.

## Why

gvmd re-planned nearly every statement it sent. With `pg_stat_statements.track_planning` on, `get_targets` with `rows=-1` over 1000 targets showed 10058 statements, 403 ms of planning against 115 ms of execution. `track_planning` is off by default, which is why `pg_stat_statements` reported only the 115 ms and the database looked harmless.

The Super clause alone was 241 ms of 387 ms.

Only statements that bind parameters are cached; the printf style calls put their values in the text, so every call would be a new key and a plan for `WHERE id = 4711` is of no use to `WHERE id = 4712`. The cache is dropped wherever the connection goes away, including `sql_close_fork`. A failed `PQprepare` is not an error, the caller falls back to `PQexecParams`.

## References

Split out of the original PR as requested. The `GET_TARGETS` part is now #3080, the severity part #3081. #3068 is based on this branch.

## Checklist

- [x] Tested manually against a Community Edition stack, ACL isolation verified in both directions

### DIFF
--- a/src/manage_acl.c
+++ b/src/manage_acl.c
@@ -479,11 +479,18 @@ acl_user_has_role (const char *user_uuid, const char *role_uuid)
  *      is "name" then "SELECT ... format" will choose arbitrarily between
  *      the resources that have the same name. */
 /**
- * @brief Super clause.
+ * @brief Super clause with the resource value and the user UUID bound.
  *
- * @param[in]  format  Value format specifier.
+ * The resource value becomes $1 and the user UUID $2.  Both appear three
+ * times in the clause; PostgreSQL allows a parameter to be used more than
+ * once.
+ *
+ * The resource type, the trash suffix and the field name stay in the
+ * statement text.  They come from gvmd's own tables of valid types, never
+ * from user input, and take only a handful of combinations, which keeps
+ * the statement cache in sql_pg.c small.
  */
-#define ACL_SUPER_CLAUSE(format)                                          \
+#define ACL_SUPER_CLAUSE_PS                                               \
   "                name = 'Super'"                                        \
   /*                    Super on everyone. */                             \
   "                AND ((resource = 0)"                                   \
@@ -491,7 +498,7 @@ acl_user_has_role (const char *user_uuid, const char *role_uuid)
   "                     OR ((resource_type = 'user')"                     \
   "                         AND (resource = (SELECT %ss%s.owner"          \
   "                                          FROM %ss%s"                  \
-  "                                          WHERE %s = " format ")))"    \
+  "                                          WHERE %s = $1)))"            \
   /*                    Super on other_user's role. */                    \
   "                     OR ((resource_type = 'role')"                     \
   "                         AND (resource"                                \
@@ -501,7 +508,7 @@ acl_user_has_role (const char *user_uuid, const char *role_uuid)
   "                                        = (SELECT %ss%s.owner"         \
   "                                           FROM %ss%s"                 \
   "                                           WHERE %s"                   \
-  "                                                 = " format "))))"     \
+  "                                                 = $1))))"             \
   /*                    Super on other_user's group. */                   \
   "                     OR ((resource_type = 'group')"                    \
   "                         AND (resource"                                \
@@ -510,12 +517,12 @@ acl_user_has_role (const char *user_uuid, const char *role_uuid)
   "                                  WHERE \"user\""                      \
   "                                        = (SELECT %ss%s.owner"         \
   "                                           FROM %ss%s"                 \
-  "                                           WHERE %s = " format ")))))" \
+  "                                           WHERE %s = $1)))))"         \
   "                AND subject_location = " G_STRINGIFY (LOCATION_TABLE)  \
   "                AND ((subject_type = 'user'"                           \
   "                      AND subject"                                     \
   "                          = (SELECT id FROM users"                     \
-  "                             WHERE users.uuid = '%s'))"                \
+  "                             WHERE users.uuid = $2))"                  \
   "                     OR (subject_type = 'group'"                       \
   "                         AND subject"                                  \
   "                             IN (SELECT DISTINCT \"group\""            \
@@ -524,7 +531,7 @@ acl_user_has_role (const char *user_uuid, const char *role_uuid)
   "                                       = (SELECT id"                   \
   "                                          FROM users"                  \
   "                                          WHERE users.uuid"            \
-  "                                                = '%s')))"             \
+  "                                                = $2)))"               \
   "                     OR (subject_type = 'role'"                        \
   "                         AND subject"                                  \
   "                             IN (SELECT DISTINCT role"                 \
@@ -533,39 +540,34 @@ acl_user_has_role (const char *user_uuid, const char *role_uuid)
   "                                       = (SELECT id"                   \
   "                                          FROM users"                  \
   "                                          WHERE users.uuid"            \
-  "                                                = '%s'))))"
+  "                                                = $2))))"
 
 /**
- * @brief Super clause arguments.
+ * @brief Structural arguments for ACL_SUPER_CLAUSE_PS.
  *
- * @param[in]  type     Type of resource.
- * @param[in]  field    Field to compare.  Typically "uuid".
- * @param[in]  value    Expected value of field.
- * @param[in]  user_id  UUID of user.
- * @param[in]  trash    Whether to search trash.
+ * The value and the user UUID are not among them; they are bound as $1
+ * and $2.
+ *
+ * @param[in]  type   Type of resource.
+ * @param[in]  field  Field to compare.  Typically "uuid".
+ * @param[in]  trash  Whether to search trash.
  */
-#define ACL_SUPER_CLAUSE_ARGS(type, field, value, user_id, trash) \
+#define ACL_SUPER_CLAUSE_PS_ARGS(type, field, trash)          \
   type,                                                       \
   trash ? (strcasecmp (type, "task") ? "_trash" : "") : "",   \
   type,                                                       \
   trash ? (strcasecmp (type, "task") ? "_trash" : "") : "",   \
   field,                                                      \
-  value,                                                      \
   type,                                                       \
   trash ? (strcasecmp (type, "task") ? "_trash" : "") : "",   \
   type,                                                       \
   trash ? (strcasecmp (type, "task") ? "_trash" : "") : "",   \
   field,                                                      \
-  value,                                                      \
   type,                                                       \
   trash ? (strcasecmp (type, "task") ? "_trash" : "") : "",   \
   type,                                                       \
   trash ? (strcasecmp (type, "task") ? "_trash" : "") : "",   \
-  field,                                                      \
-  value,                                                      \
-  user_id,                                                    \
-  user_id,                                                    \
-  user_id
+  field
 
 /**
  * @brief Test whether a user has Super permission on a resource.
@@ -581,18 +583,25 @@ static int
 acl_user_has_super_on (const char *type, const char *field, const char *value,
                        int trash)
 {
-  gchar *quoted_value;
-  quoted_value = sql_quote (value);
-  if (sql_int ("SELECT EXISTS (SELECT * FROM permissions"
-               "               WHERE " ACL_SUPER_CLAUSE ("'%s'") ");",
-               ACL_SUPER_CLAUSE_ARGS (type, field, quoted_value,
-                                      current_credentials.uuid, trash)))
-    {
-      g_free (quoted_value);
-      return 1;
-    }
-  g_free (quoted_value);
-  return 0;
+  gchar *statement;
+  int ret;
+
+  /* Value and user UUID are bound, so the statement text only varies with
+   * type, trash and field - see ACL_SUPER_CLAUSE_PS.  sql_quote is gone with
+   * the formatting: a bound parameter is never part of the statement text,
+   * so there is nothing to escape. */
+  statement = g_strdup_printf ("SELECT EXISTS (SELECT * FROM permissions"
+                               "               WHERE " ACL_SUPER_CLAUSE_PS
+                               ");",
+                               ACL_SUPER_CLAUSE_PS_ARGS (type, field, trash));
+
+  ret = sql_int_ps (statement,
+                    SQL_STR_PARAM (value),
+                    SQL_STR_PARAM (current_credentials.uuid),
+                    NULL);
+  g_free (statement);
+
+  return ret ? 1 : 0;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/gmp.c
```diff
@@ -19436,8 +19436,21 @@ handle_get_targets (gmp_parser_t *gmp_parser, GError **error)
...
+                   * statement each.  Over 1000 targets that saved two
+                   * statements per target; get_targets issued 12196 in
+                   * total while spending only 156 ms of 1498 ms actually
+                   * executing SQL, so the statement count is what costs.
+                   *
+                   * g_strdup keeps ownership as credential_name() and
+                   * credential_uuid() had it - both returned g_strdup'd
+                   * memory - so the free() calls at the end of the loop
+                   * stay correct for this branch and the trash branch
+                   * alike.  The iterator owns its own row buffer. */
```
rdttmr: This comment explains the change, not the code.

Without knowing the previous code, so outside of the context of this change set, most of this explanation is useless.

src/gmp.c
```diff
@@ -19436,8 +19436,21 @@ handle_get_targets (gmp_parser_t *gmp_parser, GError **error)
...
+                  ssh_name
+                    = g_strdup (target_iterator_ssh_credential_name (&targets));
+                  ssh_uuid
+                    = g_strdup (target_iterator_ssh_credential_uuid (&targets));
```
rdttmr: I think improving the `GET_TARGETS` behavior should be a separate PR.

Both changes are good, however reusing prepared statements and improving the performance of `GET_TARGETS` specifically are two different things.

src/gmp.c
```diff
@@ -19466,8 +19479,10 @@ handle_get_targets (gmp_parser_t *gmp_parser, GError **error)
                 {
                   credential_t found;
 
-                  smb_name = credential_name (smb_credential);
-                  smb_uuid = credential_uuid (smb_credential);
+                  smb_name
+                    = g_strdup (target_iterator_smb_credential_name (&targets));
+                  smb_uuid
+                    = g_strdup (target_iterator_smb_credential_uuid (&targets));
```
rdttmr: `GET_TARGETS` improvement, not prepared statements improvement

src/gmp.c
```diff
@@ -19498,8 +19513,10 @@ handle_get_targets (gmp_parser_t *gmp_parser, GError **error)
                 {
                   credential_t found;
 
-                  esxi_name = credential_name (esxi_credential);
-                  esxi_uuid = credential_uuid (esxi_credential);
+                  esxi_name
+                    = g_strdup (target_iterator_esxi_credential_name (&targets));
+                  esxi_uuid
+                    = g_strdup (target_iterator_esxi_credential_uuid (&targets));
```
rdttmr: `GET_TARGETS` improvement, not prepared statements improvement

src/gmp.c
```diff
@@ -19530,8 +19547,10 @@ handle_get_targets (gmp_parser_t *gmp_parser, GError **error)
                 {
                   credential_t found;
 
-                  snmp_name = credential_name (snmp_credential);
-                  snmp_uuid = credential_uuid (snmp_credential);
+                  snmp_name
+                    = g_strdup (target_iterator_snmp_credential_name (&targets));
+                  snmp_uuid
+                    = g_strdup (target_iterator_snmp_credential_uuid (&targets));
```
rdttmr: `GET_TARGETS` improvement, not prepared statements improvement

src/gmp.c
```diff
@@ -19562,8 +19581,12 @@ handle_get_targets (gmp_parser_t *gmp_parser, GError **error)
...
                   credential_t found;
 
-                  ssh_elevate_name = credential_name (ssh_elevate_credential);
-                  ssh_elevate_uuid = credential_uuid (ssh_elevate_credential);
+                  ssh_elevate_name
+                    = g_strdup (target_iterator_ssh_elevate_credential_name
+                                 (&targets));
+                  ssh_elevate_uuid
+                    = g_strdup (target_iterator_ssh_elevate_credential_uuid
+                                 (&targets));
```
rdttmr: `GET_TARGETS` improvement, not prepared statements improvement

src/gmp.c
```diff
@@ -19594,8 +19617,10 @@ handle_get_targets (gmp_parser_t *gmp_parser, GError **error)
                 {
                   credential_t found;
 
-                  krb5_name = credential_name (krb5_credential);
-                  krb5_uuid = credential_uuid (krb5_credential);
+                  krb5_name
+                    = g_strdup (target_iterator_krb5_credential_name (&targets));
+                  krb5_uuid
+                    = g_strdup (target_iterator_krb5_credential_uuid (&targets));
```
rdttmr: `GET_TARGETS` improvement, not prepared statements improvement

src/manage_acl.c
```diff
@@ -535,6 +535,104 @@ acl_user_has_role (const char *user_uuid, const char *role_uuid)
...
+ * pg_stat_statements.track_planning on, the two call sites below accounted
+ * for 241 ms of 387 ms total planning - while executing took 114 ms in
+ * total.  With the values in the text every call is a new statement to
+ * PostgreSQL, so it plans this clause from scratch every single time.
+ *
+ * Only the structural parts stay in the text: the resource type, the trash
+ * suffix and the field name.  Those come from gvmd's own tables of valid
+ * types, never from user input, and they take only a handful of
+ * combinations - which is what keeps the statement cache in sql_pg.c small.
+ */
```
rdttmr: This again explains why you changed from `ACL_SUPER_CLAUSE` to `ACL_SUPER_CLAUSE_PS`.

The comment should only provide an insight to what `ACL_SUPER_CLAUSE_PS` is, imo

src/manage_acl.c
rdttmr: Since this is no longer used, it can and should be removed

rdttmr: Also no longer used, can and should be removed

src/manage_acl.c
```diff
@@ -676,45 +792,62 @@ acl_user_owns_uuid (const char *type, const char *uuid, int trash)
...
+   * statement text only varies with type and trash.  This is called once per
+   * row of every listing - over 1000 targets it ran twice per target, for
+   * the port list and for the credential - and with the UUIDs in the text
+   * every one of those was a statement PostgreSQL had to parse and plan from
+   * scratch.
+   *
+   * type never carries user input; it comes from gvmd's own resource type
+   * names, and the callers that reach this point have already passed
+   * valid_db_resource_type.  sql_quote is therefore gone as well: a bound
+   * parameter is not part of the statement text and has nothing to escape. */
```
rdttmr: Again, explaining a change

src/manage_pg.c
```diff
@@ -1121,10 +1121,25 @@ manage_create_sql_functions ()
...
+   * Measured on the severity aggregate with track_planning on, the body of
+   * report_severity alone cost 220 ms of planning per call against 6 ms of
+   * execution.
+   *
+   * STABLE is the honest label here: all of these only read, and none can
+   * change its result within a single statement.  m_now () was already
+   * declared STABLE for the same reason.
+   *
+   * gvmd_user reads a session setting, which gvmd sets per connection and
+   * never changes mid statement. */
```
rdttmr:
```suggestion
  /* STABLE indicates that the function cannot modify the database,
   * and that within a single table scan it will consistently return the same result for the same argument values,
   * but that its result could change across SQL statements.
   */
```

src/manage_targets.h
```diff
@@ -144,6 +144,45 @@ target_iterator_ssh_elevate_credential (iterator_t *);
...
+
+const char *
+target_iterator_snmp_credential_uuid (iterator_t *);
+
+const char *
+target_iterator_ssh_elevate_credential_uuid (iterator_t *);
+
+const char *
+target_iterator_krb5_credential_uuid (iterator_t *);
+
```
rdttmr: `GET_TARGETS` improvement, not prepared statements improvement

src/manage_sql_targets.h
```diff
@@ -142,6 +142,82 @@
...
+   {                                                           \
+     "(SELECT uuid FROM credentials"                           \
+     " WHERE credentials.id"                                   \
+     "       = (SELECT credential FROM targets_login_data"     \
+     "          WHERE target = targets.id"                     \
+     "          AND type = CAST ('ssh' AS text)))",            \
+     NULL,                                                     \
+     KEYWORD_TYPE_STRING                                       \
+   },                                                          \
+   {                                                           \
```
rdttmr: `GET_TARGETS` improvement, not prepared statements improvement

(whole change, not just what I selected)

src/manage_sql_targets.c
```diff
@@ -2073,6 +2073,142 @@ target_iterator_krb5_trash (iterator_t* iterator)
...
+/**
+ * @brief Get the SMB credential name from a target iterator.
+ *
+ * @param[in]  iterator  Iterator.
+ *
+ * @return Credential name, or NULL if iteration is complete.
+ */
+DEF_ACCESS (target_iterator_smb_credential_name,
+            GET_ITERATOR_COLUMN_COUNT + 24);
+
```
rdttmr: `GET_TARGETS` improvement, not prepared statements improvement

(whole changes below as well)

src/manage_sql_resources.c
```diff
@@ -325,38 +325,57 @@ find_resource_with_permission (const char* type, const char* uuid,
...
-        g_free (quoted_uuid);
-        return TRUE;
-        break;
-    }
+  /* The UUID is bound rather than formatted in, so that the statement text
+   * only varies with type and trash - a handful of combinations, which the
+   * server side statement cache in sql_pg.c can keep plans for.  With the
+   * UUID in the text every call produced a different statement and
+   * PostgreSQL planned it afresh; get_targets over 1000 targets spent 403 ms
+   * planning against 115 ms executing.
```
rdttmr: This again explains a change

src/manage_sql_resources.c
```diff
@@ -325,38 +325,57 @@ find_resource_with_permission (const char* type, const char* uuid,
...
+        case 0:
+          break;
+        case 1:        /* Too few rows in result of query. */
+          *resource = 0;
+          break;
+        default:       /* Programming error. */
+          assert (0);
+        case -1:
+          g_free (statement);
+          g_free (quoted_uuid);
```
rdttmr: `quoted_uuid` may be freed sooner, since this code path uses a prepared statement now

src/manage_sql_tags.c
```diff
@@ -1372,21 +1372,38 @@ resource_tag_count (const char* type, resource_t resource, int active_only)
...
-                parent_type,
-                active_only ? "AND active=1": "");
+  /* Everything except the active_only clause is bound, so this has exactly
+   * two statement texts and the cache in sql_pg.c keeps a plan for each.
+   *
+   * Worth doing because it is called once per row of every listing: over
+   * 1000 targets it ran 1000 times and, with the values formatted into the
+   * text, PostgreSQL planned it 1000 times.  Measured with
+   * pg_stat_statements.track_planning on, that was 57 ms of planning per
+   * get_targets call against 2 ms of execution. */
```
rdttmr: It is worth doing anyway, because our plan is to move every parameterized query to a prepared statement :)

src/manage_sql_targets.c
```diff
@@ -2155,11 +2291,15 @@ target_task_iterator_readable (iterator_t* iterator)
...
 target_in_use (target_t target)
 {
-  return !!sql_int ("SELECT count(*) FROM tasks"
-                    " WHERE target = %llu"
-                    " AND target_location = " G_STRINGIFY (LOCATION_TABLE)
-                    " AND hidden = 0;",
-                    target);
+  /* Bound, not formatted in: this runs once per row of every target listing,
+   * and with the id in the text each call was a statement PostgreSQL had
+   * never seen before, so it parsed and planned it again every time. */
```
rdttmr: There is no need to justify moving any query to a prepared statement, as long as there are parameters in the query

robindittmar (commented):
Thank you for putting in the effort to improve our code!

I have added a whole bunch of comments and would kindly ask you to split the caching of prepared statements and the `GET_TARGETS` performance improvements into two separate PRs.

In addition, please let me know if you prefer to fix our concerns yourself, or if you would prefer for us to take over your PR(s), adjust the changes to our codebase and team decisions and take it from there.

TheBeast85:
Thanks for the review, and for the offer — I would like to do the changes myself.

On splitting: the diff falls into three parts, not two.

1. **Prepared statement reuse** — the cache in `sql_pg.c`, plus the parameterised queries in `manage_acl.c`, `manage_sql_resources.c`, `manage_sql_tags.c` and `target_in_use` in `manage_sql_targets.c`.
2. **`GET_TARGETS`** — `gmp.c`, `manage_targets.h`, `manage_sql_targets.h` and the iterator part of `manage_sql_targets.c`.
3. **`report_severity` in `manage_pg.c`**, split into a cached and a dynamic function so the expensive branch is only planned when it is reached. That is neither of the two — I would open it as a third PR unless you would rather have it elsewhere.

`manage_sql_targets.c` is the only file that lands in two of them. I will keep this PR for (1) and open separate ones for (2) and (3).

The remaining points I will fix as requested:

- Drop the comments that explain the change instead of the code.
- Remove `ACL_SUPER_CLAUSE` and `ACL_SUPER_CLAUSE_ARGS`, both unused after the change.
- Free `quoted_uuid` right after the parameter is bound.
- Apply your `STABLE` wording.

One heads-up: #3068 is based on this branch, so I will rebase it once the split is done.

TheBeast85:
@rdttmr friendly ping — this is ready for another look whenever you have time.

Everything from your review is addressed:

- Split into three PRs as discussed. This one is now prepared statements only; `GET_TARGETS` moved to #3080 and the `report_severity` planning change to #3081.
- `ACL_SUPER_CLAUSE` and `ACL_SUPER_CLAUSE_ARGS` removed.
- `quoted_uuid` freed as soon as it is no longer needed.
- Your `STABLE` wording applied (it now sits in #3081 together with the rest of `manage_pg.c`).
- Comments that explained the change rather than the code rewritten or dropped.

Just rebased onto current main; it applies cleanly and builds without new warnings.

No rush at all — I just wanted to make sure it had not gone unnoticed. #3068 sits on top of this one, so it is waiting on the same review.

PR #3068, opened 2026-07-30
Performance: speed up creating an override

TheBeast85:
## What

- Only write the session variables when the user actually changes
- Ask `to_regclass` instead of `information_schema.tables` in `manage_cert_loaded`, `manage_scap_loaded` and `manage_nvts_loaded`
- In `report_cache_counts`, only rebuild the counts that are being invalidated
- Bind the report as a parameter in the result severity iterator instead of formatting its id into the statement
- Index `overrides` on `nvt` and on `result_nvt`

## Why

Creating an override that is not tied to a task or a result affects every report the overridden NVT appears in and rebuilds the count cache for each one. With 1064 such reports that took over nine seconds, exactly linear in the number of reports.

Most of that was avoidable. `manage_session_init` costs three statements and ran about eleven times per report, nearly always setting the session to the user it was already on. `information_schema.tables` is a view over the catalogue and the privilege functions, and since the table name is formatted into the text it was re-planned on every call, twice per report. The counts that ignore overrides cannot be changed by an override at all, so rebuilding them only read the cache back and discarded it. And the `valid_overrides` clause carried the report id in its text, so every report was a statement PostgreSQL had never seen: 964 ms of the 2100 ms the database spent on the call, at only two calls per report.

Now 5.076 s instead of 9.334 s, and 31 statements per report instead of 60.

One visible change: a missing row for the unoverridden counts is no longer created as a side effect of creating an override. It appears on next access, which is what the path without auto cache rebuild always did.

## References

Depends on #3067

## Checklist

- [x] Tests: count cache compared row by row before and after, the 691 rows for the overridden counts are identical

### DIFF
--- a/src/manage_acl.c
+++ b/src/manage_acl.c
@@ -479,11 +479,18 @@ acl_user_has_role (const char *user_uuid, const char *role_uuid)
  *      is "name" then "SELECT ... format" will choose arbitrarily between
  *      the resources that have the same name. */
 /**
- * @brief Super clause.
+ * @brief Super clause with the resource value and the user UUID bound.
  *
- * @param[in]  format  Value format specifier.
+ * The resource value becomes $1 and the user UUID $2.  Both appear three
+ * times in the clause; PostgreSQL allows a parameter to be used more than
+ * once.
+ *
+ * The resource type, the trash suffix and the field name stay in the
+ * statement text.  They come from gvmd's own tables of valid types, never
+ * from user input, and take only a handful of combinations, which keeps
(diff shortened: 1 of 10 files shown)


TheBeast85:
Rebased onto the reduced #3067. The diff no longer carries the `GET_TARGETS` and severity changes, which moved to #3080 and #3081.

TheBeast85:
Rebased onto the updated #3067 and onto current main. No content change, it just follows the review fixes made over there.

PR #3098, opened 2026-08-18
Change: allow editing targets and port lists that are in use

TheBeast85:
## What

- `modify_target`: drop the "Target is in use" checks
- `create_port_range`: drop the "Port list is in use" check
- `port_list_writable`: report a port list as writable regardless of use, like target, schedule and credential already do

## Why

A target and its port list are frozen as soon as the first task refers to them, for the lifetime of that task. Correcting a wrong host in a scan target means creating a replacement target and repointing the task. For a port list it is worse: `port_list_writable` returns 0, so the web interface disables the edit icon altogether.

Nothing reads the target once a scan has started. `launch_osp_openvas_task` and `launch_openvasd_openvas_task` read hosts, port range, exclude hosts, alive test, reverse lookup settings and the credentials once, when they hand the scan to the scanner, and a report keeps the hosts it recorded. A later change therefore cannot reach a scan that is under way, nor make an earlier report unreadable.

Deleting is unchanged. `delete_target` and `delete_port_list` still refuse while a task refers to the resource.

## Testing

24.10 container stack, two tasks with their own targets sharing one port list. With the second task in state Requested: changed its target hosts, switched its port list, and added and removed a port range on the port list in use. All accepted, and the scan carried on. Deleting the target and the port list is still refused.

### DIFF
--- a/src/gmp.c
+++ b/src/gmp.c
@@ -24523,11 +24523,6 @@ gmp_xml_handle_end_element (/* unused */ GMarkupParseContext* context,
                                     "Port range TYPE must be TCP or UDP"));
                 log_event_fail ("port_range", "Port Range", NULL, "created");
                 break;
-              case 5:
-                SEND_TO_CLIENT_OR_FAIL
-                 (XML_ERROR_SYNTAX ("create_port_range",
-                                    "Port list is in use"));
-                break;
               case 6:
                 SEND_TO_CLIENT_OR_FAIL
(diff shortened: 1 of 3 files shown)


bjoernricks:
Same as https://github.com/greenbone/gsa/pull/5541

The "in use" feature is an important concept in our software and removing it has several consequences. It's not going to be removed soon.

PR #3099, opened 2026-08-19
Change: allow modifying a scan config that a task refers to

TheBeast85:
## What

- `modify_config`: drop the "Config is in use" checks from the family selection, the NVT selection and the preferences

## Why

A scan config is frozen as soon as the first task refers to it, for the lifetime of that task. Only name and comment stay editable, so correcting a single NVT selection means cloning the config and repointing every task that uses it.

Nothing reads the config once a scan has started. `launch_osp_openvas_task` and `launch_openvasd_openvas_task` take the selected NVTs and the preferences once, when they hand the scan to the scanner, so a later change cannot reach a scan that is under way.

Deleting is unchanged: `delete_config` still refuses while a task refers to the config, and predefined configs are still rejected in `modify_config_run`.

## Testing

24.10 container stack, a clone of "Full and fast" with a task pointing at it. Setting a scanner preference, selecting three NVTs of one family and sending a full family selection were all refused before and are accepted now; the family ended up at 3 of 9 NVTs. Deleting the config is still refused.

### DIFF
--- a/src/gmp_configs.c
+++ b/src/gmp_configs.c
@@ -817,10 +817,6 @@ modify_config_handle_family_selection (config_t config,
     {
       case 0:
         return 0;
-      case 1:
-        SENDF_TO_CLIENT_OR_FAIL_WITH_RETURN
-          (-1, XML_ERROR_SYNTAX ("modify_config", "Config is in use"));
-        return -1;
       case 2:
         SENDF_TO_CLIENT_OR_FAIL_WITH_RETURN
           (-1,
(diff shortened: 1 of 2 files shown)


bjoernricks:
Again as in https://github.com/greenbone/gvmd/pull/3098 this change is not going to be merged anytime soon. The "in use" feature is a concept of the whole application and removing it may have serious unwanted side effects. Without thoughtful consideration we wont change it.

Issue #2920, opened 2026-05-18
Bug: 26.26.0 build failure due to implicit-function-declaration


Narrat:
### This is an issue and not a support question which should be asked at https://forum.greenbone.net/?

- [x] This is an issue and not a support question

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Tried to compile the new 26.26.0 release and it fails due to `-Wimplicit-function-declaration` on `manage_utils.c`,  `manage_sql_alerts.c` and  `manage_alerts.c` when `assert()` is used but `assert.h` not included beforehand.

```
-- Configuring Greenbone Vulnerability Manager...
-- The C compiler identification is GNU 16.1.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found PkgConfig: /usr/bin/pkg-config (found version "2.5.1")
-- Could NOT find Git (missing: GIT_EXECUTABLE) 
-- Looking for clang-format...
-- Looking for clang-format... /usr/bin/clang-format
-- Install prefix: /usr
-- Log file: -
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Checking for module 'libcjson>=1.7.14'
--   Found libcjson, version 1.7.19
-- Checking for module 'libgvm_base>=22.39'
--   Found libgvm_base, version 23.0.0
-- Checking for module 'libgvm_util>=23.00'
--   Found libgvm_util, version 23.0.0
-- Checking for module 'libgvm_osp>=23.00'
--   Found libgvm_osp, version 23.0.0
-- Checking for module 'libgvm_gmp>=22.30'
--   Found libgvm_gmp, version 23.0.0
-- ENABLE_OPENVASD flag is not enabled
-- ENABLE_AGENTS flag is not enabled
-- ENABLE_CREDENTIAL_STORES flag is not enabled
-- ENABLE_CONTAINER_SCANNING flag is not enabled
-- ENABLE_JWT_AUTH flag is not enabled
-- Checking for module 'gnutls>=3.2.15'
--   Found gnutls, version 3.8.13
-- Checking for module 'glib-2.0>=2.42'
--   Found glib-2.0, version 2.88.1
-- Checking for module 'libbsd'
[...]

PR #3100, opened 2026-08-20
explicitly include assert.h in locations where used

Narrat:
## What

Without those explicit imports the compilation errored out in the changed locations. Reason being `implicit-function-declaration` errors. Happens since `gvmd 26.26.0` (up until now which is as of writing  `26.36.1`).

Example log:
```
gvmd-26.26.0/src/manage_sql_alerts.c: In function ‘copy_alert’: gvmd-26.26.0/src/manage_sql_alerts.c:67:3: error: implicit declaration of function ‘assert’ [-Wimplicit-function-declaration]
   67 |   assert (current_credentials.uuid);
      |   ^~~~~~
gvmd-26.26.0/src/manage_sql_alerts.c:18:1: note: ‘assert’ is defined in header ‘<assert.h>’; this is probably fixable by adding ‘#include <assert.h>’
   17 | #include <gvm/base/hosts.h>
  +++ |+#include <assert.h>
   18 |
```

With those changes the compilation doesn't error out anymore in regards to missing inclusion of `assert.h`.

## Why

It triggers `-Wimplicit-function-declaration`.
Odd is only that it appeared. IIRC the changelogs for GCC 15 and GCC 16 don't mention a change in that regard and my distribution `CFLAGS` weren't adjusted either.

## References

Got reported in gh-2920.

### DIFF
--- a/src/manage_alerts.c
+++ b/src/manage_alerts.c
@@ -24,6 +24,7 @@
 #include "manage_tickets.h"
 #include "manage_users.h"
 
+#include <assert.h>
 #include <bsd/unistd.h>
 #include <glib/gstdio.h>
 #include <grp.h>
--- a/src/manage_cve_scan.c
+++ b/src/manage_cve_scan.c
@@ -14,6 +14,8 @@
 #include "manage_sql.h"
 #include "manage_sql_assets.h"
 
+#include <assert.h>
+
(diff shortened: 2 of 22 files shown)


bjoernricks (approved):

bjoernricks:
@Narrat thanks!

PR #3103, opened 2026-08-24
Add web application VTs to results

timopollmeier:
## What
Information about web application VTs is added to results and their severity scores are used to set the severity of results.

## Why
This will allow the removal of redundant info from the description field and more specific severity scores of Web Application results.

## References
GEA-1885


src/gmp.c
```diff
@@ -10098,12 +10102,13 @@ results_xml_append_nvt (iterator_t *results, GString *buffer, int cert_loaded)
...
         buffer_xml_append_printf (buffer,
                                   "<nvt oid=\"%s\">"
-                                  "<type>nvt</type>"
+                                  "<type>%s</type>"
                                   "<name>%s</name>"
                                   "<family>%s</family>"
                                   "<cvss_base>%s</cvss_base>"
                                   "<severities score=\"%s\">",
                                   oid,
+                                  vt_type,
```
mattmundell: Could we put a `?: ""` on this, in case it's ever NULL?

timopollmeier: It should always be non-NULL, but I've added a check just in case.

a-h-abdelsalam (approved):

Issue #3106, opened 2026-08-28
Bug: relation "all_vts" does not exist (gvmd 26.37.0)


pcsegal:
### This is an issue and not a support question which should be asked at https://forum.greenbone.net/?

- [x] This is an issue and not a support question

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When running the current gvmd:stable Docker image from the greenbone community registry (which currently refers to version 26.37.0), using the most current docker compose file from https://greenbone.github.io/docs/latest/22.4/container/index.html#docker-compose-file, I get the error in the logs below when querying the GMP get_reports API (using python-gvm) after running an immediate scan against the host. Note that if use the docker image from an older version (by building it from https://github.com/greenbone/gvmd using the v26.36.1 tag, running `docker build -t gvmd . -f .docker/prod.Dockerfile` and replacing the image name in docker compose and recreating the container using the new image), the problem stops happening, so this seems to be an issue that started happening in 26.37.0.

```
md   main:MESSAGE:2026-08-28 10h00.59 utc:21:    Greenbone Vulnerability Manager version 26.37.0 (DB revision 281)
md   main:   INFO:2026-08-28 10h00.59 utc:21:    Migrating database.
md   main:MESSAGE:2026-08-28 10h00.59 utc:21: No Web Application VTs database found for migration
md   main:   INFO:2026-08-28 10h00.59 utc:21: gvmd: databases are already at the supported version
md   main:MESSAGE:2026-08-28 10h00.59 utc:22:    Greenbone Vulnerability Manager version 26.37.0 (DB revision 281)
md manage:   INFO:2026-08-28 10h00.59 utc:22:    Creating user.
User exists already.
md   main:MESSAGE:2026-08-28 10h01.05 utc:32:    Greenbone Vulnerability Manager version 26.37.0 (DB revision 281)
md manage:   INFO:2026-08-28 10h01.05 utc:32:    Getting users.
md   main:MESSAGE:2026-08-28 10h01.05 utc:35:    Greenbone Vulnerability Manager version 26.37.0 (DB revision 281)
md manage:   INFO:2026-08-28 10h01.05 utc:35:    Modifying setting.
md   main:MESSAGE:2026-08-28 10h01.11 utc:44:    Greenbone Vulnerability Manager version 26.37.0 (DB revision 281)
md manage:   INFO:2026-08-28 10h01.11 utc:44:    Modifying setting.
[...]

PR #3108, opened 2026-08-31
Change vt schema init

timopollmeier:
## What
Instead of the Web Application VTs tables, the `vts` schema is now
versioned. The schema is initialized and checked for migrations
independent of the `enable_web_application_scanning` feature flag.

## Why
This fixes SQL errors due to the vts.all_vts table not existing
if the web application scanning feature is disabled.

## References
#3106
GEA-2037


src/manage_sql.c
```diff
@@ -3970,15 +3970,12 @@ check_db_versions ()
...
-    {
-      ret = check_secinfo_db_version (
-                "Web Application VTs",
-                manage_web_application_vts_db_version,
-                manage_web_application_vts_db_supported_version);
-      if (ret)
-        return ret;
-    }
+
+  ret = check_secinfo_db_version ("VTs",
```
ozgen, line 3974: **Just to confirm:** does the upgrade path correctly migrate an existing `vts` schema using the old metadata key to the new `vts_database_version`?

ozgen (dismissed):

src/manage_sql.c
timopollmeier: Thanks for the hint! I've added check for this case.

ozgen (approved):

PR #3116, opened 2026-09-03
Change: Add an upper bound for images in OCI targets

a-h-abdelsalam:
## What
Added a command-line option `--max-images-per-oci-target=<number>` (Default: 1024, Maximum absolute value: 65536) to limit the number of images per OCI target.

Support image digests in overrides and notes scope.

## Why
To upper bound the number of images per OCI target.
Digests were validated as IPs when adding/modifying overrides and notes and therefore rejected.

## References
GEA-1820


src/manage_oci_image_targets.c
```diff
@@ -317,4 +348,119 @@ validate_oci_image_references (const char *image_refs_input,
...
+      gchar *digest = g_strstrip (*p);
+
+      if (*digest == '\0')
+        continue;
+
+      if (!is_valid_oci_image_digest (digest))
+        {
+          g_strfreev (parts);
+          g_hash_table_destroy (seen);
+          g_warning ("Invalid OCI image digest: %s", digest);
```
mattmundell: `digest` points into the parts array, so this should come before the `g_strfreev`.

a-h-abdelsalam: Thanks, should be fixed now.

src/manage_sql_oci_image_targets.c
```diff
@@ -326,6 +342,26 @@ modify_oci_image_target (const char *oci_image_target_id, const char *name,
           return MODIFY_OCI_IMAGE_TARGET_INVALID_IMAGE_URLS;
         }
 
+      gchar *target_excludes = oci_image_target_exclude_images (oci_image_target);
```
ozgen: I think the limit should be checked against the final target state. Currently this uses the existing `exclude_images`, and the check only runs when `image_references` is modified. For example, removing an existing exclusion without changing `image_references` could make the effective image count exceed the configured maximum without being validated.

a-h-abdelsalam: Should be fixed in [7bd3f44](https://github.com/greenbone/gvmd/pull/3116/commits/7bd3f441c2c28b67d12911f0715e95a759ac9bae)

ozgen (approved):

PR #3117, opened 2026-09-03
add: support optional encryption for agent support bundles

ozgen:
## What

- Add an optional `encryption` attribute to `get_agent_support_bundle`.
- Keep encrypted support bundles as the default behavior.
- Use the plain support bundle endpoint when `encryption="0"`.
- Update the required `gvm-libs` version when agent support is enabled.
- Update the GMP schema documentation and example.

## Why

- Allow clients to request decrypted agent support bundles through GMP.
- Keep the existing behavior backward compatible.

## References

GEA-2040

### DIFF
--- a/src/gmp_agent_support_bundle.c
+++ b/src/gmp_agent_support_bundle.c
@@ -31,6 +31,7 @@ typedef struct
 {
   gchar *agent_uuid; ///< UUID of the agent.
   gchar *days;       ///< Optional number of days to include.
+  gchar *encryption; ///< Whether to get encrypted the support bundle (0 or 1).
 } get_agent_support_bundle_t;
 
 /**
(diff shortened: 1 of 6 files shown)


src/gmp_agent_support_bundle.c
```diff
@@ -31,6 +31,7 @@ typedef struct
 {
   gchar *agent_uuid; ///< UUID of the agent.
   gchar *days;       ///< Optional number of days to include.
+  gchar *encryption; ///< Whether to get encrypted the support bundle (0 or 1).
```
a-h-abdelsalam, line 34: `encryption` is not freed in `get_agent_support_bundle_reset` but it is allocated in `get_agent_support_bundle_start`

ozgen: fixed with: https://github.com/greenbone/gvmd/pull/3117/changes/761cd48be86702371e7e84dbd509a136ffede538

a-h-abdelsalam (approved):

PR #3121, opened 2026-09-07
Added the option "-i" to the "sendmail" command of alerts.

jhelmold:
## What
Added the option "-i" to the "sendmail" command of alerts.

## Why
Without this option TXT-reports included in mails from alerts may end prematurely (see GEA-2032).

## References
GEA-2032

## Checklist
I could not reproduce the problem but according to the documentation in e. g. https://www.ibm.com/docs/en/aix/7.2.0?topic=s-sendmail-command and https://linux.die.net/man/8/sendmail.sendmail the "-i" option should be set and solve the Problem described in GEA-2032.

### DIFF
--- a/src/manage_alerts.c
+++ b/src/manage_alerts.c
@@ -986,7 +986,7 @@ email (const char *to_address, const char *from_address, const char *subject,
     }
 
   command = g_strdup_printf ("read FROM TO < %s;"
-                             " /usr/sbin/sendmail -f \"$FROM\" \"$TO\" < %s"
+                             " /usr/sbin/sendmail -f \"$FROM\" \"$TO\" -i < %s"
                              " > /dev/null 2>&1",
                              args_file_name,
                              content_file_name);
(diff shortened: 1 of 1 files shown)


src/manage_alerts.c
```diff
@@ -986,7 +986,7 @@ email (const char *to_address, const char *from_address, const char *subject,
     }
 
   command = g_strdup_printf ("read FROM TO < %s;"
-                             " /usr/sbin/sendmail -f \"$FROM\" \"$TO\" < %s"
+                             " /usr/sbin/sendmail -f \"$FROM\" \"$TO\" -i < %s"
```
ozgen: Should we move `-i` before `$TO`? The documented sendmail syntax is sendmail [option ...] [recipient ...]

jhelmold: Okay, sounds good.

ozgen (approved):

PR #3124, opened 2026-09-08
Add/download report export

ozgen:
## What

- Add the `download_report_export` GMP command to download completed asynchronous report exports.
- Return report export metadata together with the Base64 encoded generated file content.
- Remove the report export and generated file after a successful download.
- Reorganize `manage_delete_report_export` parameters to support direct cleanup by export ID and file path.
- Add periodic cleanup for old report exports in terminal states.
- Add GMP documentation for `download_report_export`.

## Why

- Allow clients to download completed asynchronous report exports.
- Avoid keeping downloaded or old report export files indefinitely.
- Clean up stale report export entities and files automatically.
- Document the new GMP command and its response format.

## References

GEA-2046


src/gmp_report_exports.c
```diff
@@ -258,3 +259,318 @@ get_report_exports_run (gmp_parser_t *gmp_parser, GError **error)
...
+  if (ret)
+    {
+      g_free (report_uuid_value);
+      g_free (delta_report_uuid_value);
+      g_free (report_format_uuid_value);
+      g_free (report_config_uuid_value);
+
+      report_export_data_free (data);
+      download_report_export_reset ();
+      return;
```
a-h-abdelsalam, line 552: Does it make sense to send an error here ?

src/gvmd.c
```diff
@@ -2011,6 +2011,94 @@ fork_report_export_scheduler ()
...
+  if (in_progress)
+    {
+      g_debug (
+        "%s: Cleaning old report exports skipped"
+        " because one is already in progress",
+        __func__);
+      return 1;
+    }
+
+  in_progress = TRUE;
```
a-h-abdelsalam, line 2036: `in_prog`ress is not reset if `sigemptyset` or `pthread_sigmask` fails

src/gvmd.c
```diff
@@ -2198,6 +2288,22 @@ run_report_exports (periodic_times_t *t)
...
+ * @param[in,out] t   Periodic timestamps; updates t->last_ on run.
+ */
+static void
+run_cleanup_old_report_exports (periodic_times_t *t)
+{
+  time_t now = time (NULL);
+  if (!time_to_run (t->last_cleanup_old_report_export,
+                    REPORT_EXPORT_CLEANUP_PERIOD, now))
+    return;
+  fork_cleanup_old_report_exports ();
```
a-h-abdelsalam: Do you want to check the return value from fork before updating the timestamp?

src/gmp_report_exports.c
ozgen: I don't think we should send an error here because the `200 OK` response has already started.

Instead, I am planning to add a file availability check before starting the response, so we can return a proper error earlier.

ozgen: fixed: https://github.com/greenbone/gvmd/pull/3124/changes/ef59884c45733ae80f1274d58d675b4a28a99698

src/gvmd.c
ozgen: fixed: https://github.com/greenbone/gvmd/pull/3124/changes/28fd664469974395b09e27f1305c682cec4cad7b

src/gvmd.c
ozgen: fixed: https://github.com/greenbone/gvmd/pull/3124/changes/28fd664469974395b09e27f1305c682cec4cad7b

a-h-abdelsalam (approved):

PR #3132, opened 2026-09-14
Add VT technical descriptions

timopollmeier:
## What
The feed sync can now load additional technical information about
vulnerability tests.

The `nvt` element in the `get_info` GMP response with type `nvt`
and details enabled now contains a `tech_info` element with the
technical description markdown.

## Why
The technical descriptions should eventually be shown on the GSA NVT details pages if they have been downloaded from the feed server.

## References
GEA-2075

## Checklist

- [ ] I have added tests for the changes
- [ ] I have used the following LLMs/AI tools in this pull request:


src/manage_sql_secinfo.c
```diff
@@ -7545,6 +7564,219 @@ manage_sync_web_application_vts (sigset_t *sigmask_current)
...
+      if (error->code == G_FILE_ERROR_NOENT)
+        stamp = 0;
+      else
+        {
+          g_warning ("%s: Failed to get timestamp: %s",
+                     __func__,
+                     error->message);
+          stamp = time(NULL);
+        }
+    }
```
mattmundell, line 7600: Should free `error`.

src/manage_sql_vt_tech_info.c
```diff
@@ -0,0 +1,287 @@
...
+  if (vt_tech_info_json_skip_to_items (&parser, &event))
+    {
+      gvm_json_pull_event_cleanup (&event);
+      gvm_json_pull_parser_cleanup (&parser);
+      fclose (stream);
+      sql_rollback ();
+      return -1;
+    }
+
+  sql ("TRUNCATE vts.vt_tech_info");
```
ozgen: I think the `TRUNCATE` should happen once before processing the feed rather than inside `update_vt_tech_info_from_file()`.

src/manage_sql_vt_tech_info.c
```diff
@@ -0,0 +1,287 @@
...
+
+  g_info ("Updating VT Technical Information");
+
+  while ((file_path = g_dir_read_name (dir)))
+    if (g_str_has_suffix (file_path, ".json.gz")
+        || g_str_has_suffix (file_path, ".json"))
+      {
+        gchar *full_path = g_build_filename (GVM_VT_TECH_INFO_DIR,
+                                             file_path, NULL);
+        update_vt_tech_info_from_file (full_path);
```
ozgen: `update_vt_tech_info_from_file()` can return `-1`, but the result is ignored here. Is this intentional? at least can we add warning log if return `-1`?

src/manage_sql_vt_tech_info.c
ozgen: For example:

```
TRUNCATE
import file1
TRUNCATE // removes the data imported from file1
import file2
```

src/manage_sql_vt_tech_info.c
```diff
@@ -0,0 +1,287 @@
...
+ * @return 0 success, -1 error.
+ */
+int
+update_vt_tech_info_from_feed_files () {
+
+  GError *error = NULL;
+  GDir *dir;
+  const gchar *file_path;
+
+  dir = g_dir_open (GVM_VT_TECH_INFO_DIR, 0, &error);
```
ozgen, line 249: I think we should also close `dir` with `g_dir_close(dir)` before returning. There are also a couple of error paths where `json` is not freed before returning.

ozgen:
if possible can we also extend our `glossary` with VT technical description?

src/manage_sql_vt_tech_info.c
```diff
@@ -0,0 +1,295 @@
...
+    {
+      g_warning ("%s: Failed to open directory '%s': %s",
+                 __func__, GVM_VT_TECH_INFO_DIR, error->message);
+      g_error_free (error);
+      return -1;
+    }
+
+  g_info ("Updating VT Technical Information");
+
+  sql ("TRUNCATE vts.vt_tech_info");
```
ozgen, line 260: Should the whole technical-info feed update be handled in one transaction?

timopollmeier: In theory it would make sense to handle the whole update in a single transaction to ensure consistency in case of an error.
In practice, inserting too much data in a single transaction can cause performance issues, so for this reason it makes sense to split the update into multiple transactions.

ozgen (approved):

PR #3150, opened 2026-09-22
Set user to not editable if the user is the current user.

jhelmold:
## What
Set the user of the users table in GSA to not editable, when the user of the resource is the current user.

## Why
This was a bug and caused problems.

## References
GEA-1971

## Checklist


src/gmp_get.c
```diff
@@ -349,9 +349,13 @@ send_get_common (const char *type, get_data_t *get, iterator_t *iterator,
...
            == 0))
-      /* Or the user is effectively the owner. */
-      || acl_user_has_super (current_credentials.uuid,
-                             get_iterator_owner (iterator))
+      /* Or the user is effectively the owner and the user is not the user
+       * of the resource if the resource is of type user. */
+      || (acl_user_has_super (current_credentials.uuid,
+                              get_iterator_owner (iterator))
+          && (strcmp (type, "user")
+              || strcmp (get_iterator_name (iterator), current_credentials.username)
```
ozgen, line 357: Would it be better to compare the user UUIDs here instead of the usernames?
Since we want to check whether this user resource represents the current user, comparing the UUIDs seems more direct.

ozgen (approved):

